PR #56, opened 2016-11-29
Fixed recording duration

erickadbay:


app/assets/javascripts/recordings.coffee
```diff
@@ -140,8 +140,14 @@ class @Recordings
...
-        totalMinutes = Math.round((new Date(recording.end_time) - new Date(recording.start_time)) / 1000 / 60)
-        recording.duration = totalMinutes
+        hours = Math.floor(recording.length/60)
+        minutes = recording.length % 60
+        recording.duration =  if minutes>=60
+                                hours+"hours and "+minutes+" minutes"
+                              else if minutes>1
+                                minutes+" minutes"
+                              else
+                                "< 1 minute"
```
zach-chai: The only problem is the strings need to be localized
might be simpler to just leave it in minutes

PR #168, opened 2017-04-26
LDAP Authentication

Kristof-VD-Ouweland:
Adding LDAP authentication to the Greenlight application for BigBlueButton


Kristof-VD-Ouweland:
Off course, all comments should be removed but gives an indication why I added.

zach-chai:
It looks like there are some unrelated changes that should be removed, mostly the changes without comments

config/initializers/omniauth.rb
```diff
@@ -1,11 +1,28 @@
...
+  provider :ldap,
+ 	:title => "LDAP Authenitication",
+	:host => 'ldap server',
+	:port => 389,
+	:method => :plain,
+	:base => ' base location ou=users,dc=domain,dc=com',
+	:uid => 'uid',
+	:name_proc => Proc.new {|name| name.gsub(/@.*$/,'')},
+   	:bind_dn => 'bind user dn=admin,ou=users,dc=domain,dc=com',
+	:password => 'bind user password'
```
zach-chai: These need to be environment variables as you mentioned in the comment

Kristof-VD-Ouweland: You are correct, I'm still having an issue with one variable ':plain' which is not passed correctly

Depending on how I set the varibale (single, double quotes or none) the variable is not correctly passed to the auth module

OmniAuth::LDAP::Adaptor::ConfigurationError (":plain" is not one of the available connect methods: :ssl, :tls, :plain):

joshua-arts:
Closing this pull request because it makes many unneeded changes to Greenlight that are not related to LDAP authentication.

Issue #181, opened 2017-07-25
add hd parameter for google_oauth2 


phavekes:
please add the hd parameter for google_oauth in the env file and in greenlight/config/initializers/omniauth.rb as parameter for 'provider :google_oauth2'

> hd: (Optional) Limit sign-in to a particular Google Apps hosted domain. This can be simply string 'domain.com' or an array %w(domain.com domain.co). 

[From here](https://github.com/zquestz/omniauth-google-oauth2)

joshua-arts:
I've gone ahead the merged PR #182. Thanks!

PR #182, opened 2017-07-25
untested suggestion for issue #181

phavekes:
#181


config/initializers/omniauth.rb
```diff
@@ -8,8 +8,13 @@
...
   provider :twitter, ENV['TWITTER_ID'], ENV['TWITTER_SECRET']
-  provider :google_oauth2, ENV['GOOGLE_OAUTH2_ID'], ENV['GOOGLE_OAUTH2_SECRET'],
-    scope: ['profile', 'email', 'youtube', 'youtube.upload'], access_type: 'online', name: 'google'
+  provider :google_oauth2,
+    ENV['GOOGLE_OAUTH2_ID'],
+    ENV['GOOGLE_OAUTH2_SECRET'],
+    scope: ['profile', 'email', 'youtube', 'youtube.upload'], 
+    access_type: 'online', 
+    name: 'google',
+    hd: ENV['GOOGLE_OAUTH2_HD']
```
joshua-arts: This line should be:

`hd: ENV['GOOGLE_OAUTH2_HD'].blank? ? nil : ENV['GOOGLE_OAUTH2_HD']`

joshua-arts:
Hey @phavekes,

Thanks for the PR!

I did some testing and it appears then when the `GOOGLE_OAUTH2_HD` variable is set in the `env` file that it works as expected, only allowing authentication from the specified domain.

However, when the `GOOGLE_OAUTH2_HD` variable is left blank (and thus equal to an empty string), it still registers the `hd` provider option as an empty string. Since that's not a domain, it doesn't allow authentication from anyone, and gives back this error:

`ERROR -- omniauth: (google) Authentication failure! invalid_credentials: OmniAuth::Strategies::OAuth2::CallbackError, invalid_hd | Invalid Hosted Domain`

If `GOOGLE_OAUTH2_HD` is empty, it should not register the `hd` option when creating the provider. An easy solution to this is to just set it to `nil` (see comment in code).

config/initializers/omniauth.rb
phavekes: thnx. corrected it

env
```diff
@@ -33,8 +33,10 @@ TWITTER_SECRET=
 #   For the callback URL use 'http://<your hostname:port>/auth/google/callback'
 #   Once registered copy the ID and Secret here
 #
+#   GOOGLE_OAUTH2_HD: (Optional) Limit sign-in to a particular Google Apps hosted domain. This can be simply string 'domain.com'
```
joshua-arts: Line 36 should follow same format as rest of the `env` file.

#   The GOOGLE_OAUTH2_HD variable is used to limit sign-in to a particular Google Apps hosted domain. This
#    can be a string such as, 'domain.com'. If left blank, GreenLight will allow sign-in from all Google Apps hosted
#    domains.
#

ffdixon:
Hi Peter, we received your signed contributor agreement -- thanks!

PR #184, opened 2017-08-15
Feature qrcode

jfederico:
When this feature is enabled, a room QR code can be generated by the owner.


mathjayden:
How can I enable this feature back in the latest greenlight?

Issue #180, opened 2017-07-25
Support SAML2 for authentication


phavekes:
Could SAML2 be added as authentication method?

[https://github.com/omniauth/omniauth-saml](https://github.com/omniauth/omniauth-saml
)

phavekes:
This should be a starting-point : https://github.com/omniauth/omniauth-saml

joshua-arts:
Closing as #185 has been merged and GreenLight now supports SAML authentication.

miconx:
i cannot find any documentation  about this - has this been removed?

farhatahmad:
Hi @miconx,

I'm not sure if this was ever merged into master. At the moment, SAML authentication is not supported in Greenlight.

PR #185, opened 2017-08-18
Added SAML2 support

phavekes:
#180


jfederico:
Hi @phavekes,
First of all, thanks for the PR. I know it has been a while, but we are trying to catch up with projects.

Can you please condensate all your code in one commit and update the Gemfile?

I'll check it out as soon as you do those updates

foobarable:
Is there any chance, that this MR would make it in the current master? It would help us a lot in our setup

farhatahmad:
Hi @foobarable,

We ran into tons of issues with SAML after we implemented it, so it was removed.

You're free to pull this PR into you're own deployment if you want

foobarable:
Uh, thats sad to hear. Then we may not want to head that way. What kind of problems were those? Maybe we could be of any help with those?

foobarable:
Are other omniAuth strategies more likely to be merged into master? What about OpenID? (I could open an issue for that, if it's more fitting to discuss this someplace else)

hex-m:
There is a post about OIDC (OpenID Connect) in #569 

> We actually have this in place in another repo.

I guess he is refering to [this](https://github.com/bigbluebutton/omniauth-bbbltibroker) which is not very active.

I vote for a dedicated issue to discuss this. :)

PR #239, opened 2018-08-14
Brazilian Portuguese Translate

britoederr:


config/application.rb
```diff
@@ -34,7 +34,7 @@ class Application < Rails::Application
     config.exceptions_app = routes
 
     # Configure I18n localization.
-    config.i18n.available_locales = %w(en)
+    config.i18n.available_locales = [:en, :"pt-br"]
```
joshua-arts: We should probably stay consistent and use `%w()` string array syntax. Instead of defining the array literally, we should use `%w(en pt-br)`.

PR #261, opened 2018-09-12
Fix #260 issues with privacy policy

### LINKED ISSUE #260: Fix privacy policy sign up flow
Currently when the privacy policy is enabled and a user clicks signup their account is created before they accept the privacy policy. We need to change the privacy policy flow to not save a user until they have agreed to the privacy policy.

image

jiama843:
Changed privacy policy workflow from:

Input Credentials -> Create user -> Accept terms -> Redirect to Main Room

to:

Input Credentials and Accept terms -> Create User -> Redirect to Main Room

### DIFF
--- a/app/controllers/rooms_controller.rb
+++ b/app/controllers/rooms_controller.rb
@@ -17,7 +17,6 @@
 # with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
 
 class RoomsController < ApplicationController
-  before_action :validate_accepted_terms, unless: -> { !Rails.configuration.terms }
   before_action :find_room, except: :create
   before_action :verify_room_ownership, except: [:create, :show, :join, :logout]
 
@@ -177,9 +176,5 @@ class RoomsController < ApplicationController
       # Redirect unauthenticated users to root.
       redirect_to root_path
     end
-  end
-
-  def validate_accepted_terms
-    redirect_to terms_path unless current_user.accepted_terms
   end
 end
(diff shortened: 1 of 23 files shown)


config/routes.rb
```diff
@@ -26,12 +26,16 @@
   get '/signup', to: 'users#new', as: :signup
   post '/signup', to: 'users#create', as: :create_user
 
+  # Redirect to terms page
+  match '/terms', to: 'users#terms', via: [:get, :post]
```
jfederico: Why do you have get and post?

config/routes.rb
```diff
@@ -26,12 +26,16 @@
...
+
+  # Accept terms for existing accounts
+  get '/accept', to: 'users#accept', as: :accept
```
jfederico: Is this really necessary? Can't be just an update to the terms?

app/controllers/users_controller.rb
```diff
@@ -97,14 +97,16 @@ def destroy
...
+    # Render the terms page
+    render "terms"
+  end
 
-    if params[:accept] == "true"
-      current_user.update_attribute(accepted_terms: true)
-      redirect_to current_user.main_room
-    end
+  # GET /accept
+  def accept
```
jfederico: lets be more ecplicit here. Lets call the method accept_terms

app/controllers/rooms_controller.rb
```diff
@@ -17,7 +17,6 @@
 # with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
 
 class RoomsController < ApplicationController
```
jfederico: Code for validation in the room has to be resored

config/routes.rb
jiama843: Yeah, post is used to accept the terms

PR #279, opened 2018-09-26
500 internal server error (fixes #277)

### LINKED ISSUE #277: Logging in using Greenlight credentials while there is an existing omniauth account causes 500 internal server error
If I log into an account like this:

image

BUT the email is already linked with an existing omniauth account, I get a 500 internal server error upon login:

image

BN#GRN-28

jiama843:
Fix for the following issue:

https://github.com/bigbluebutton/greenlight/issues/277


app/controllers/sessions_controller.rb
```diff
@@ -28,7 +28,9 @@ def destroy
   # POST /users/login
   def create
     user = User.find_by(email: session_params[:email])
-    if user.try(:authenticate, session_params[:password])
+    if user && !user.greenlight_account? && session_params[:password]
```
jfederico: Do we really need the veryfy that password is included?
```
if user && !user.greenlight_account? && session_params[:password]
```
shouldn't be something like this
```
if user && !user.greenlight_account?
```
Password will be always there as the form passes it either empty or with a value

PR #289, opened 2018-10-09
Fixed room view broken when no format presentation is included (GRN-34)

jfederico:


jfederico:
It also fixes some other issues with the view

image

PR #290, opened 2018-10-10
Fixed #269 Add health-check endpoint (GRN-3)

### LINKED ISSUE #269: Add a health-check endpoint
As of now, there is no endpoint that is used to check the status of the rails server, database, migrations and other details. Adding a dedicated endpoint can make this app further robust to run in dev/production.

Here is a good start: https://github.com/ianheggie/health_check

Note:
   We probably want to suppress health-check logs as it could clutter production logs

jiama843:
Added health check endpoint for the following issue:
https://github.com/bigbluebutton/greenlight/issues/269

Access the checkpoint by adding "/health_check" to the end of the base url.


bruckwubete:
Looks good!

PR #296, opened 2018-10-12
Updated styles for the Google Auth button (GRN-31)

tylercopeland:
We received feedback from our community that the current Google button used for "Sign In" and Sign Up" did not meet Google's branding guidelines. 

https://developers.google.com/identity/branding-guidelines#top_of_page

**Current Google authentication buttons**
screen shot 2018-10-12 at 10 46 23 am

screen shot 2018-10-12 at 10 46 33 am

**Updated Google authentication buttons**

screen shot 2018-10-12 at 10 51 30 am
screen shot 2018-10-12 at 10 51 23 am

<hr>

**A couple of things to note about this update.** 
-  I broke the Google and Twitter auth buttons out so that way we can apply Google specific styling. Thanks, @jiama843 for the support on this. 

-  Google requires that we use "Sign In" and "Sign Up" for their buttons. That means I needed to update the en.yml file. Specifically  "login: Sign in". I'm not sure if it matters, but we may want to clean up the en.yml file to use "signin:Sign in". I was going to tackle this myself, but I realized "login" is used throughout several pages and didn't want to break anything. What are your thoughts on this @jfederico @john?


sample.env
```diff
@@ -25,8 +25,8 @@ BIGBLUEBUTTON_SECRET=
 # The GOOGLE_OAUTH2_HD variable is used to limit sign-in to a particular Google Apps hosted
 # domain. This can be a string such as, 'domain.com'. If left blank, GreenLight will allow
 # sign-in from all Google Apps hosted domains.
-GOOGLE_OAUTH2_ID=
-GOOGLE_OAUTH2_SECRET=
+GOOGLE_OAUTH2_ID=59890108241-4601a8gll5olv445aa0j9lskec0dm69t.apps.googleusercontent.com
+GOOGLE_OAUTH2_SECRET=5Q8fmAY9DMDjveSkIIMcfEP3
```
jiama843: Just make sure the sample.env was the way it was previously (i.e unfilled).

app/views/users/new.html.erb
```diff
@@ -22,7 +22,17 @@
         </div>
         <div class="card-body background">
           <% unless configured_providers.length.zero? %>
-            <% configured_providers.each do |provider| %>
+
+            <% if provider = configured_providers[0] %>
```
jfederico: A couple of details here. in the if condition, it should be == oinstead of =. As it is right now, both are always true.
<% if provider == configured_providers[0] %>

Second, I am not sure we want to do it this way, If the code has to be specific for google, we can make the comparision looking specifically for google as the tool provider (I don;t remember how the strategies are identified, but we'll figure it out)

PR #300, opened 2018-10-16
Fix for email verification issue (GRN-36)

jiama843:
The remainder of the fix for the email verification issue.


config/environments/development.rb
```diff
@@ -42,7 +42,7 @@
   }
 
   # Don't care if the mailer can't send.
```
jfederico: Is it now "DO care if the mailer can't send'?

PR #301, opened 2018-10-17
Unify the UI style in sign up, sign in buttons when using omniauth (GRN-35)

jfederico:


app/views/users/new.html.erb
```diff
@@ -22,30 +22,14 @@
         </div>
         <div class="card-body background">
           <% unless configured_providers.length.zero? %>
-
-            <% if configured_providers.include? :google %>
-              <% provider = :google %>
+            <% configured_providers.each do |provider| %>
```
jiama843:
```suggestion
            <% iconset_providers.each do |provider| %>
```

PR #307, opened 2018-10-23
Fixed #306 Restrict domain on Office365 OAuth2 (GRN-39)

### LINKED ISSUE #306: Add restriction by domain for Office365 signin
As an addition to Office365 OAuth2, since specialized domains for certain groups using Office365 accounts is pretty common, it would be helpful to allow restriction by domain.

jiama843:
Added the feature:
https://github.com/bigbluebutton/greenlight/issues/306


jfederico:
The only problem with this is that the Gem we selected is not updated and the developer is not providing support, even though we submiuted a PR.
Since we don't want to fork and maintain a gem, me may want to look for a different one.

jfederico:
As the guy who developed the gem we are using has not responded, and we don't want to maintain a OAuth2 gem for this, we have to find a different gem to replace the one we originally found.

jfederico:
The Gem was replaced.
Closing this PR

PR #308, opened 2018-10-23
Fixed UI for join button (GRN-40)

jiama843:
This is what it looks like now:

image


jfederico:
I am not sure if we should merge this one as the button has been like that since ever.

jiama843:
Yeah, the current button might not be as uniform, but it is bigger which might be better from a UX perspective.

PR #311, opened 2018-10-26
Allow for specification of email sender

jiama843:
This feature allows a user to specify the email address outgoing mail is sent from and is necessary to set up aws ses.

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -17,7 +17,7 @@
 # with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
 
 class UserMailer < ApplicationMailer
-  default from: 'notifications@example.com'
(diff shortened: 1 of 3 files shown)


app/views/user_mailer/verify_email.html.erb
```diff
@@ -22,7 +22,7 @@
     <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
   </head>
   <body>
-    <h1>Welcome to Greenlight!, <%= @user.name %></h1>
+    <h1>Welcome to BigBlueButton!, <%= @user.name %></h1>
     <p>
       You have successfully signed up for Greenlight,
```
jfederico: Shouldn't we be consistent here too? Greenlight or BigBlueButton

PR #324, opened 2018-11-15
 Fixed #323 Allow users to select a language in settings (GRN-17)

### LINKED ISSUE #323: Allow users to select a language in settings (GRN-17)
Greenlight and BBB are multilanguage applications. BUT, while BBB allow users to choose the preferred language, greenlight shows the content in the language set up as default for the browser.

A registered user can choose from a list (or group) of languages enabled.

jiama843:
Added feature for the following:
https://github.com/bigbluebutton/greenlight/issues/323

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -181,7 +181,7 @@ GEM
     public_suffix (3.0.3)
     puma (3.11.4)
     pyu-ruby-sasl (0.0.3.3)
-    rack (2.0.5)
+    rack (2.0.6)
     rack-test (0.6.3)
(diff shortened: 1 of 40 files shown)


jfederico:
Two more details:
When selecting a language other than English
1. If the string is not translated, it should failover English
2. When applying the language, the flash message should be shown in the correct language. Right now the language has to be appliued twice.

jiama843:
I was able to fix the flash message but it seems that the default locale works as intended. I did some testing with existing components and it appears that when there is no value specified for the translation, the key is used instead.

PR #330, opened 2018-12-07
Add tags to recordings (GRN-13)

jiama843:
Users have a new way to rename recordings, furthermore, they can add tags to recordings and list them with an addition of a new modal:

image


farhatahmad:
Out of scope

PR #331, opened 2018-12-12
Fixed 500 error when BBB endpoint cannot be accessed (GRN-44)

jiama843:
Banner message shows when bbb endpoint cannot be accessed:
image

Limited functionality includes:
- Creating rooms
- Renaming rooms
- Modifying user settings

All other functions are disabled:
e.g Start meeting Button:
image


app/helpers/application_helper.rb
```diff
@@ -46,6 +46,28 @@ def bigbluebutton_endpoint_default?
     Rails.configuration.bigbluebutton_endpoint_default == Rails.configuration.bigbluebutton_endpoint
   end
 
+  def bbb_endpoint
```
jfederico: This new code in the helper is duplicated in the controller. You can either keep it in the helper and include the helper in the controller when required OR, keep it in the controller and declare each method as a helper.

app/views/shared/components/_room_block.html.erb
```diff
@@ -55,9 +55,11 @@
               <a href="" id="rename-room-button" class="dropdown-item">
                 <i class="dropdown-icon far fa-edit"></i> <%= t("rename") %>
               </a>
-              <a href="" data-toggle="modal" data-target="#deleteRoomModal_<%= room.uid %>"class="dropdown-item">
-                <i class="dropdown-icon far fa-trash-alt"></i> <%= t("delete") %>
-              </a>
+              <% if test_bbb_connection %>
```
jfederico: change this request with a global variable

spec/controllers/application_controller_spec.rb
```diff
@@ -21,8 +21,14 @@
 describe ApplicationController, type: :controller do
   describe "Application Controller Tests" do
     before { allow(Rails.configuration).to receive(:recording_thumbnails).and_return(true) }
-    before { allow(Rails.configuration).to receive(:bigbluebutton_endpoint).and_return(:defaultBBBendpoint) }
-    before { allow(Rails.configuration).to receive(:bigbluebutton_endpoint_default).and_return(:defaultBBBendpoint) }
+    before do
+      allow(Rails.configuration).to receive(:bigbluebutton_endpoint)
+        .and_return("http://test-install.blindsidenetworks.com/bigbluebutton/api/")
```
jfederico: The URL should not be hardcoded

spec/controllers/application_controller_spec.rb
```diff
@@ -21,8 +21,14 @@
...
+    end
+    before do
+      allow(Rails.configuration).to receive(:bigbluebutton_endpoint_default)
+        .and_return("http://test-install.blindsidenetworks.com/bigbluebutton/api/")
```
jfederico: Same here. URL is hardcoded

app/views/shared/_room_event.html.erb
```diff
@@ -40,4 +40,6 @@
   </div>
 </div>
 
-<%= render "shared/sessions", recordings: @room.public_recordings, only_public: true %>
+<% if test_bbb_connection %>
```
jfederico: change this request with a global variable

app/views/rooms/show.html.erb
```diff
@@ -77,4 +79,6 @@
   </div>
 </div>
 
-<%= render "shared/sessions", recordings: @recordings, only_public: false %>
+<% if test_bbb_connection %>
```
jfederico: change this request with a global variable

app/views/rooms/show.html.erb
```diff
@@ -49,7 +49,9 @@
         </form>
       </div>
       <div class="col-lg-3 col-sm-12 force-bottom mt-5">
-        <% if @is_running %>
+        <% if !test_bbb_connection %>
```
jfederico: change this request with a global variable

app/views/rooms/join.html.erb
```diff
@@ -22,7 +22,12 @@
           placeholder: t("enter_your_name"),
           value: "#{current_user ? current_user.name : ''}",
           readonly: !current_user.nil? %>
-      <%= f.submit t("room.join"), class: "btn btn-primary btn-sm col-sm-3 form-control join-form" %>
+
+      <% if test_bbb_connection %>
```
jfederico: change this request with a global variable

app/views/layouts/application.html.erb
```diff
@@ -47,6 +47,13 @@
         <% end %>
       <% end %>
 
+      <% if !test_bbb_connection %>
```
jfederico: You are sending too many requests to BBB. You can better test the server availability once and assign its state to a global variable @bbb_available = [true | false]

spec/controllers/application_controller_spec.rb
```diff
@@ -43,5 +49,9 @@
     it "Returns relative root" do
       expect(controller.relative_root).to eq("")
     end
+
+    it "tests bbb connection" do
```
jfederico: change this request with a global variable

jfederico:
As this approach may lead to an overflow of requests to the BBB server(s), we should store the result in cache memory and look for the server status in the cache before making a new request. This way if there are tons of users in GL and they are switching pages, we don't have tons of unnecessary requests in BBB.

The other thing to keep in mind is that the expiration for the cahced key, that may be 5 mins by default, should be configured through and ENV variable.

PR #338, opened 2018-12-21
Update ru.yml

plastilincheg:
Change some mistakes


jiama843:
Hi @plastilincheg, you've been added to the transifex team for the Russian locale. You should be able to modify the translations there.

PR #346, opened 2019-01-25
Updates to scripts for CI/CD

jfederico:


jfederico:
@bruckwubete  I am externalizing all the scripts we have in CI, right now it is very tight. But no worries, this was just to trigger jenkins and see if the example file was working.

I am also running some tests on GitLab, not sure yet which one should we keep at the end. 

The Open Source  image is more likely going to be build with TravisCI

PR #351, opened 2019-01-30
GRN-45: Fixed the URL being sent in the password reset email

farhatahmad:


test/mailers/previews/user_mailer_preview.rb
```diff
@@ -1,4 +1,20 @@
 # frozen_string_literal: true
 
 class UserMailerPreview < ActionMailer::Preview
+  # Preview this email at
+  # http://localhost:3000/rails/mailers/user_mailer/password_reset
+  def password_reset
+    user = User.first
+    user.reset_token = User.new_token
+    url = "http://localhost:3000" + "/password_resets/" + user.reset_token + "/edit?email=" + user.email
```
jfederico: Avoid the use of localhost, specially for testing environments, unless you really want to use localhost for something, like a database.
It is a common practice to use something that is more evident to be a mock-up, like example.com

test/mailers/previews/user_mailer_preview.rb
```diff
@@ -1,4 +1,20 @@
...
+    UserMailer.password_reset(user, url)
+  end
+
+  # Preview this email at
+  # http://localhost:3000/rails/mailers/user_mailer/verify_email
+  def verify_email
+    user = User.first
+    url = "http://localhost:3000" + "/u/verify/confirm/" + user.uid
```
jfederico: Same here

test/mailers/previews/user_mailer_preview.rb
farhatahmad: That would make sense. Done

test/mailers/previews/user_mailer_preview.rb
farhatahmad: Done

PR #367, opened 2019-02-11
GRN-56: Correctly implemented the account verification flow

farhatahmad:


app/controllers/account_activations_controller.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
```
jfederico: Why do we need the frozen literals?

app/controllers/rooms_controller.rb
```diff
@@ -240,7 +242,16 @@ def validate_accepted_terms
 
   def validate_verified_email
     if current_user
-      redirect_to resend_path unless current_user.email_verified
+      redirect_to account_activation_path(current_user) unless current_user.email_verified
+    else
+      redirect_to root_path
```
jfederico: Why are we sending the user to root?, shouldn't they just stay where they were?

app/controllers/rooms_controller.rb
```diff
@@ -240,7 +242,16 @@ def validate_accepted_terms
...
+    end
+  end
+
+  def verify_room_owner_verified
+    unless @room.owner.email_verified
+      flash[:alert] = t("room.unavailable")
+      redirect_to root_path
```
jfederico: Same here, why to send the user to root?

app/controllers/account_activations_controller.rb
farhatahmad: Based on its configuration, Rubocop currently forces us to include it at the top of every file. 

A good explanation why can be found here: https://stackoverflow.com/questions/37799296/what-does-the-comment-frozen-string-literal-true-do

app/controllers/rooms_controller.rb
farhatahmad: In the case where someone (who is not signed in) attempts to make an http request (to any action other than show or join), I believe that they should be redirected to the root path. Maybe it would be better if we add a flash informing the user that they need to sign in before they access that functionality

app/controllers/rooms_controller.rb
farhatahmad: Would it be preferred to send the user to their main room with the current alert?

PR #371, opened 2019-02-14
GRN-10: Added an indication that a recording is processing

farhatahmad:


farhatahmad:
Various issues have appeared so this current implementation needs to be reworked

PR #403, opened 2019-03-21
Added the administrator role and functionality that comes with it 

farhatahmad:


app/assets/javascripts/admins.js
```diff
@@ -0,0 +1,77 @@
...
+          users_found = users_found + 1;
+          $(this).show();
+        }
+        else{
+          $(this).hide();
+        }
+      });
+
+      // Show "No users match your search" if no users found
+      console.log(users_found)
```
jfederico: console.log should not be included in production

app/assets/javascripts/admins.js
```diff
@@ -0,0 +1,77 @@
...
+      if(users_found === 0){
+        console.log("Showing")
+        $('#no_users_found').show();
+      }
+      else{
+        console.log("Hiding")
```
jfederico: Same

app/assets/javascripts/admins.js
```diff
@@ -0,0 +1,77 @@
...
+        }
+        else{
+          $(this).hide();
+        }
+      });
+
+      // Show "No users match your search" if no users found
+      console.log(users_found)
+      if(users_found === 0){
+        console.log("Showing")
```
jfederico: Same

app/assets/javascripts/settings.js
```diff
@@ -20,7 +20,7 @@ $(document).on('turbolinks:load', function(){
   var action = $("body").data('action');
 
   // Only run on the settings page.
-  if ((controller == "users" && action == "edit") || (controller == "users" && action == "update")){
+  if ((controller == "users" && action == "edit") || (controller == "users" && action == "update") || (controller == "admins" && action == "index")){
```
jfederico: Was this fixed somewhere else? I have the idea that you had found a better way of filtering this

app/assets/javascripts/admins.js
farhatahmad: Yeah I caught this in one of the other PRs and must have forgot to do it here as well. Good catch

app/assets/javascripts/admins.js
farhatahmad: Done

app/assets/javascripts/settings.js
farhatahmad: As far as I know, this is the simplest & cleanest way to prevent js from running on every page. In this case the code is very specific to the settings pages so allowing it to work on any page would not be ideal

PR #425, opened 2019-03-28
GRN-15: Added the ability to change color and image from admin interface

farhatahmad:


lib/tasks/administrator.rake
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+namespace :admin do
+  task create: :environment do
+    # Create administrator account if it doesn't exist
+    unless User.where(email: "admin@admin.com").exists?
+      admin = User.create(name: "Administrator", email: "admin@admin.com", password: "administrator",
```
jfederico: The task should receive parameters to allow system administrators to create their own account if they want. If non is provided then we use the example as default

PR #426, opened 2019-03-28
Fixed #398 Fixed issue with recording length and sort (GRN-83)

### LINKED ISSUE #398: Show length in minutes
Currently, we're showing meetings in hours and minutes

image

It would be easier to read if all the length was minutes and sorting.  This would also correct an error when sorting

image

farhatahmad:
Fixes #398


app/assets/javascripts/sort.js
```diff
@@ -18,13 +18,13 @@ $(document).on('turbolinks:load', function(){
   var controller = $("body").data('controller');
   var action = $("body").data('action');
 
-  if(controller == "rooms" && action == "show" || controller == "rooms" && action == "update"){
+  if(controller == "rooms" && action == "show" || controller == "rooms" && action == "update" || controller == "users" && action == "recordings"){
```
jfederico: The only thing I am not sure about is this if condition. Does this mean that everytime we use the sorting we are going to add references to the place we are going to use it? It would not be berrer to have a function that is calles only when needed?

farhatahmad: Good point. Code will be changed so that the code will run any time there exists a table on the view

PR #427, opened 2019-03-29
Ldap authentication CSS customization

rafaeelaudibert:
This PR has the intention to be able to customize the current LDAP page, which comes with the default CSS by the Omniauth and Omniauth-Ldap gem.


rafaeelaudibert:
For demonstration purposes, this is the old LDAP view login modal:
Old LDAP

And this is the new LDAP authentication page
New LDAP

For comparison purposes, this is the default login modal
Default modal

jfederico:
My big concern with this one is that we are bringing the view to the initializer. If there is no way to style the UI from the view or extend the gem, it may be cleaner to add the gem as a library and apply the changes to the UI there.
The gem is not updated anyways, we are looking into using a fork or forking ourselves.

rafaeelaudibert:
The biggest problem on this is that I needed to extend **2** gems to make this work. I really think that using a maintained fork is a better option than this, but it would work as a MonkeyPatch while the 'use fork/create fork' problem is not solved/decided.

jfederico:
There is a new way for customizing the view through the template instead of breaking the MVC.
Closing this PR

PR #429, opened 2019-03-30
Added Google calendar button

rafaeelaudibert:
Added a new "Add to Google Calendar" button in the Room#show view, which easily enables someone to create a reminder for a future event in the Google Calendar.

The Google Calendar Button can be enabled/disabled in the .env file. The env.sample file has it enabled by default.

The layout is as follow:
Layout

As you can see in the former image, the localization for other languages needs to be added later.

This PR also address a issue in the "Copy" button. After clicking on it, it was hardcoded in the JS file to switch the text to "Copy", but now it correctly uses the I18n translations. I've also added a new translation "Copied" which must me added in the other locales as well. The only against is that I needed to change the room.js file to room.js.erb, but I assume it won't use more server resources as it is only a small rendering.


ffdixon:
Thanks for this pull request!  I think others will enjoy having this feature.  Could we get you to sign a contributor's agreement before we review this request?  See

  http://docs.bigbluebutton.org/support/faq.html#contributing-to-bigbluebutton

jfederico:
The buttons in the way they are look kind of wonky. I am forseing the situation in whcih we need to add a new one. It may be better to change the design to something more compact. And also, it should be configurable as not everybody would like to have this feature.
I am going to put it on hold as it may be bnetter to wait until we have the feature manager in 2.1

rafaeelaudibert:
> The buttons in the way they are look kind of wonky. I am forseing the situation in whcih we need to add a new one. It may be better to change the design to something more compact. 

They have all the mobile-friendly classes, so they would arrange themselves without most changes when the need for more buttons arrived, just dividing the space which I left for them (we could add up to 5 buttons, but I agree we would need to make the text smaller, because it would wrap to the next line, it is a problem in this case).

> And also, it should be configurable as not everybody would like to have this feature.

It is configurable, actually, through the flag `ENABLE_GOOGLE_CALENDAR_BUTTON = true|false` in the `.env` file, which I was considering myself as an eficcient way to manage the features, but if we will have a more properly way to enable/disable features, I agree with it.

sample.env
```diff
@@ -139,6 +139,9 @@ PAGINATION_NUMBER=25
 # Default is set to 10 rows
 NUMBER_OF_ROWS=10
 
+# Specify if you want to display the Google Calendar button
+ENABLE_GOOGLE_CALENDAR_BUTTON=true
```
jfederico: This should be disabled by default

rafaeelaudibert: Agreed.

rafaeelaudibert: Maybe it could have been written it defaults to false, but now it's been merged I think it's okay to keep it this way.

PR #433, opened 2019-04-03
GRN-88: Fixed some things that Scrutinizer was complaining about

farhatahmad:


app/helpers/users_helper.rb
```diff
@@ -17,4 +17,8 @@
 # with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
 
 module UsersHelper
+  # Returns the link the user needs to click to verify their account
+  def user_verification_link
```
rafaeelaudibert: Helpers are supposed to provide methods to be used in the views, not exclusively in the controllers.

I think the `Concerns` folder is more appropriate for this use case.

farhatahmad: Good catch. Moved the code to a controller concern

PR #436, opened 2019-04-05
Added mailer i18n

rafaeelaudibert:
This PR adds I18n to the Greenlight default mailers.

English is the only locale already filled, so we must add this fields to be translated on Transifex


rafaeelaudibert:
I tried to translate the email at [room.js:46](https://github.com/bigbluebutton/greenlight/blob/master/app/assets/javascripts/room.js#L46), but the file is cached and wouldn't work correctly.

Any idea how we could make so? I thought about creating a controller which would render the mailer and return to the JS code through AJAX, what do you think?

farhatahmad:
> I tried to translate the email at [room.js:46](https://github.com/bigbluebutton/greenlight/blob/master/app/assets/javascripts/room.js#L46), but the file is cached and wouldn't work correctly.
> 
> Any idea how we could make so? I thought about creating a controller which would render the mailer and return to the JS code through AJAX, what do you think?

Have you tried converting the file to a js.erb extension instead? That way, you should be able to include the translations directly into the js file. If not, then yeah I think AJAX is your best call

In the future, we may want to look at converting in to a mailer that way the format can be more consistent with the other emails we send

rafaeelaudibert:
> Have you tried converting the file to a js.erb extension instead? That way, you should be able to include the translations directly into the js file.

It won't work to convert to a .js.erb file, because as I said, Rails asset pipeline renders it only once and caches it. This way it will be rendered with the first person which access that page locale, and stay with it until we reset the cache

If we inlined the JS in the .html.erb view, it would work actually, but I don't know if this is aligned with the rest of the code.

If we would to use a Mailer, would we remove the ability to customize the email, or would we open a modal where the user could customize it, and after we send a POST to the server to send the email with the data provided by the user?

farhatahmad:
@rafaeelaudibert You can follow the link below to get localization into js files:

https://stackoverflow.com/questions/2701749/rails-internationalization-of-javascript-strings

I've tested it on my local and it seems to work without any issues

rafaeelaudibert:
I wouldn't personally like to expose all the locales to the final user, but reading the top voted answer comments, I see how I can only expose the mailer ones, which will probably make it fine.

Thanks for the finding! Will implement it soon

rafaeelaudibert:
Added a new entry to the en.yml file, on respect to the locales which are available in the Javascript. The `function t(key)` is available in all Greenlight pages. While we don't create the new keys in the locales files, an empty string is returned by default.

config/locales/en.yml
```diff
@@ -81,6 +81,12 @@ en:
   invalid_credentials: Login failed due to invalid credentials. Are you sure you entered them correctly?
   invalid_login_method: Login failed due to account mismatch. You need to log in with omniauth.
   invite_message: "To invite someone to the meeting, send them this link:"
+  javascript:
+    room:
+      mailer:
+        subject: 'has invited you to view a recording.'
+        body: 'To view the recording, follow the link below:\n\n'
```
farhatahmad: I think it might be better to move the \n to the javascript file (since it doesn't need to be translated)

config/locales/en.yml
```diff
@@ -81,6 +81,12 @@ en:
...
+        footer: '\n\nThis e-mail is auto-generated by BigBlueButton.\nBigBlueButton is an open source web conferencing system. For more information on BigBlueButton, see https://bigbluebutton.org/.'
```
farhatahmad: Same as above. You might need to split these into 2 separate entries

rafaeelaudibert: Agreed

config/locales/en.yml
rafaeelaudibert: Agreed

PR #470, opened 2019-04-18
Added I18n to recording types

rafaeelaudibert:
Added I18n to the recording types.

Although we currently only have "Presentation" type for the Recordings displayed in the rooms recordings table, in the future we might have more, so a new "array" was created in the locale files, namely "recording.format"


rafaeelaudibert:
This PR shouldn't contain the code created on the fork mconf/greenlight, so a new PR will be open

PR #496, opened 2019-05-03
Admin panel

jfederico:


app/assets/javascripts/settings.js
```diff
@@ -20,7 +20,7 @@ $(document).on('turbolinks:load', function(){
   var action = $("body").data('action');
 
   // Only run on the settings page.
-  if ((controller == "users" && action == "edit") || (controller == "users" && action == "update")){
+  if ((controller == "users" && action == "edit") || (controller == "users" && action == "update") || (controller == "admins" && action == "index")){
```
jfederico: This logic should not be in the client. The authorization should be implemented in the Server and passed from the controller to the view as a variable, perhaps in an object.

PR #513, opened 2019-05-08
Refactor users.html to only make one db call to get roles (GRN2-121)

shawn-higgins1:


app/views/shared/admin_settings/_users.html.erb
```diff
@@ -68,15 +68,16 @@
                     <td><%= user.email && user.email != "" ? user.email : user.username%></td>
                     <td><%= user.provider %></td>
                     <td class="text-center">
-                      <% if user.has_role? :denied %>
+                    <% @roles = user.roles() %>
```
farhatahmad: You might want to do user.roles.pluck(:name) here

app/views/shared/admin_settings/_users.html.erb
```diff
@@ -68,15 +68,16 @@
...
+                      <% if @roles.any?{|r| r[:name] == "denied"}%>
```
farhatahmad: And then replace this with @roles.include?

PR #534, opened 2019-05-21
GRN2-140: Added Email to inform user they've been promoted

shawn-higgins1:


config/locales/en.yml
```diff
@@ -189,6 +189,16 @@ en:
...
         ignore: You can safely ignore this email if you did not request a password reset.
+      promoted:
+        admins_link: Visit the Organization Page
+        info: You are now an administrator on %{url}.
+        more-info: As an administrator, you can manage users, their role and configure your site settings by changing the logo and brand colour.
+        subtitle: Administrator Rights Granted
+      demoted:
+        root_link: Visit the Organization Page
+        info: You are no longer an administrator on %{url}.
+        more-info: You now have the same privileges as a regular user and will no longer be able to configure the site settings.
```
farhatahmad: I think it might be a better idea to replace this with:
"You now have the same privileges as a regular user and will no longer be able to access any of the Administrator settings."

PR #538, opened 2019-05-21
GRN2-6: Notify admins when a approve/invite user signs up

shawn-higgins1:


app/controllers/concerns/emailer.rb
```diff
@@ -41,13 +41,32 @@ def send_user_approved_email(user)
     UserMailer.approve_user(user, root_url, logo_image, user_color).deliver_now
   end
 
+  def send_approval_user_signup_email(user)
+    UserMailer.approval_user_signup(user, root_url, logo_image, user_color, admin_emails).deliver_now
```
farhatahmad: You might want to replace root_url with admins_url, so that the admins can click the link and go directly to their Organization Settings

app/controllers/concerns/emailer.rb
```diff
@@ -41,13 +41,32 @@ def send_user_approved_email(user)
...
+  end
+
+  def send_invite_user_signup_email(user)
+    UserMailer.invite_user_signup(user, root_url, logo_image, user_color, admin_emails).deliver_now
```
farhatahmad: Same as above

PR #541, opened 2019-05-22
GRN2-118: Create a setting to require authentication to join a room

### LINKED ISSUE #526: Restrict rooms to only allow authenticated users to joi
Currently, an invite URL allows any user to join.  

There are some use cases where you want a meeting to occur with only users who have logged into their Green Light account.

shawn-higgins1:
fixes #526


app/controllers/rooms_controller.rb
```diff
@@ -98,6 +98,9 @@ def update
 
   # POST /:room_uid
   def join
+    # If this setting is turned on only authenticated users are allowed to join rooms
+    return if room_authentication_required && current_user.nil?
```
farhatahmad: I think it might be a better idea to redirect the user to the sign in page with a flash saying "You must be logged in to join the room"

app/views/shared/admin_settings/_site_settings.html.erb
```diff
@@ -28,6 +28,19 @@
       </div>
     </div>
   </div>
+  <div class="row">
+    <div class="col-12">
+      <div class="mb-7 form-group">
+        <label class="form-label"><%= t("administrator.site_settings.authentication.title") %></label>
+        <label class="form-label text-muted"><%= t("administrator.site_settings.authentication.info") %></label>
+        <div class="row gutters-xs">
```
farhatahmad: You should use the tabler checkbox instead of just the standard checkbox. You can see how here: https://preview.tabler.io/form-elements.html

spec/controllers/admins_controller_spec.rb
```diff
@@ -226,5 +226,22 @@
         expect(response).to redirect_to(admins_path)
       end
     end
+
+    context "POST #room_authentication" do
+      it "changes the primary on the page" do
```
farhatahmad: Make sure you update this to match what you are testing

spec/controllers/rooms_controller_spec.rb
```diff
@@ -206,6 +206,14 @@ def random_valid_room_params
       expect(flash[:alert]).to be_present
       expect(response).to redirect_to(root_path)
     end
+
+    it "should not allow the user to join if the user isn't signed in and room authentication is required" do
+      allow_any_instance_of(Setting).to receive(:get_value).and_return("true")
+
+      post :join, params: { room_uid: @room }
+
+      expect(response).to render_template(:join)
```
farhatahmad: This doesn't actually test the a join used can't join a room. It just test if a join template will render (which it will regardless of what the setting is set to). This can be fixed once you make the change to the redirect

config/routes.rb
```diff
@@ -39,6 +39,7 @@
   scope '/admins' do
     post '/branding', to: 'admins#branding', as: :admin_branding
     post '/coloring', to: 'admins#coloring', as: :admin_coloring
+    post '/roomAuthentication', to: 'admins#room_authentication', as: :admin_room_authentication
```
farhatahmad: roomAuthentication should be snake case (instead of camelCase)

PR #570, opened 2019-06-18
Room creation auth

madumlao:
This PR adds a preference setting for which users are allowed to create rooms.

By default it preserves current behavior in that anyone authorized is allowed to create rooms. However, it also adds the option that only users with the admin (and super_admin) role create rooms.


jfederico:
Same as the other PR. Can you clean up the code first and make sure to pass Travis?

farhatahmad:
This has been solved by #607. An Admin can disable the Create a Room block by setting the Number of Rooms per User to 1 in the Site Settings

PR #571, opened 2019-06-20
updated application version code as env

jineshb:


jfederico:
@jineshb Just to be consistent with other projects we have already, I would like the files to be renamed to 

cloudbuild.yaml for production
cloudbuild-staging.yaml for staging
cloudbuild-dev.yaml for development (if it is ever needed)
cloudbuild-WHATEVERYOULIKE.yaml (for anything else)

jineshb:
@jfederico files names have been changed for consistency.

PR #582, opened 2019-06-25
Improve container build: GRN2-156

shawn-higgins1:
I believe the documentation for Greenlight should be updated. 

First to make a change to the .env file you simply need to do a docker-compose down and then a docker-compose up. 

Second to apply a code change you need to run the image build script and then do docker-compose up. You shouldn't delete the pre-existing container as that removes the caches which speed up the build process.


scripts/image_build.sh
```diff
@@ -56,7 +56,13 @@ if [ "$CD_REF_NAME" != "master" ] && [[ "$CD_REF_NAME" != *"release"* ]] && ( [
 fi
 
 # Include sqlite for production
-sed -i "/^group :production do/a\ \ gem 'sqlite3', '~> 1.3.6'" Gemfile
+sqliteCount="$(grep "gem 'sqlite3'" Gemfile | wc -l)"
+
+if [ $sqliteCount -lt 2 ]; then
+  echo "test"
```
farhatahmad: Forgot to remove some debugging code

shawn-higgins1: No that's necessary otherwise each time you run the build script you add the gem. It should only be added once.

PR #602, opened 2019-07-02
GRN2-166: Fixed issue with language fallback not working (Fixed #593)

### LINKED ISSUE #593: Missing translations causing odd UI bug
Whenever there are missing translations, it results in an odd visual bug that makes the application difficult to use. 

In the case where there are the missing translations, the string should fallback to the English version.

Examples can be found here:
https://groups.google.com/forum/#!topic/bigbluebutton-dev/OwjRAC4Yo9U

farhatahmad:
Fixes #593


jfederico:
This PR fixes two bugs at once, although only the one refered in the title is addressed

PR #606, opened 2019-07-03
GRN2-30: Add custom ldap signin

shawn-higgins1:
Creates a custom sign in page for LDAP and allows LDAP auth to be used in conjunction with other providers


app/controllers/sessions_controller.rb
```diff
@@ -100,4 +102,97 @@ def passes_invite_reqs
...
+      return redirect_to root_path, flash: { success: I18n.t("registration.approval.signup") }
+    end
+
+    send_invite_user_signup_email(user) if Rails.configuration.enable_email_verification &&
+                                           invite_registration && !@user_exists
+
+    login(user)
+  end
+
+  def send_ldap_request
```
farhatahmad: I'm not quite sure that this is the right place to put this code. Maybe move it into a concern? or some sort of module?

app/controllers/sessions_controller.rb
```diff
@@ -45,32 +54,25 @@ def create
     login(user)
   end
 
+  def ldap
+    result = send_ldap_request
+
+    if result
+      result = result.first
+    else
+      redirect_to(ldap_signin_path, alert: I18n.t("invalid_credentials"))
```
farhatahmad: You can do these 2 lines in one: return && redirect_to...

app/controllers/sessions_controller.rb
```diff
@@ -100,4 +102,97 @@ def passes_invite_reqs
...
+    send_invite_user_signup_email(user) if Rails.configuration.enable_email_verification &&
+                                           invite_registration && !@user_exists
+
+    login(user)
+  end
+
+  def send_ldap_request
+    host = ''
+    port = 389
+    bind_dn = ''
```
farhatahmad: I'm not quite sure that this is the right place to put this code. Maybe move it into a concern? or some sort of module?

app/controllers/sessions_controller.rb
```diff
@@ -45,32 +54,25 @@ def create
...
 
+  def ldap
+    result = send_ldap_request
+
+    if result
+      result = result.first
+    else
+      redirect_to(ldap_signin_path, alert: I18n.t("invalid_credentials"))
+      return
+    end
```
farhatahmad: You can do these 2 lines in one: return redirect_to...

PR #615, opened 2019-07-08
GRN2-155: Begin preparing for removal of Twitter accounts

shawn-higgins1:


app/controllers/sessions_controller.rb
```diff
@@ -51,6 +51,10 @@ def omniauth
       @auth = request.env['omniauth.auth']
       @user_exists = check_user_exists
 
+      if !@user_exists && @auth['provider'] == "twitter"
+        return redirect_to root_path, flash: { alert: I18n.t("registration.deprecated.twitter_signup") }
+      end
```
farhatahmad: This can be done in 1 line (see the line below it)

app/helpers/sessions_helper.rb
```diff
@@ -19,6 +19,24 @@
 module SessionsHelper
   # Logs a user into GreenLight.
   def login(user)
+    if !session["old_twitter_user_id"].nil? && user.provider != "twitter"
+      old_user = User.find(session["old_twitter_user_id"])
+
```
farhatahmad: I would suggest moving this code into its own private method

app/controllers/sessions_controller.rb
shawn-higgins1: Then it goes over the line column limit

farhatahmad: Hm. Good point. In that case I would leave it as is

PR #619, opened 2019-07-09
GRN2-30: Add custom ldap sign in page

shawn-higgins1:


Gemfile
```diff
@@ -47,6 +47,8 @@ gem 'omniauth-google-oauth2'
 gem 'omniauth-bn-office365', git: 'https://github.com/blindsidenetworks/omniauth-azure-ad.git'
 gem 'omniauth-ldap'
 gem 'omniauth-bn-launcher', '~> 0.1.0'
+gem 'bn-ldap-authentication', git: 'https://github.com/blindsidenetworks/bn-ldap-authentication.git'
+gem 'net-ldap'
```
jfederico: @shawn-higgins1 do we need net-ldap as part of Greenlight? can it be part of the dependences declared in the gem itself?

PR #620, opened 2019-07-09
GRN2-192: Made error pages more descriptive

farhatahmad:


jfederico:
@farhatahmad second thoughts on this one.

GL is localized in so many languages and we give support to all of them through the mailing list (sam as commercially we and others support multiple languages), if the strings for those error messages are translated, it is hard for the people giving support to figure out what happened from a message that is translated.

We have to add a code to every one of these friendly messages (same we do with the HTML5 client) so we can lookup by number in the en.yml file when the end users report the issues.

PR #650, opened 2019-07-15
GRN2-202: Create a room in which everyone joins as Moderator (Fixed #628)

### LINKED ISSUE #628: Create a room in which everyone joins as Moderator
Currently, the invite URL only allows the owner to join as a moderator.  However, there are cases where you want an open room that everyone joins as a moderator (that way, if the owner of the room is away, users can still meet).

Requested in https://groups.google.com/d/msg/bigbluebutton-dev/DV5RXPo5ty4/91IQmLUGCQAJ

shawn-higgins1:
fixes #628


sample.env
```diff
@@ -130,7 +130,7 @@ RELATIVE_URL_ROOT=/b
 # Current settings available:
 #   default-client: Room owners can decide between the Flash Client and the HTML5 Client for a room
 #   mute-on-join: Automatically mute users by default when they join a room
-ROOM_FEATURES=default-client,mute-on-join
+ROOM_FEATURES=default-client,mute-on-join,anyone-can-start
```
farhatahmad: Add a description at the top explaining what the setting does

PR #652, opened 2019-07-16
update ru_RU

RasimKalimullin:


farhatahmad:
Hi Rasim, 

Thank you for taking the time to update the Russian translations. 

We use Transifex to manage all of the languages used in Greenlight. Please add your translations for Greenlight here: https://www.transifex.com/bigbluebutton/greenlight/

PR #662, opened 2019-07-19
GRN2-129: Added server recordings and refactored adminsitrator panel

### LINKED ISSUE #668: Email Recording from All Recordings page does nothing
Link "Email Recording" from page "All Recordings" does nothing.
Tested in Firefox and Chrome (latest).
No errors in browser's dev console.

From "Room Recordings" this link works, but not as expected (see #666).

farhatahmad:
fixes #668


app/controllers/admins_controller.rb
```diff
@@ -40,6 +41,29 @@ def index
     @pagy, @users = pagy(user_list)
   end
 
+  # GET /admins/site_settings
+  def site_settings
+  end
+
+  # GET /admins/server_recordings
+  def server_recordings
+    server_rooms = []
```
shawn-higgins1: Would it be easier to just do Room.includes(:owner).where(users: { provider: 'user1' }).pluck(:bbb_id) for multitenant and for single tenant just not pass a meeting id in the getRecordings call?

app/controllers/concerns/recorder.rb
```diff
@@ -59,7 +98,8 @@ def filter_recordings(api_res, search)
                ((r[:metadata].nil? || r[:metadata][:name].nil?) &&
                  r[:name].downcase.include?(search)) ||
                r[:participants].include?(search) ||
-               !r[:playbacks].select { |p| p[:type].downcase.include?(search) }.empty?
+               !r[:playbacks].select { |p| p[:type].downcase.include?(search) }.empty? ||
+               Room.find_by(bbb_id: r[:meetingID]).owner.email.downcase.include?(search)
```
shawn-higgins1: Will this mean that if you search for recordings under a room then it will select recordings based on the owner email even though that isn't displayed in the UI?

app/controllers/admins_controller.rb
farhatahmad: Fixed

app/controllers/concerns/recorder.rb
farhatahmad: Fixed

app/controllers/concerns/recorder.rb
```diff
@@ -59,7 +98,8 @@ def filter_recordings(api_res, search)
                ((r[:metadata].nil? || r[:metadata][:name].nil?) &&
                  r[:name].downcase.include?(search)) ||
                r[:participants].include?(search) ||
-               !r[:playbacks].select { |p| p[:type].downcase.include?(search) }.empty?
+               !r[:playbacks].select { |p| p[:type].downcase.include?(search) }.empty? ||
+               (Room.find_by(bbb_id: r[:meetingID]).owner.email.downcase.include?(search) && search_name)
```
shawn-higgins1: You might want to flip around the and statement. That way if search_name is false it won't do the search by room owner which should improve the query speed.

farhatahmad: Done

PR #701, opened 2019-07-26
GRN2-xx: General bug fixes and improvements for 2.3

farhatahmad:


app/helpers/recordings_helper.rb
```diff
@@ -36,7 +36,7 @@ def recording_length(playbacks)
     elsif len.zero?
       "< 1 min"
     else
-      "#{len} min"
+      "#{len} mins"
```
jfederico: This should be localized, is it?

farhatahmad: Yes correct. I'm going to remove it from this PR and create another PR to fix it and other localization issues

PR #733, opened 2019-08-12
Configurable default registration

shawn-higgins1:


config/application.rb
```diff
@@ -131,7 +131,15 @@ class Application < Rails::Application
     config.primary_color_darken_default = "#316cbe"
 
     # Default registration method if the user does not specify one
-    config.registration_method_default = config.registration_methods[:open]
+    config.registration_method_default = if ENV["DEFAULT_REGISTRATION"] == "open"
```
farhatahmad: No need for the first if. Just check if invite elsif approval else open

PR #754, opened 2019-08-20
GRN2-164: DON'T MERGE: Change Greenlight default database to Postgres

shawn-higgins1:


shawn-higgins1:
I've tested this script using the demo Greenlight sqlite database both inside and outside of a docker container and it appears to work.

PR #758, opened 2019-08-21
GRN2-229: Added a Maintenance Window flash

farhatahmad:


config/application.rb
```diff
@@ -118,6 +118,10 @@ class Application < Rails::Application
 
     config.google_analytics = ENV["GOOGLE_ANALYTICS_TRACKING_ID"].present?
 
+    # MAINTENANCE
+    config.maintenance_window = ENV["MAINTENANCE_WINDOW"]
```
jfederico: What if this environment variable is empty? Is it going to show an empty flash message box?

farhatahmad: No, because the check in the Application Controller is `Rails.configuration.maintenance_window.present?` which returns false if the it is blank

PR #769, opened 2019-08-27
Update it.yml

andcastellani:
I've modified some wrong words:
"BigBlueButton" was translated in "GrandeBottoneBlu" (but it daesn't need to be translated)
"GreenLight" was translated in "LuceVerde" (but it daesn't need to be translated)
"DENTAL LIVE Manager" now is translated with "BigBlueBUtton"


andcastellani:
Ok, I've submitted my join request for the Italian (it) language in 
Trasifex localization of Greenlight.

Andrea

Il 27/08/19 16:49, farhatahmad ha scritto:

PR #781, opened 2019-09-03
Translations en yml  master de de

abdishakur54:
tijabo


farhatahmad:
Hi abdishakur,

Thank you for taking the time to update the German translations.

PR #820, opened 2019-10-02
typo fix and auto_join translation

alfonsofernandezyf:
fixed typo ls -> la
auto_join translation was not accurate.


alfonsofernandezyf:
Hi, I wanted to correct  a typo in both spanish locales and fix auto_join translation which is not accurate.

Both locales need to be changed, es.yml and es_MX.yml

I didn't know hoy to add the other change to this pull request.

farhatahmad:
Hi @alfonsofernandezyf,

Thank you for taking the time to update the Spanish translations.

PR #872, opened 2019-11-18
from server

ssh713:


farhatahmad:
These changes should be made to the forked directory (origin), not the upstream one.

You've also exposed your DockerHub password, so I suggest you change that immediately.
(I removed the file that contained it from the PR)

PR #945, opened 2020-01-31
GRN2-266: Autofocus room name when create room modal appears

etiennevvv:


app/assets/javascripts/room.js
```diff
@@ -127,6 +127,9 @@ $(document).on('turbolinks:load', function(){
 });
 
 function showCreateRoom(target) {
+  $('#createRoomModal').on('shown.bs.modal', function () {
```
farhatahmad: Good implementation, but not the right place to add it. Might want to move it to below the `#create-room-block` click handler

farhatahmad:
@etiennevvv Might also be worth discussing whether we want this to happen for both the Create Room and the Update Room Settings (since they use the same modal). If we do, the current implementation is fine. If not, we'll need to make additional checks before focusing on the input

etiennevvv:
@farhatahmad Ok. Fixed it for now so that the autofocus only appears when creating the room. I don't think it makes much sense to have the autofocus when updating the room name but can discuss this later.

PR #951, opened 2020-02-10
GRN2-264: Allow users to change the room settings on Home Room

etiennevvv:


app/views/rooms/components/_room_block.html.erb
```diff
@@ -45,8 +41,7 @@
             <% end %>
           </div>
         </td>
-        <td class="text-right">
-          <% unless room == current_user.main_room %>
+        <td class="text-right">          
             <div class="item-action dropdown" data-display="static">
```
farhatahmad: Inconsistent spacing here

app/views/rooms/show.html.erb
```diff
@@ -77,11 +77,9 @@
     </div>
 
     <div id="room_block_container" class="row pt-7 pb-5">
-      <% if current_user.rooms.length > 1 %>
         <div class="col-lg-4 col-md-6 col-sm-12">
           <%= link_to current_user.main_room do %>
-            <%= render "rooms/components/room_block", room: current_user.main_room %>
-          <% end %>
+            <%= render "rooms/components/room_block", room: current_user.main_room %>
```
farhatahmad: Missing an `<% end %>` here

app/views/rooms/show.html.erb
```diff
@@ -77,21 +77,19 @@
...
+      </div>
+      <% current_user.secondary_rooms.each do |room| %>
         <div class="col-lg-4 col-md-6 col-sm-12">
-          <%= link_to current_user.main_room do %>
-            <%= render "rooms/components/room_block", room: current_user.main_room %>
+          <%= link_to room do %>
+            <%= render "rooms/components/room_block", room: room %>
           <% end %>
         </div>
-        <% current_user.secondary_rooms.each do |room| %>
```
farhatahmad: Just noticed that this is doing a separate render for main room, and then looping through the secondary rooms. We can probably join this into 1 statement `<% current_user.rooms.each do |room| %>` and then remove the secondary rooms method since its not used anywhere else

PR #953, opened 2020-02-12
GRN2-269: Pulling room settings using GET instead of data-attributes

etiennevvv:


app/controllers/rooms_controller.rb
```diff
@@ -248,6 +248,14 @@ def shared_users
     end
   end
 
+  # GET /:room_uid/room_settings
+  def room_settings
+    # Respond with JSON object of users that have access to the room
```
farhatahmad: Forgot to change the comment

farhatahmad:
Might also be a good idea to write a test case or two that tests and makes sure the room_settings method works correctly

spec/controllers/rooms_controller_spec.rb
```diff
@@ -180,6 +180,22 @@ def random_valid_room_params
...
+      @request.session[:user_id] = @owner.id
+      name = Faker::Games::Pokemon.name
+
+      room_params = { name: name, "mute_on_join": "1",
+        "require_moderator_approval": "1", "anyone_can_start": "1", "all_join_moderator": "1" }
+
+      json_room_settings = "{\"muteOnStart\":true,\"requireModeratorApproval\":true," \
+        "\"anyoneCanStart\":true,\"joinModerator\":true}"
+
+      post :create, params: { room: room_params }
```
farhatahmad: You should try avoiding make 2 HTTP requests in a single test case. If you want to create a room, you can use the room factory (in `factories.rb`). Once you create the room, you can set the room_settings on that object

PR #959, opened 2020-02-19
GRN2-233: Made account activation & password reset links based on tokens only

etiennevvv:


app/controllers/account_activations_controller.rb
```diff
@@ -60,14 +60,14 @@ def resend
...
+    @user = User.find_by!(activation_digest: User.digest(params[:token]), provider: @user_domain)
   end
 
   def ensure_unauthenticated
     redirect_to current_user.main_room if current_user
   end
 
   def email_params
-    params.require(:email).permit(:email, :token)
+    params.require(:activation_digest).permit(:activation_digest, :token)
```
farhatahmad: The only param that is required is the token param. You can remove the activation_digest param

app/models/user.rb
```diff
@@ -121,8 +121,7 @@ def create_reset_digest
   # Returns true if the given token matches the digest.
   def authenticated?(attribute, token)
     digest = send("#{attribute}_digest")
-    return false if digest.nil?
```
farhatahmad: You need to keep the explicit false return if the digest is nil (in case other places are explicitly checking if the return value is false

PR #961, opened 2020-02-24
GRN2-273: Fixed bug where incorrect recording length returned when "Notes" format was included

etiennevvv:


app/helpers/recordings_helper.rb
```diff
@@ -24,18 +24,13 @@ def recording_date(date)
...
-
-    len = valid_playbacks.first[:length]
-    if len > 60
-      "#{(len / 60).to_i} h #{len % 60} min"
-    elsif len.zero?
-      "< 1 min"
-    else
-      "#{len} min"
+    # Looping through playbacks array and returning first non-zero length value
+    playbacks.each do |i|
```
farhatahmad: Avoid using i and use a more meaningful variable name

PR #969, opened 2020-03-01
Update en.yml

osbornhodges2000:


ffdixon:
Thanks for the suggestion, but I think we'll keep the product named Greenlight.  :-)

PR #995, opened 2020-03-10
GRN2-283: Fixed bug where users can't join a shared room (Fixed #989)

### LINKED ISSUE #989: shared room issue
BigBlueButton 2.2.0 release
Greenlight 2.5 release

2 user ： super admin  user A + normal user B 
A create a room_A and share to user B
room_A set anybody can start the meeting

if anonymous C (anybody not sign in)  start room_A , B can never join room_A anymore.

so, B must be the first to enter(start)  shared room_A.

etiennevvv:
fixes #989


app/controllers/rooms_controller.rb
```diff
@@ -93,7 +93,7 @@ def join
     return redirect_to root_path,
       flash: { alert: I18n.t("administrator.site_settings.authentication.user-info") } if auth_required
 
-    unless @room.owned_by?(current_user)
+    if current_user && !(@room.owned_by?(current_user) || room_shared_with_user)
```
farhatahmad: You don't need to check for current_user here since .owned_by? and .shared_by? both return false if the current_user is nil

PR #998, opened 2020-03-11
GRN2-258: Added the ability to select a time zone

etiennevvv:


app/controllers/users_controller.rb
```diff
@@ -202,7 +202,7 @@ def check_user_signup_allowed
 
   def user_params
     params.require(:user).permit(:name, :email, :image, :password, :password_confirmation,
-      :new_password, :provider, :accepted_terms, :language)
+      :new_password, :provider, :accepted_terms, :language, :time_zone)
```
farhatahmad: Not needed

app/assets/javascripts/user.js
```diff
@@ -0,0 +1,28 @@
...
+//
+// BigBlueButton is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+// PARTICULAR PURPOSE. See the GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License along
+// with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
+
+$(document).on('turbolinks:load', function() {
+    $.rails.refreshCSRFTokens();
```
farhatahmad: What's the purpose of this?

farhatahmad:
@etiennevvv Can you squash the 7 commits that have the same name into just 1 commit? The other 2 are fine

app/controllers/users_controller.rb
etiennevvv: Need it there to update the time_zone attribute

farhatahmad:
@etiennevvv Can you also add some tests for updating timezone in the user profile?

https://github.com/bigbluebutton/greenlight/blob/master/spec/controllers/users_controller_spec.rb#L281

PR #1015, opened 2020-03-17
Fixes to allow me to use Greenlight

### LINKED ISSUE #1012: Auto-Approve LDAP Signups
I set up greenlight so that it requires approval for a new account sign up. (DEFAULT_REGISTRATION=approval) Now, the bulk of our users already has an account in our LDAP server, so I connected Greenlight to our LDAP server and "Sign up using LDAP" works fine except for needing admin approval in this case also.

I'd like (at least an option) to allow all LDAP users by default and require Approval for Non-LDAP-Accounts.

einhirn:
perhaps they are useful to others?

Our (internal) Mail gateway doesn't support authentication, so I needed to remove auth support. For this pull request I made it depend on "SMTP_AUTH" existing or being set to "none". This isn't yet reflected in the tests.

The other commit fixes #1012 for me: I'm just bypassing Approval process when a new user is signed in via LDAP. This could of course be made to be configurable - for those out there that want to approve each ldap user manually for Greenlight. I think I'd rather work with the role attribute if I want a subset of ldap users allowed into Greenlight...

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -214,7 +214,7 @@ class SessionsController < ApplicationController
     logger.info "Support: Auth user #{user.email} is attempting to login."
 
     # Add pending role if approval method and is a new user
-    if approval_registration && !@user_exists
+    if approval_registration && !@user_exists && @auth['provider'] != :ldap
       user.add_role :pending
 
       # Inform admins that a user signed up if emails are turned on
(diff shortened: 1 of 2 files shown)


farhatahmad:
HI @einhirn,

Thanks for taking the time to submit the PR. I think the mail check may actually be useful for other users but I think the LDAP approval is very specific to your deployment. 

Can you resubmit the PR without the LDAP code?

Also, please make sure the PR passes the Travis check.

einhirn:
Ok, since I can't seem to change this Pull request I'll just close it and supersede it with a new one.

PR #1035, opened 2020-03-19
Locales: Improve Spanish translation

mpellicer:


config/locales/es.yml
```diff
@@ -34,30 +34,30 @@ es:
...
+        info: Cambiar el color regular cambiará tanto el color claro como oscuro. Los valores de Aclarar y Oscurecer se pueden cambiar individualmente
         title: Color principal
         regular: Regular
         lighten: Aclarar
         darken: Oscurecer
       recording_visibility:
-        info: Establecer la visibilidad por default en las grabaciones nuevas
-        title: Visibilidad de grabación por default
-        warning: Esta configuración solo será aplicada a las nuevas sesiones
+        info: Establecer la visibilidad por defecto en las nuevas grabaciones
```
jfederico: If we are going to the purest of the language, it should be grabaciones nuevas, as the adjective in Spanish goes after the noun.

config/locales/es.yml
```diff
@@ -34,30 +34,30 @@ es:
...
         placeholder: Enlace de la imagen...
         title: Imagen personalizada
       cache:
-        info: Limpia el proveedor almacenado en cache lo cual forza todas la actualización de las peticiones nuevas
+        info: Limpia el proveedor almacenado en cache lo cual fuerza todas la actualizaciones de las peticiones nuevas
         title: Limpiar cache de proveedor
         button: Limpiar cache
       color:
-        info: Cambiar el color regular cambiará tanto el color claro como obscuro. Los valores de Aclarar y Oscurecer se pueden cambiar individualmente
+        info: Cambiar el color regular cambiará tanto el color claro como oscuro. Los valores de Aclarar y Oscurecer se pueden cambiar individualmente
```
jfederico: oscuro or obscure are correct both.

config/locales/es.yml
```diff
@@ -34,30 +34,30 @@ es:
       authentication:
         disabled: Deshabilitada
         enabled: Habilitada
-        info: Permitir solo a usuarios autenticados ingresar a una sala
+        info: Permitir solo a usuarios autenticados acceder a una sala
```
jfederico: acceder/ingresar in this context ids the same meaning

config/locales/es.yml
```diff
@@ -94,8 +94,8 @@ es:
       invalid_order: Ocurrió un error al actualizar la prioridad del rol. Verifique los valores asignados e inténtelo nuevamente
       invalid_update: Ocurrió un error al actualizar los permisos del rol. Verifique los valores asignados e inténtelo nuevamente
       name: Nombre del rol
-      new_role: Crear un rol nuevo
-      role_has_users: "Este rol se encuentra asignado a %{user_count}cuentas de usuario. Debe borrar todas las cuentas asociadas a este rol antes de eliminarlo."
+      new_role: Crear un nuevo rol
```
jfederico: Same as before. Ok, my mistake, It can be used before or after, but in this (and the former case), the adjective is used to make a distintion of another suibject, therefore it should go after the noun.

jfederico:
@mpellicer thanks for the PR, although the translations should be done through transifex. 

We can argue some of the reasons behind, but we are not going to get too far as my perspecti will always be in American Spanish and as I can see, your will be Spanish Catillian :D

Either way, we may be of better look if we bring this discussion to that app https://www.transifex.com/bigbluebutton/greenlight/dashboard/

The PRs are opened automatically once the translation is complete.

config/locales/es.yml
mpellicer: Obscuro es bastante menos común para los usuarios.

config/locales/es.yml
mpellicer: Este cambio es solamente por consistencia.

mpellicer:
I understand, from my perspective, this PR improves the Spanish translation and provides more consistency for the terminology, and I think new translations can be imported intro transifex.

I don't see a reason to argue, I'm new here and I just wanted to contribute some good stuff because the Spanish locale has some issues, feel free to decide what's best for your product.

Regards

jfederico:
I am sorry. I didn't mean to snap at you. Again, semantics are lost in translations, shall we use the word discuss instead of argue? ;) ("discutir el asunto es diferente que argumentarlo")

And no, we don't have time to be synchronizing translations both ways. So, we just don't merge direct PRs, only the ones generated by transifex. Not only for Spanish, but any other language.

I am looking at the files in transifex and actually most of your recommendation are already there. The issue is that no one has finished the translations in Spanish. Both strings (es and es_MX are 52 lines behind). If you want to take a look your help WILL be more appreciated there.

I can even promote you as a reviewer for the Spanish file if you are interested. We have always openings for voluteers, and you seem to be knowledgable of the grammar, so you will have the chance to aprove or reject the suggestions of other commiters.

PR #1110, opened 2020-03-29
Add files via upload

zibens:
Initial Latvian translation


jfederico:
Thanks for the PR @zibens but all translations must be done through Transifex. You would have to create an account, request the language if it is not there and then submit your proposal there.

I would encourage you to do so. We are always looking for contributors for transaltions

https://www.transifex.com/bigbluebutton/greenlight/dashboard/

PR #1116, opened 2020-03-29
[feature/change_messages]added messages

POLAPANASIUK:


jfederico:
Thanks for the PR @POLAPANASIUK but there are two incovenients.
1. All translations must be done through Transifex. You would have to create an account and submit your proposal there.
2. We have no interest whatsoever of translating Greenlight. If we did, we would have done it long time ago. It is valid to remove the BBB branding, and all the names we use to identify the project, but if you need customization, you should build your own image. The scripts for that are also included as part of the code.

jfederico:
https://www.transifex.com/bigbluebutton/greenlight/dashboard/

PR #1128, opened 2020-03-30
create proper test message and use correct env var for from address. …

marklackey:
…designed to work with sandboxed AWS ses


farhatahmad:
Hi @marklackey 

Thanks for taking the time to submit this. Rails already comes built in with an email client (Action Mailer) which we user to send emails. Switching the test to MailFactory while the actual emails are being sent by ActionMailer does not seem like a good idea.

Issue #1122, opened 2020-03-30
Users Mobile Homepage


mrkeksi:
Just a quick question regarding the mobile website: Is this only for me?
![CSS](https://i.ibb.co/fFHhVn7/Unbenannt6.png)

sotiris08:
I also have this problem on the password reset page.

![Screenshot_20200330-115250_Chrome](https://user-images.githubusercontent.com/62189294/77893871-2d04f700-727d-11ea-92fa-c7b7a2848752.png)

sotiris08:
> I also have this problem on the password reset page.
> 
> ![Screenshot_20200330-115250_Chrome](https://user-images.githubusercontent.com/62189294/77893871-2d04f700-727d-11ea-92fa-c7b7a2848752.png)

And it is actually the same on demo.bigbluebutton.org

farhatahmad:
@mrkeksi Thanks for reporting this. It's a visual issue that we need to fix on our end

mrkeksi:
Added new changes in #1342

![C824638E-F922-4BE3-8950-FE114B3F9C75](https://user-images.githubusercontent.com/8763656/79784603-ca19f380-8342-11ea-833a-0bd69dcc1ce6.jpeg)
![D543C4AF-4FA9-4FA4-B5D6-567EBB6ACA27](https://user-images.githubusercontent.com/8763656/79784608-cbe3b700-8342-11ea-9258-3f04d57e826a.jpeg)
![91FC43B2-AA1A-411F-B49C-43100E344AF3](https://user-images.githubusercontent.com/8763656/79784610-cc7c4d80-8342-11ea-9eae-c22e22e8bda9.jpeg)

mrkeksi:
Will be released in GL 2.6

PR #1131, opened 2020-03-31
CSS changes

mrkeksi:
It is my first pull request ever. I hope I can help at least a little. Dont know how to link to the issue #1122 

Another thing: I don't know if the media query is right there or if it would be better in another SCSS.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -59,6 +59,11 @@ html, body {
   background-color: white;
 }
 
+@media (max-width: 992px)
+.h-100 {
+    height: auto !important;
+}
+
 a {
   text-decoration: none !important;
   cursor: pointer;
--- a/app/views/rooms/cant_create_rooms.html.erb
+++ b/app/views/rooms/cant_create_rooms.html.erb
@@ -20,7 +20,7 @@
         <div class="row h-100 pt-7 align-items-center">
             <div class="col-12">
                 <div class="row">
-                    <div class="col-lg-5 col-sm-12">
+                    <div class="col-lg-5 d-none d-lg-flex">
                         <div class="card card-profile h-100">
                             <div class="card-header bg-primary h-50"></div>
                             <div class="card-body text-center">
--- a/app/views/rooms/cant_create_rooms.html.erb
+++ b/app/views/rooms/cant_create_rooms.html.erb
@@ -17,10 +17,10 @@
 
 <div id="cant-create-room-wrapper" class="background h-100 cant-create-room-wrapper">
     <div class="container h-100">
-        <div class="row h-100 align-items-center">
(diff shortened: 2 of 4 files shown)


farhatahmad:
Hi @mrkeksi 

Thanks for the PR. Everything looks good. I pulled the code to my local and ran it and it looks like you've solved the issue. 

I left some comments, feel free to respond when you have some time

app/assets/stylesheets/application.scss
```diff
@@ -59,6 +59,11 @@ html, body {
   background-color: white;
 }
 
+@media (max-width: 992px)
```
farhatahmad: Is the media query necessary? I tried testing with and without it and couldn't find the difference

app/views/rooms/cant_create_rooms.html.erb
```diff
@@ -17,10 +17,10 @@
 
 <div id="cant-create-room-wrapper" class="background h-100 cant-create-room-wrapper">
     <div class="container h-100">
-        <div class="row h-100 align-items-center">
+        <div class="row h-100 pt-7 align-items-center">
             <div class="col-12">
                 <div class="row">
-                    <div class="col-5">
+                    <div class="col-lg-5 col-sm-12">
```
farhatahmad: We may actually want to hide the profile portion when on mobile, which would make it easier for a user on mobile to join a room

app/assets/stylesheets/application.scss
mrkeksi: It is actually only for desktop users as there is this fixed height in the wrapper. It would be better to work with "viewheight" instead of percentage.

farhatahmad: Just coming back around to this now. The media query would better be wrapped in something specific for the "Can't create rooms" view and moved to the `rooms.scss`. Something like this:

```
#cant-create-room-wrapper {
  @media....
}
```

app/views/rooms/cant_create_rooms.html.erb
```diff
@@ -17,10 +17,10 @@
 
 <div id="cant-create-room-wrapper" class="background h-100 cant-create-room-wrapper">
     <div class="container h-100">
-        <div class="row h-100 align-items-center">
+        <div class="row h-100 pt-7 align-items-center">
```
farhatahmad, line 20: The `pt-7` can be removed from here and instead added to the media query as `padding-top: 3rem`

app/assets/stylesheets/application.scss
```diff
@@ -59,6 +59,11 @@ html, body {
...
+.h-100 {
```
farhatahmad, line 63: Instead of changing the h-100 for everything, you might be better off just changing the vertical align so it's no longer centered. `align-items: unset !important;` should work inside the media query

PR #1138, opened 2020-03-31
Add shibboleth omniauth provider

schrd:
The provider can be configured using the SHIB_* env variables. It is
possible to map attributes to roles using the SHIB_ROLE_FIELDS env
variable:

assume you have configured your shibboleth SP to provide HTTP_AFFILIATION and
HTTP_SHIB_EP_ENTITLEMENT. Any entry in those two attributes will be
compared to the existing roles in greenlight. If such a role exists, it
will be assigned. If the user has roles assigned which are not mentioned
in those attributes, the roles will be removed from the user.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -48,6 +48,7 @@ gem 'bootsnap', '>= 1.1.0', require: false
 gem 'omniauth'
 gem 'omniauth-twitter'
 gem 'omniauth-google-oauth2'
+gem 'omniauth-shibboleth'
 gem 'omniauth-bn-office365', git: 'https://github.com/blindsidenetworks/omniauth-bn-office365.git', tag: '0.1.0'
 gem 'omniauth-bn-launcher', git: 'https://github.com/blindsidenetworks/omniauth-bn-launcher.git', tag: '0.1.3'
 gem 'bn-ldap-authentication', git: 'https://github.com/blindsidenetworks/bn-ldap-authentication.git'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -201,6 +201,8 @@ GEM
     omniauth-oauth2 (1.5.0)
       oauth2 (~> 1.1)
       omniauth (~> 1.2)
+    omniauth-shibboleth (1.3.0)
+      omniauth (>= 1.0.0)
     omniauth-twitter (1.4.0)
       omniauth-oauth (~> 1.1)
       rack
@@ -374,6 +376,7 @@ DEPENDENCIES
   omniauth-bn-launcher!
   omniauth-bn-office365!
   omniauth-google-oauth2
+  omniauth-shibboleth
   omniauth-twitter
   pagy
   pg (~> 0.18)
@@ -399,3 +402,6 @@ DEPENDENCIES
   uglifier (>= 1.3.0)
   web-console (>= 3.3.0)
   webmock
+
+BUNDLED WITH
+   1.17.3
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -213,6 +213,41 @@ class SessionsController < ApplicationController
 
     logger.info "Support: Auth user #{user.email} is attempting to login."
 
+    if @auth['provider'] == 'shibboleth'
+      logger.info("Shibboleth login: #{@auth}")
+      logger.info("Shibboleth info: #{@auth.info}")
+      logger.info("Shibboleth extra: #{@auth.extra[:raw_info]}")
+      wanted_roles = []
+      Rails.application.config.omniauth_shibboleth_role_fields.each{|field|
+        # multivalue fields have their values joined by ; escaped by \
+        field_values = @auth.extra.raw_info[field].split(/(?<!\\);/)
+        field_values.each{|role_name|
+          # never assign reserved roles
+          next if Role::RESERVED_ROLE_NAMES.include?(role_name)
+          r = Role.find_by_name(role_name)
+          if r
+            wanted_roles << r
+          end
+        }
+      }
+      logger.info("wanted_roles: #{wanted_roles}")
+      current_roles = user.roles.where.not(name: Role::RESERVED_ROLE_NAMES).to_a
+      logger.info("current_roles: #{current_roles}")
+      # remove_unwanted_roles
+      current_roles.each{|r|
+        if not wanted_roles.include?(r)
+          logger.info("removing role #{r.name} from #{user.social_uid}")
+          user.remove_role(r.name)
+        end
+      }
+      # add missing roles
+      wanted_roles.each{|r|
+        if not current_roles.include?(r)
+          logger.info("adding role #{r.name} to #{user.social_uid}")
+          user.add_role(r.name)
+        end
+      }
+    end
     # Add pending role if approval method and is a new user
     if approval_registration && !@user_exists
       user.add_role :pending
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -17,6 +17,12 @@ Rails.application.config.omniauth_twitter = ENV['TWITTER_ID'].present? && ENV['T
 Rails.application.config.omniauth_google = ENV['GOOGLE_OAUTH2_ID'].present? && ENV['GOOGLE_OAUTH2_SECRET'].present?
 Rails.application.config.omniauth_office365 = ENV['OFFICE365_KEY'].present? &&
                                               ENV['OFFICE365_SECRET'].present?
+Rails.application.config.omniauth_shibboleth = ENV['SHIBBOLETH'].present?
+Rails.application.config.omniauth_shibboleth = ENV['SHIB_UID_FIELD'].present? &&
+                                               ENV['SHIB_NAME_FIELD'].present?  &&
+                                               ENV['SHIB_SESSION_ID_FIELD'].present? &&
+                                               ENV['SHIB_APPLICATION_ID_FIELD'].present? &&
+                                               ENV['SHIB_EMAIL_FIELD'].present?
 
 SETUP_PROC = lambda do |env|
   OmniauthOptions.omniauth_options env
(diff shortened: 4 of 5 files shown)


farhatahmad:
Hi @schrd 

For the time being, we're not looking to add more authentication methods to Greenlight. The more authentication methods we add, the more work we need to put in to supporting them, which we don't have the capacity to do.

I can already forsee issues arising as the shibboleth gem hasn't been maintained in over ~3 years. You're welcome to merge this into your own deployment and use it as needed, but it's just not something we want to add and support anytime soon

schrd:
Hi @farhatahmad 

sorry for hearing this. I understand that supporting lots of authentication methods may be hard. However I believe that the reason there have not been any updates on the shibboleth gem is the fact that all authentication in this gem is done by the webserver. The shibboleth gem just trusts headers sent by the webserver. So all the complicated SAML stuff is done by mod_shib together with apache, nothing is done in ruby.

We have this already merged into our own deployment and wanted to give back our enhancement. Omniauth-shibboleth is used by gitlab too, so I believe there may be good reason to trust it.

farhatahmad:
We appreciate you guys giving back and providing this PR. If any one asks, I'll point them to this PR as a guide for them to follow. 

Merging it means we agree to support it, and it's just not something were able to do at the moment.

Thank you again and please feel free to open any issues or PRs for anything else

aguerson:
https://github.com/bigbluebutton/greenlight/issues/1472

wmchris:
@farhatahmad i just wanted to inform you, that shibboleth is the main authentication method used in german universities based (and many other european states). 

Shibboleth is created by edu and real FOSS. Maybe this big userbase will change your optinion - maybe ditch one of the commercal auth providers (for ex. office365) in favor of it?

basisbit:
@farhatahmad is there any chance to get this PR reviewed again? `shibboleth` is by far the main authentication method used in universities around the world for such stuff. Thus it would be a nice addition for a teaching platform like bbb. 

The changes looks very easy to maintain as they are only rather small additions using already existing logic. If you want, this PR could be extended by tests in the testsuite.

I am sure that if anything of this code would need changes, someone from the community will quickly do that work.
Merging this PR would make it much easier for universities to abandon their forks and instead contribute here again.

aguerson:
you have to be careful
shibboleth is easy, but not the nginx part !
https://github.com/bigbluebutton/greenlight/issues/1859

aguerson:
@basisbit check my issue

aguerson:
@basisbit my own conf ( with the good cgi params )

```
/opt/bigbluebutton/greenlight_dev/Gemfile
52a53
> gem 'omniauth-shibboleth'
```

```
/opt/bigbluebutton/greenlight_dev/Gemfile.lock
249a250,251
>     omniauth-shibboleth (1.3.0)
>       omniauth (>= 1.0.0)
442a445
>   omniauth-shibboleth

```
```
/opt/bigbluebutton/greenlight_dev/app/assets/stylesheets/main.scss
135a136,146
> .customBtn-shibboleth {
>   @extend .customBtn;
>   background: $button-color-blue;
> 
>   .customBtn-image {
>     background: #ffffff image-url("shibboleth-logo.png") no-repeat left top;
>     background-size: 18px 18px;
>     padding:10px 10px 10px 10px;
>   }
> }
> 
167a179
> 

```
```
/opt/bigbluebutton/greenlight_dev/app/helpers/application_helper.rb
25c25
<     providers = configured_providers & [:google, :twitter, :office365, :ldap]
---
>     providers = configured_providers & [:google, :twitter, :office365, :ldap, :shibboleth]
```

```
/opt/bigbluebutton/greenlight_dev/config/initializers/omniauth.rb
18a19
> Rails.application.config.omniauth_shibboleth = ENV['SHIBBOLETH'].present?
61a63,78
>     if Rails.configuration.omniauth_shibboleth
>       Rails.application.config.providers << :shibboleth
> 
>       provider :shibboleth, {
>         #:debug => true,
>         :request_type => :header,
>         :uid_field  => "eppn",
>         :name_field => "displayName",
>         :shib_session_id_field     => "Shib-Session-ID",
>         :shib_application_id_field => "Shib-Application-ID",
>         :info_fields => {
>           :email => "mail",
>         },
>       }
>     end
>
```

```
/opt/bigbluebutton/greenlight_dev/config/locales/en.yml
475a479
>     shibboleth: Shibboleth
```
```
/opt/bigbluebutton/greenlight_dev/config/locales/fr.yml
473a477
>     shibboleth: Shibboleth
```

```
/opt/bigbluebutton/greenlight_dev/sample.env
88a95,98
> # Shibboleth Signin
> # true or false
> SHIBBOLETH=
>
```

basisbit:
@aguerson I am not sure what to do with your comment or why you post that here. @schrd already uses the code from this PR successfully in production for half a year or so.

farhatahmad:
@basisbit Not anytime soon - it's actually gotten even less likely as the resources we have to maintain Greenlight is even less than before. There were some discussions of completely replacing all external authentications methods with something like KeyCloak but those discussions are still in their early stages

aguerson:
@basisbit
These two lines in not in the @schrd conf

```
>         :shib_session_id_field     => "Shib-Session-ID",
>         :shib_application_id_field => "Shib-Application-ID",
```

And it is mandatory !

```
# SHIB_SESSION_ID_FIELD='HTTP_SHIB_SESSION_ID'
# SHIB_APPLICATION_ID_FIELD='HTTP_SHIB_APPLICATION_ID'
```
These lines are not good !

**"_"** is **now** replaced by **"-"**

!!

So if you valide the PR,  you will have to resolv many issue...

aguerson:
@farhatahmad @basisbit this one ? https://github.com/imagov/keycloak

basisbit:
https://www.keycloak.org/

moschlar:
@aguerson You might want to check out #185 - since Shibboleth is basically just an implementation of the SAML protocol that should also apply. And it is much easier to configure, because it's all application-side, no need to fiddle with third party nginx modules and such.

We are using that PR for some months now without problems.

aguerson:
you think Shibboleth and SAML is the same, but it is not truth. The way you install the IDP is not the same. A native shibboleth solution is better is my case. But you are right, shibboleth works with SAML protocol.

schrd:
We ported this patch to 2.7 (and added other things such as the choice to force authentication for rooms and a fix with missing database connections in the pool for background workers). Our fork is available at https://gitlab.hrz.tu-chemnitz.de/bigbluebutton/greenlight/-/tree/tuc-urz-fixes-2.7 and is already used by some universities in Germany. We run this in production with a few thousand users since April 2020. 

Shibboleth auth allows us to authenticate users from other universities, so we can provide access to our BBB instance to smaller universities which don't have network capacity to offer BBB to all of their users. With Shibboleth in federation we don't see passwords of the users from foreign universities because authentication happens at their IDP.

We run this behind apache instead of nginx. There are instructions from our national education network provider how to configure Apache with shibboleth. The only thing greenlight has to do is trust some request headers that are sent from apache. So this is not complicated at all from a greenlight perspective. All the magic SAML stuff happens inside apache. In apache you make rules which identity provider from the SAML federation will be allowed to sign in etc. Implementing LDAP auth for example is much more difficult to get right.

We also use shibboleth for authentication for security reasons. We educate our users to not enter their passwords into any web forms except for the IdP login form. We implemented some measures to protect our IdP login forms from being phished. If you implement 2FA on your IdP, you don't have to implement 2FA in all of the applications. All of this has many advantages and is used in production at many universities world wide since many years.

I am willing to fix issues with this PR or the current version for 2.7 if there is a chance to get this in. If you would like to have some things implemented differently, just tell me and I will try to solve it. If you still don't want this in, thats also okay for me. Maintaining a fork for greenlight isn't that hard. But I must admit that the contributor and operator experience of BigBlueButton and its related projects is very frustrating. To me it seems like the developers don't understand real world operating problems or are just ignoring them. Let me point out some examples (there are many more):

- https://github.com/bigbluebutton/bigbluebutton/pull/10349 This helps to protect from overload because scalelite has such a stupid balancing strategy. This strategy works well for school, where classes have the same size. But it does not work at all for universitites where you have lectures with a few hundred participants and seminars with something like 20 users
- https://github.com/blindsidenetworks/scalelite/issues/99 more clever balancing strategies. LMS software such as moodle usually knows how man participants are available in a class. Greenlight could easily store a planned room size in rooms and provide this information to scalelite. This would help a lot for scalelite to make better decisions.
- https://github.com/bigbluebutton/bigbluebutton/issues/9024 This is a real world problem and it has legal implications due to the GDPR in Europe. 
- https://github.com/bigbluebutton/bigbluebutton/pull/9792 It took almost 2 month to bring in this rather simple fix. This issue was really annoying for our users
- not providing packaging sources for BBB makes it hard for operators to bring fixes early to users and enhance BBB user experience. It makes it hard to test PRs and give feedback as well.

I don't think that greenlight, scalelite oder BigBlueButton are bad products, not at all. I really like them and its awesome to have something like this. And it is absolutely understandable that people working at the supporting companies had a hard time after the COVID19 lockdowns. There are a lot of people who are willing to help. So please be more open for help so that there is no need for the community to fork things like it was necessary in OpenOffice. It was the non-functional community process in OpenOffice which was the cause for Libreoffice to exist, not the product itself.

schrd:
> you think Shibboleth and SAML is the same, but it is not truth. The way you install the IDP is not the same. A native shibboleth solution is better is my case. But you are right, shibboleth works with SAML protocol.

Implementing SAML in a correct way is hard. Many SAML libraries such as https://github.com/omniauth/omniauth-saml only support one identity provider. This might work for some cases but not for all. In our case this wouldn't work because we provide greenlight to other universities as well, which means that there is need for federation with multiple identity providers. Shibboleth solves all of this outside of greenlight.

moschlar:
> Implementing SAML in a correct way is hard. Many SAML libraries such as https://github.com/omniauth/omniauth-saml only support one identity provider. This might work for some cases but not for all. In our case this wouldn't work because we provide greenlight to other universities as well, which means that there is need for federation with multiple identity providers. Shibboleth solves all of this outside of greenlight.

We solve this by having a dedicated WAYF service that serves as an IdP for applications such as greenlight and as a SP for the actual IdPs of other universities.

schrd:
> > Implementing SAML in a correct way is hard. Many SAML libraries such as https://github.com/omniauth/omniauth-saml only support one identity provider. This might work for some cases but not for all. In our case this wouldn't work because we provide greenlight to other universities as well, which means that there is need for federation with multiple identity providers. Shibboleth solves all of this outside of greenlight.
> 
> We solve this by having a dedicated WAYF service that serves as an IdP for applications such as greenlight and as a SP for the actual IdPs of other universities.

Sure, but you need metadata for trust relationship between IdP and SP. If you have only one IdP you can import the certificate of your IdP into the SP. For multiple IdPs you need signed metadata which bundle the certificates of participating IdPs. To my knowledge omnibus-saml can't handle that.

moschlar:
> Sure, but you need metadata for trust relationship between IdP and SP. If you have only one IdP you can import the certificate of your IdP into the SP. For multiple IdPs you need signed metadata which bundle the certificates of participating IdPs. To my knowledge omnibus-saml can't handle that.

I think in our case the metadata just gets re-signed by the WAYF IdP, so for the application it just seems like just one IdP.

PR #1144, opened 2020-04-01
Test email configuration during setup

marklackey:
Used ActionMailer instead of mail factory.


farhatahmad:
@marklackey finally got around to this. Thanks for the PR. I've updated it to match some of the newer changes

farhatahmad (approved):

PR #1150, opened 2020-04-01
Revert "Fixed issue with duplicate roles from LDAP sigin (#1090)"

mrkeksi:
This reverts commit fab3b479c9015b8c6e0187099cbb63b0fedc7c47.

With this fix I no longer get an LDAP role at the first login of the user which is declared in the .env file with "LDAP_ROLE_FIELD".

Without the code the role will be retrieved directly at login (no matter if new or existing).

### DIFF
--- a/app/models/concerns/auth_values.rb
+++ b/app/models/concerns/auth_values.rb
@@ -57,14 +57,13 @@ module AuthValues
   end
 
   def auth_roles(user, auth)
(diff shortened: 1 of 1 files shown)


farhatahmad:
Hi @mrkeksi 

Thanks for the heads up. Turns out I left a test variable included the merged PR. Ill create another PR fixing it

PR #1160, opened 2020-04-02
Set the number of puma workers via the WEB_CONCURRENCY environment variable

trs80:
This allows setting WEB_CONCURRENCY in .env to scale up greenlight performance.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -23,7 +23,7 @@ environment ENV.fetch("RAILS_ENV") { "development" }
 # Workers do not work on JRuby or Windows (both of which do not support
 # processes).
 #
-# workers ENV.fetch("WEB_CONCURRENCY") { 2 }
+workers ENV.fetch("WEB_CONCURRENCY") { 2 }
 
 # Use the `preload_app!` method when specifying a `workers` number.
 # This directive tells Puma to first boot the application and load code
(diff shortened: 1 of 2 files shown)


farhatahmad:
Hi @trs80 

Thanks for PR. I'm not quite sure about defaulting this to 2. We already have the `RAILS_MAX_THREADS` defaulted to 5. My worry is that increasing the number of puma workers will probably cause the machine to run out of memory (if on a small machine), or take away memory from BigBlueButton (which needs it a lot more).

Is there a reason you find the need to set this? Are you noticing speed issues?

trs80:
We have 1600 students hitting greenlight, with one process it was manifestly inadequate so I bumped it up to 8 which seemed fine. On memory, each one looks to be about 200MB resident. I'm happy to have it default to 1, the server I'm running it on has BBB running but isn't included in the load balancer so it's unused.

trs80:
ActionCable stops working with multiple processes, you need to set its adapter to postgresql instead of async. This makes automatically joining when the meeting starts work again.

FreeLanceCoder2020:
I also have a problem with slow speed in a large number of users. (404 eroor)
I use scalelite but the greenlight server is separate.
I hope these settings solve the problem.

farhatahmad:
Going to be replaced by #1701

PR #1161, opened 2020-04-02
make help link configurable

sualko:
This new option is similar to the REPORT_ISSUE_URL option and let's hide the "need help?" link inside the dropdown navigation. It's also helpful if you like to provide an user guide in your language.


sample.env
```diff
@@ -172,6 +172,12 @@ MAINTENANCE_WINDOW=
 # Button can be disabled by setting the value to blank
 REPORT_ISSUE_URL=https://github.com/bigbluebutton/greenlight/issues/new
 
+# The link to the Need help? button that appears on the Account Dropdown
+#
+# Defaults to the Greenlight documentation
+# Button can be disabled by setting the value to blank
+HELP_URL=http://docs.bigbluebutton.org/install/greenlight-v2.html
```
farhatahmad, line 179: Can you change this to https://docs.bigbluebutton.org/greenlight/gl-overview.html please

sualko: done

config/application.rb
```diff
@@ -126,6 +126,7 @@ class Application < Rails::Application
     config.maintenance_mode = ENV["MAINTENANCE_MODE"] == "true"
 
     config.report_issue_url = ENV["REPORT_ISSUE_URL"]
+    config.help_url = ENV["HELP_URL"]
```
farhatahmad, line 129: This will also hide the help link for all current installations since the `ENV["HELP_URL"]` for those users is "". You'll probably want to change this to `ENV["HELP_URL"].presence || "https://docs.bigbluebutton.org/greenlight/gl-overview.html"` so that it defaults to the URL if the ENV variable isn't set

PR #1162, opened 2020-04-02
add option to customize app title

sualko:
I know the recommended way to customize the look and feel is to create your own docker image, but I think for the most common use cases that's way to much. App title is a good example.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -28,8 +28,8 @@
       </script>
     <% end %>
 
-    <title><%= t("bigbluebutton") %></title>
-    <meta property="og:title" content="<%= t("bigbluebutton", locale: :en) %>" />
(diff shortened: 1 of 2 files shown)


farhatahmad:
Hi @sualko 

Thanks for the PR. This isn't really the full solution. There are tons of other places that the word "BigBlueButton" appears on the website, which is why we recommend changing the `bigbluebutton` string in the translation file instead. We have plans to make branding easier, but if we do it, it'll need to be the full solution

sualko:
> There are tons of other places that the word "BigBlueButton" appears on the website

In my opinion the title is different from the product name. For example in my case I don't want to hide that BBB is used, but I want to have a meaningful title if multiple tabs are open. Also I think it's quite common that you like to have a title like "teaching via company", so replacing the translation would be no good idea.

Issue #899, opened 2019-12-23
Rails asset precompile error


GremL1N:
When install from source, have error with:
`RAILS_ENV=production rails assets:precompile`

```
yarn install v1.21.1
[1/4] Resolving packages...
success Already up-to-date.
Done in 0.07s.
rails aborted!
SassC::SyntaxError: Error: Undefined variable: "$gray-400".
        on line 33:14 of vendor/bundle/ruby/2.5.0/bundler/gems/tabler-rubygem-4c16e6dc930f/assets/stylesheets/tabler/_core.scss
        from line 20:9 of app/assets/stylesheets/_tabler-custom.scss
>>      background: $gray-400;
   -------------^
(Stack trace here)
```
I think this error comes from tabler-rubygem (https://github.com/blindsidenetworks/tabler-rubygem.git). Issues disabled in it repo. Thats why i post it here.

```
$ bundle list
  * actioncable (5.2.4.1)
  * actionmailer (5.2.4.1)
  * actionpack (5.2.4.1)
  * actionview (5.2.4.1)
  * activejob (5.2.4.1)
  * activemodel (5.2.4.1)
  * activerecord (5.2.4.1)
  * activestorage (5.2.4.1)
  * activesupport (5.2.4.1)
  * arel (9.0.0)
  * ast (2.4.0)
  * autoprefixer-rails (9.7.3)
  * bcrypt (3.1.13)
  * bigbluebutton-api-ruby (1.7.0)
  * bn-ldap-authentication (1.0.1 9b93c64)
  * bootsnap (1.4.5)
  * bootstrap (4.3.1)
  * builder (3.2.4)
  * bundler (1.17.3)
  * cancancan (2.3.0)
  * coffee-rails (4.2.2)
  * coffee-script (2.4.1)
  * coffee-script-source (1.12.2)
  * concurrent-ruby (1.1.5)
  * coveralls (0.8.23)
  * crass (1.0.5)
  * docile (1.3.2)
  * erubi (1.9.0)
  * execjs (2.7.0)
  * faraday (0.17.1)
  * ffi (1.11.3)
  * font-awesome-sass (5.9.0)
  * globalid (0.4.2)
  * hashie (3.6.0)
  * hiredis (0.6.3)
  * http_accept_language (2.1.1)
  * i18n (1.7.0)
  * i18n-language-mapping (0.1.0)
  * jaro_winkler (1.5.4)
  * jbuilder (2.9.1)
[...]

PR #1177, opened 2020-04-03
Freezing sprocket dependency

l0rn:
Version 4.0.0 introduces breaking behavior for this project.

Freeze is needed until migration is done as in: https://github.com/rails/sprockets/blob/master/UPGRADING.md

Solves #899

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,8 @@ end
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'rails', '~> 5.2.3'
 
+gem 'sprockets', '< 4.0.0'
+
 # Use Puma as the app server
 gem 'puma', '~> 3.12'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -392,6 +392,7 @@ DEPENDENCIES
   shoulda-matchers (~> 3.1)
   spring
   spring-watcher-listen (~> 2.0.0)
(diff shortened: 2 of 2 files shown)


Gemfile.lock
```diff
@@ -392,10 +392,14 @@ DEPENDENCIES
...
+  sprockets (< 4.0.0)
   sqlite3 (~> 1.3.6)
   tabler-rubygem!
   turbolinks (~> 5)
   tzinfo-data
   uglifier (>= 1.3.0)
   web-console (>= 3.3.0)
   webmock
+
+BUNDLED WITH
```
farhatahmad, line 404: Can you remove this from the PR? That should make Travis pass

l0rn: done

Gemfile
```diff
@@ -10,6 +10,8 @@ end
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'rails', '~> 5.2.3'
 
+gem 'sprockets', '< 4.0.0'
```
jfederico, line 13: Perhaps we should go to `4.0` instead

l0rn: well sure - if you adjust the scss code to mitigate the breaking changes. But for now be aware everyone who sets up a fresh environment using the current master will find this software broken

farhatahmad:
I believe I did this as part of the move to Rails 5.2.4.3

PR #1199, opened 2020-04-04
use email input for sign in

sualko:
This is specially useful on mobile devices.﻿


mrkeksi:
Also here: https://github.com/bigbluebutton/greenlight/blob/8f454cad0ece0a14c678bf0ecc1c6dddff2b873a/app/views/shared/modals/_invite_user_modal.html.erb#L30

https://github.com/bigbluebutton/greenlight/blob/8f454cad0ece0a14c678bf0ecc1c6dddff2b873a/app/views/users/components/_account.html.erb#L30

This cannot be implemented in the Invite Participant field, otherwise it would not be possible to invite multiple participants (separated by commas).

sualko:
Is there a reason why this small pr is not merged?

lkiesow:
@sualko, there seems to be a problem that  @mrkeksi pointed out in the comment above yours, doesn't it?

sualko:
> there seems to be a problem that @mrkeksi pointed out in the comment above yours, doesn't it?

It's no problem, it's an addition. I'm happy to add it, but if there is no willingness to merge this, I don't know why I should commit my time. If this change is required that my pr is merged, I'm happy to do it.

farhatahmad:
@sualko There is. Having trouble working my way through issues, reviewing PRs and doing development. If you can get those extra fields added, then I can get it into the next version

sualko:
Done. Feel free to squash.

PR #1222, opened 2020-04-06
GRN2-xx: Added room configuration tab to allow admins to set room settings as always enabled, optional or disabled

### LINKED ISSUE #1114: Add ENABLED_ROOM_FEATURES to .env to control Room Setting defaults
By default all Room Settings are disabled when you create a new room.

In some scenarios it would be useful to have some of the Room Settings default to enabled on creation of a new room, so users don't forget to set them.

For example, if you have Site Setting - "Require authentication for Rooms" disabled, I would like to have Room Setting - "Require moderator approval before joining" enabled by default, so unauthenticated users can't jump into a meeting.

Please consider adding an ENABLED_ROOM_FEATURES to the Greenlight .env that allows you to set one or more default Room Setting to enabled.

Enabling a setting in ENABLED_ROOM_FEATURES and removing it from ROOM_FEATURES should give you a room setting that is enabled and users cannot modify it. This may be useful in some situations.

farhatahmad:
fixes #1114

### DIFF
--- a/app/assets/javascripts/room.js
+++ b/app/assets/javascripts/room.js
@@ -144,10 +144,11 @@ function showCreateRoom(target) {
   $("#room_access_code").val(null)
 
   $("#createRoomModal form").attr("action", $("body").data('relative-root'))
-  $("#room_mute_on_join").prop("checked", false)
-  $("#room_require_moderator_approval").prop("checked", false)
-  $("#room_anyone_can_start").prop("checked", false)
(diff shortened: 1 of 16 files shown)


jfederico:
@farhatahmad can you please resolve conflicts?

jfederico:
@farhatahmad Can you resolve the conflict?

farhatahmad:
@jfederico Done

PR #1250, opened 2020-04-09
GRN2-xx: Replaced bbb_id field with a better string generator

### LINKED ISSUE #1242: Potential room bbb_id conflict
https://github.com/bigbluebutton/greenlight/blob/3f3c69d85edef5222787cfe256d86854118c802a/app/models/room.rb#L88

This line will produce the same `bbb_id` if two rooms are created in the same second.

farhatahmad:
fixes #1242


farhatahmad:
@jfederico Your change works as well. The only reason I went with hex(20) instead of uuid is to stay consistent with the current length and format of bbb_id. Either way solves the problem

PR #1287, opened 2020-04-13
Additional LDAP Authentication Methods

### LINKED ISSUE #1082: Anonymous ldap bind doesn't work/not available
Does greenlight support anonymous binding? It is not clear from the documentation and commenting out or leaving empty LDAP_BIND_DN and LDAP_PASSWORD doesn't seem to work.
Anonymous binding is pretty standard and I guess should be easy to implement.
Thank you.

lkiesow:
This patch allows anonymous and user based authentication to LDAP
servers. This change is based on a patch against bn-ldap-authentication:

-  https://github.com/blindsidenetworks/bn-ldap-authentication/pull/2

The patch introduces a new environment variable `LDAP_AUTH` which
controls the authentication method used against the LDAP server:

- `anonymous` enables an anonymous bind to the LDAP with no password
  being used.

- `user` uses the user's own credentials to search for his data,
  enabling authenticated login to LDAP without the need for a user with
  global read privileges.

The default still remains at using a bind user, allowing for a seamless
upgrade path from the previous version.

This fixes #1082

---

This patch should also come with a change to the [LDAP authentication documentation](https://docs.bigbluebutton.org/greenlight/gl-config.html#ldap-auth) but I'm not sure how to contribute to that documentation.


lkiesow:
If you want to test this but are lacking an LDAP server,
here is the configuration for a test server you can use:

```sh
export LDAP_SERVER=ldap.forumsys.com
export LDAP_PORT=389
export LDAP_METHOD=plain
export LDAP_UID=uid
# anonymous, simple, user
export LDAP_AUTH=user
export LDAP_BASE=dc=example,dc=com
export LDAP_BIND_DN=
export LDAP_PASSWORD=
export LDAP_ROLE_FIELD=

export ALLOW_GREENLIGHT_ACCOUNTS=false
export DEFAULT_REGISTRATION=open

./bin/start
```

You can then sign in with:

```
u: jmacy
p: password
```

farhatahmad:
Hi @lkiesow,

Thanks for this PR and your work with helping others. We'll get this included in the next release

PR #1291, opened 2020-04-14
Update and rename sample.env to .env

jerry1807:


lazouz:
Hello @jerry1807 you commited some secrets. You have to change your commit AND to update your secrets on your instance.

lkiesow:
Is there any reason to rename the `sample.env` to `.env`?
This will essentiallly hide the file and people will not find it any longer.

Issue #1265, opened 2020-04-10
Feature Request: Force Access code for Rooms


holzhannes:
Due to security of the system it would be very nice if an admin could force access codes for all rooms or a group of users.

MarcelWaldvogel:
Would increasing the length of the URL help you as well? One of the reasons for access codes is to add entropy, so uninvited joins to a room are hard. Making the URL harder to guess might be an alternative to solve the problem behind your request.

Right now, the first three digits of a GreenLight room name are determined by the user name and the remaining two triplets provide less than 32 bits of entropy.

Sea444:
I think for users it's better to have a short room-name and a seperate pin instead of a long and complex roomname. 
(Maybe you remember transition from Kontonummer and Bankleitzahl to IBAN. :-D )

And yes: Room names without first three letters of host name would be better.

holzhannes:
> Would increasing the length of the URL help you as well? One of the reasons for access codes is to add entropy, so uninvited joins to a room are hard. Making the URL harder to guess might be an alternative to solve the problem behind your request.
> 
> Right now, the first three digits of a GreenLight room name are determined by the user name and the remaining two base-36 triplets provide at most 31.02 bits of entropy.

Thank you @MarcelWaldvogel for your Idea and thanks @Sea444 I share your opinion for short room links without user indication within the link.
From my point of view it is an important security setting to force codes for rooms maybe also by a setting within the user roles. Just extending the length would not help in the end if you don't trust the channel where the link is shared or you can't control how long a room will exist. If you have a policy to share codes on a second channel because you don't trust the first channel longer links would not help as well.

MarcelWaldvogel:
@holzhannes If you don't trust the channel where you share the link, you need a separate channel for the access code anyway. Why not use that channel to send the entire link instead?

(Yes, I know that this might require manually typing more, but I would claim that most users which would like to communicate in a video conference have a trustworthy channel from where they can click a link. Making it more complicated for the rest of us is not an option, IMHO. See e.g. the access code URL patch #1331, which increases the entropy by another 20 bits while still being as easy to use for most users, i.e., just a single click on the URL.)

MarcelWaldvogel:
I just submitted a pull request (#1451) which makes the room URLs much less guessable.

I believe that URLs will be communicated over digital media, thus the invitee just has to click on it (or, in the worst case, copy/paste it). So users rarely have to type it, which is in contrast to the IBAN example above.

[...]

PR #1331, opened 2020-04-18
Access code as part of the URL

### LINKED ISSUE #1303: Feature request: Pass access code as part of URL
I would like to be able to send out room URLs which already contain the access code, e.g., in the form `https://demo.bigbluebutton.org/b/abc-def-ghi?access_code=123456`.

This is related to #1265, but independent of the increase in URL entropy, when dialin is necessary.

MarcelWaldvogel:
- Allow access code to be passed as part of the URL query string (`?code=`)
- Include access code in the copyable URL

Uses `code` as the query parameter in the `GET` request instead of `access_code` as some users will have to type it in, so preferring a shorter string.

Fixes #1303

### DIFF
--- a/app/controllers/rooms_controller.rb
+++ b/app/controllers/rooms_controller.rb
@@ -66,6 +66,11 @@ class RoomsController < ApplicationController
     @room_running = room_running?(@room.bbb_id)
     @shared_room = room_shared_with_user
 
+    # Accept access code as part of the URL
+    if params.key?(:code)
+      session[:access_code] = params[:code]
+    end
+
     # If its the current user's room
     if current_user && (@room.owned_by?(current_user) || @shared_room)
       if current_user.highest_priority_role.get_permission("can_create_rooms")
--- a/app/controllers/rooms_controller.rb
+++ b/app/controllers/rooms_controller.rb
@@ -67,9 +67,7 @@ class RoomsController < ApplicationController
     @shared_room = room_shared_with_user
 
     # Accept access code as part of the URL
-    if params.key?(:code)
-      session[:access_code] = params[:code]
-    end
+    session[:access_code] = params[:code] if params.key?(:code)
 
     # If its the current user's room
     if current_user && (@room.owned_by?(current_user) || @shared_room)
--- a/app/helpers/rooms_helper.rb
+++ b/app/helpers/rooms_helper.rb
@@ -37,4 +37,13 @@ module RoomsHelper
     @diff = current_user.rooms.count - limit
     @diff.positive? && current_user.rooms.pluck(:id).index(room.id) + 1 > limit
   end
+
+  def room_join_url
+    if @room.access_code.nil? || @room.access_code.empty?
+      return request.base_url + @room.invite_path
+    else
+      return request.base_url + @room.invite_path + "?code=" + @room.access_code
+    end
+  end
+
 end
--- a/app/helpers/rooms_helper.rb
+++ b/app/helpers/rooms_helper.rb
@@ -40,10 +40,9 @@ module RoomsHelper
 
   def room_join_url
     if @room.access_code.nil? || @room.access_code.empty?
-      return request.base_url + @room.invite_path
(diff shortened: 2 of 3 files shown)


lkiesow:
I'm a bit confused about the purpose of this patch. Isn't the purpose of the access key to provide a second factor for authentication? By adding this to the URL, you are defeating that purpose. If you just want a URL you can just not generate an access key, can't you?

Additionally – running with the idea that a URL with key makes sense – putting the key in the URL hash, meaning `http://localhost:3000/adm-xkp-4wc#code=076581` instead of `http://localhost:3000/adm-xkp-4wc?code=076581`, would be preferable since that would mean that the code would stay in the user's browser and would not be sent out each time you access the URL. You could then still enter the code via JavaScript.

MarcelWaldvogel:
@lkiesow Is my assumption correct that your concerns are less with the *ability* to pass the access code as part of the URL than with the proposed *default behavior* of passing it in the URL?

To answer your questions: As (partly) outlined in the feature request #1303 and motivated by #1265, the purpose is:
- Increase the entropy of the room code to reduce unauthorized access by brute-force search aka Zoom-bombing like attacks (maybe better called B⁴ombing here ;-) )
- Not increase the complexity for the user
- Continue to provide an ability for dial-in users
- Ability to change the change the access code without having to recreate the room

Not sending the code directly to the server using a hash would only make a difference if you do not trust the server handling the data, but still trust it to deliver unmodified code. If the server is not trustworthy, we need to take entirely different measures, including certificate-based E2E encryption (which would defeat the purpose of an audio bridge).

If the purpose was only to increase the entropy of the room URL, then adding a further base-36 triplet (e.g. `https://demo.bigbluebutton.org/b/mar-a2d-3df-qxv`) would be my preferred solution. However, the access code right now serves two purposes:
1. additional entropy (with the possibility to pass it via a separate channel; an option which presumably is not used by ≫99% of the users), and
2. the ability to use dial-in.

This modification allows to use all of them and does not prevent anyone from not copying the `?code=XXXXXX` part of the URL and transmitting it in a different path.

lkiesow:
I fully agree with the issue of adding further entropy: Makes sense but would make more sense to enlarge the room's token since that will protect the room from “bombing” regardless of users generating an access token or not.

I'm not sure though what this has to do with dial-in. You are able to dial in without generating an access token at all. So, how is that related?

One additional thing I realized recently is also that people share screenshots of just anything – including their web conferencing tools. In case you visually show the access code directly, that's, unfortunately, part of the image.

MarcelWaldvogel:
AFAIK, dialling in happens as follows:
- The user dials in to a SIP phone number provided by a BBB instance
- The user is then prompted to provide the access code
- This access code is the one which is also used to protect the room and is used to select the room (that explains also why the room owner cannot chose an access code, but the system always has to select one, as only this can guarantee the uniqueness required)

If the above describes the actual dial-in process, then an access code is required for any dial-in enabled room. This triggered my path to reuse this access code for entropy. If I am wrong, then a separate URL triplet would be the better option.

(I have not been able to set up dial-in myself yet, partly due to the instructions being outdated and incomplete; I was planning to submit a pull request to the documentation once I succeeded…)

PS re sending around screenshots: I fully agree that a secure system should not be easily degraded by incompetent users; but I also believe that the normal users' workflow should not be hampered as much that they start to not use the security offered by the system.

lkiesow:
Okay, that assumption is not correct. The access code is not related to the dial-in code. I have got a few servers where dial-in is activated. On one of them I just created a room, generating an access code:

Screenshot from 2020-04-29 01-50-55

Entering the room, I'm then presented with a totally different dial-in code:

Screenshot from 2020-04-29 01-51-06

You also don't need an access code for dial-in. That's also why my assumption was that the access code is meant as a second factor. 


Going forward and taking away the dial-in argument, I think we agree that increasing the size of the generated room identifier thus makes the most sense.

Additionally, from a usability point of view, I also think that making it easier to copy the access code would be nice. But that's something for another pull request.

PR #1332, opened 2020-04-18
Puma Worker Configuration

lkiesow:
We noticed that the current default settings perform very poorly under
load. We managed to literally take down Greenlight during a larger event
when people where accessing the landing page for rooms and when doing
some tests, I was more or less able to DoS Greenlight on my own.

This patch adds a default worker configuration which significantly
improves the situation. The small, 4 core machine I was testing on could
handle about thrice the amount of requests.

While the new default configuration should be reasonably well suited for
most deployments, this patch further allows users to easily configure
the worker pool on their own in the environment file.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -23,14 +23,14 @@ environment ENV.fetch("RAILS_ENV") { "development" }
 # Workers do not work on JRuby or Windows (both of which do not support
 # processes).
 #
-# workers ENV.fetch("WEB_CONCURRENCY") { 2 }
+workers ENV.fetch("WEB_CONCURRENCY") { 4 }
 
 # Use the `preload_app!` method when specifying a `workers` number.
 # This directive tells Puma to first boot the application and load code
 # before forking the application. This takes advantage of Copy On Write
 # process behavior so workers use less memory.
 #
-# preload_app!
+preload_app!
 
 # Allow puma to be restarted by `rails restart` command.
 plugin :tmp_restart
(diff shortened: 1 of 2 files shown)


farhatahmad:
Thanks @lkiesow for the PR. I'd like to get this in the next version. Just need to do some research/testing on some of the implications this may have

lkiesow:
That would be great, thanks.
I think the main implication is that this will take slightly more memory although the effect isn't that noticeable with the `preload_app!` and shouldn't be a problem unless you are working with really restricted memory resources in which case you can still tune this down via configuration.

farhatahmad:
I reverted this in #1667. I still would like to merge this into a different release, but I haven't been able to really test the implications this has and would prefer not to delay the release of a version.

Won't be included in 2.6, but hopefully will get this in 2.7

lkiesow:
@farhatahmad, is there any reason for reverting this? Did you see any problems? There are multiple organizations that have this in production right now and we didn't see any problems. Not including this does mean that we need to build custom versions yet again…

Couldn't you at least change the default configuration instead, making puma use one worker only? In that case we could at least configure this properly. Even though, again, we have seen major issues without this patch.

farhatahmad:
I haven't had any time to test the implications this would have on the variety of different deployment methods. While I have no doubts that you guys are successfully running this in production, I need to make sure that this does not affect other deployments. To do that, I need to delay the release of v2.6 and test everything out. I'd prefer to get v2.6 released with some of the bug fixes and features first, and then explore getting this into the next version

lkiesow:
I opened a pull request #1701 to keep track of this.

Can you tell me what deployment methods you are concerned about? Maybe, I can help testing or provide test scenarios/infrastructure for that.

PR #1389, opened 2020-04-24
Update uk.yml

ostasevych:
Updated Ukrainian translation of strings


Sea444:
For translations we should user https://www.transifex.com/bigbluebutton/greenlight/

ostasevych:
> For translations we should use https://www.transifex.com/bigbluebutton/greenlight/

Thank you! I fixed the translation directly in the file uk.yml. Is it possible to merge it with the existing one?

farhatahmad:
@ostasevych, The PR will be generated automatically once all of the translations are completed

PR #1394, opened 2020-04-24
Fixed CSS class of sign-up button

ByteHamster:
The CSS class of the sign-up button is called `sign-in-button`. This makes it harder to style the button because the same class is used for both, the sign-in button _and_ the sign-up button.


farhatahmad:
This actually breaks some functionality as there is an onlick event handler for ".sign-in-button". Might be better to rename the class to something else and then separately add another class or to change the click handler target (in header.js)

PR #1194, opened 2020-04-04
Adds OpenID Connect.

mapidentity:
Dear BigBlueButton,

Thank you for creating and maintaining Greenlight! This pull request contains an implementation for OpenID Connect  support. We required it because we do not use LDAP but do manage our own users (through Keycloak). Apparently we are not alone here, see also issue #1151. Perhaps this PR will be closed mercilessly per this comment: https://github.com/bigbluebutton/greenlight/issues/1151#issuecomment-607435319, although we do hope you might reconsider.

~~known caveat: OAUTH2_REDIRECT must be set to avoid invalid redirect_uri from Keycloak.~~

**Edit:**
Google and Microsoft specific provider dependencies have been removed. Configurations for most popular social logins have been added. Help with further testing is desired.

PR #1399, opened 2020-04-26
Refine OpenID Connect Authentication from #1194

moschlar:
This is a followup to #1194 which fixes some layout problems (CSS and translations - at least german and english) and makes the uid field configurable.


jfederico:
Once again, thanks for the @moschlar. I'm just waiting for the tests to be fixed so we can merge without remorse.

jfederico (approved):

colfes:
Hey, thanks for merging! In the commit it says : For in-depth steps on setting up a OpenID Connect Login Provider, see: https://docs.bigbluebutton.org/greenlight/gl-config.html#openid-connect

Unfortunately, there is nothing and I'd need some documentation to set it up in my BBB (Docker Version). Im looking for callback-url specifically.

Issue #1268, opened 2020-04-11
Feature request: Add links "Privacy policy" and "Imprint" to footer


Sea444:
**What is missing:**
- Privacy policy
- Imprint

**Desire/implementation proposal:**
- Link "Privacy policy" in the footer (same ligne with footer/powered_by)
- Link "Imprint" in the footer (same ligne with footer/powered_by)
- Add entries in language file
- urls can be defined via the administration interface (similar to the link for the logo)
- if no url is defined, the link is not visible for users

**Example:** 
Provided by Greenlight. 2.5.2 | Privacy policy | Imprint

**Note:**
This information is mandatory in the European Union.

farhatahmad:
Hi @Sea444 

Thanks for creating this feature request. While I definitely agree with you, the problem is that we  haven't written up either of these files specifically for Greenlight. This means I can't include them until they get done. I'll leave this open as a feature request until we update it with atleast the Privacy Policy

Sea444:
Hi @farhatahmad 

I think there's been a misunderstanding. Just because there are no corresponding files in Greenlight included (yet), my suggestion is to include links to external (!) websites that contain this information. For example, if a school uses Greenlight, the school can simply link to the imprint of the school website and the privacy information of the school website. 

In my opinion, this would be a very simple solution to the problem in order to comply with legal requirements. It would be great if this could be included in the next update.


Footer:

![grafik](https://user-images.githubusercontent.com/63501663/79391925-bb15f880-7f72-11ea-8ca5-fe240a13267f.png)

Administration interface:

![grafik](https://user-images.githubusercontent.com/63501663/79393014-b4888080-7f74-11ea-8227-ca32d35f73be.png)

einhirn:
I'm taking a stab at this.

farhatahmad:
Fixed in v2.6

PR #1411, opened 2020-04-27
add customizable Links to Imprint and Privacy Policy

einhirn:
I added two items in "site settings" as suggested in #1268.
When you put an URL in there it gets added as imprint or privacy policy link, respectively, in the page footer. If left empty, it doesn't display.

Let's see what rubocop says 😁

### DIFF
--- a/app/assets/javascripts/admins.js
+++ b/app/assets/javascripts/admins.js
@@ -122,6 +122,18 @@ function changeBrandingImage(path) {
   $.post(path, {value: url})
 }
 
+// Change the Imprint URL to the one provided
+function changeImprintURL(path) {
+  var url = $("#imprint-url").val()
+  $.post(path, {value: url})
+}
+
+// Change the Privacy Policy URL to the one provided
+function changePrivacyPolicyURL(path) {
+  var url = $("#privpolicy-url").val()
+  $.post(path, {value: url})
+}
+
 function mergeUsers() {
   let userToMerge = $("#from-uid").text()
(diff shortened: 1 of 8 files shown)


Sea444:
I don't understand Travis CI check. It fails because we change privacy policy? Yes, that's waht the pull request is about. So everything is ok?

einhirn:
An automated test failed. I added the specification for the test by copy&paste, so I don't know yet why the one for imprint went through and the one for privacy policy didn't. I'll check and double check...

einhirn:
Yep. Changing "privacy policy url" and checking whether the change is reflected in "imprint url" won't work. Travis was right...

mabras:
I do not thinks you should add the `de_DE.yml` text. this will be overwritten from Transifex service. Maybe just English. I am not sure though.

robbi5:
Thanks for the PR. Could we change »Imprint« to something like »Legal«? Imprint is an often used wrong translation: http://s521972252.online.de/wiki/index.php?title=Impressum_-_getting_it_right

Sea444:
Yeah, I looked at the english version of a german website for translation... :-D

Ok, so "Legal" and "Privacy policy" would be correct. Like on apple.com:

grafik

einhirn:
Thanks @robbi5 for pointing that out!

Starting with the least important part, I renamed the branch, hoping it would be reflected here. Nope, it wasn't. So #1421 replaces this PR.

Issue #1423, opened 2020-04-28
Show home page for logged in users


weeman1337:
## Story

As a logged in user I want to see the home page directly,  
so that I don't have to click on "home page".

## Background

If you visit for example a Greenlight instance by https://meetings.example.com/ you see the public index page "Welcome to BigBlueButton...". If I am logged in and want to see my rooms I have to click on the logo or the "home page" nav item. There is no obvious way back to the public index page as logo and "home page" nav item send you to your home page.

In my opinion it makes sense to send logged in users to the home page directly.

farhatahmad:
I agree, that probably makes more sense to in this case. Since the user has already seen the landing page when they first signed in, redirecting straight to the home room would be the correct flow

PR #1453, opened 2020-04-29
Redirect logged in user's to their home page

weeman1337:
Relates to #1423


app/controllers/main_controller.rb
```diff
@@ -22,5 +22,8 @@ class MainController < ApplicationController
   def index
     # Store invite token
     session[:invite_token] = params[:invite_token] if params[:invite_token] && invite_registration
+
+    # Redirect logged in users to the home page
+    redirect_to helpers.home_page if current_user
```
farhatahmad, line 27: I believe this can just be `home_page` and not `helpers.home_page`

farhatahmad:
Done in #1726

PR #1477, opened 2020-05-01
Fix wrong conditional (reported by LGTM)

stweil:
Signed-off-by: Stefan Weil <sw@weilnetz.de>


stweil:
This fixes the [warning from LGTM](https://lgtm.com/projects/g/bigbluebutton/greenlight/alerts), but also changes the functionality (to the desired one, I assume), so some testing might be required.

farhatahmad:
Hi @stweil. Thanks for the PR. I don't even know how that's gone unnoticed, but it's been like that since the creation of the project. I'll need to see what implications this has first

PR #1203, opened 2020-04-05
room block enhancements

sualko:
In the current implementation it's not easy to recognize which room is selected:
2020-04-05 12 19 14 localhost 26cf0e44123b

With some small changes the usability is enhanced (at least in my opinion :smile:) 
2020-04-05 12 08 19 localhost e258fcc7f8e7

I also removed all id attributes inside the room loop.

PR #1490, opened 2020-05-02
Highlight currently selected room

fnwbr:
This is mostly a suggestion coming from our users, feel free to go ahead and close this if you feel otherwise.

When users have access to multiple rooms, and when switching work contexts or windows on their computers they sometimes forget which room is currently selected.
Copying the link to the currently active room then sometimes gives them a link they didn't actually plan on copying.

This was my attempt at fixing that by putting some emphasis on the currently selected room in the list of rooms:

Screenshot 2020-05-03 at 00 18 31
(used green color to showcase that it's working with custom colors too)


Sea444:
Have you seen #1203 ? It was published with greenlight 2.5.5 this week.

farhatahmad:
Yeah something similar to this was implemented in the latest version. Give it a try and let me know what you think

fnwbr:
Nice! Yeah, I was indeed running an older version.

I feel like setting or not setting the corresponding class should still be responsibility of the backend layer, and not reside in a Javascript file. If any of you feels strongly too, I'll happily adjust the code. Otherwise I'll keep this closed. Thanks guys!

Issue #1518, opened 2020-05-06
LDAP auth: Disallow name change


maikmerten:
In a setup using LDAP authentication, users are still able to change their displayed name via profile settings. There does not appear to be a configuration setting to disallow the name change or otherwise ensure that the displayed name matches that delivered by LDAP. Thus names displayed in the list of participants are "unreliable", which makes attribution to real-life persons difficult, facilitating mischievous behavior.

A mechanism to disallow name changes (or to ensure that the LDAP user is displayed in addition) would be very helpful.

farhatahmad:
Hi @maikmerten,

I would go a step further and say that any oauth account (LDAP, Google, Office365) shouldn't be able to edit their name.

maikmerten:
Indeed, I guess there are not many use cases where name edits are beneficial when users are managed in a central way.

farhatahmad:
Fixed in v2.6

relecand:
Hello, 

with the pull requests mentioned above, LDAP users can no longer change their profile data in Greenlight. This feature is useful. But ...

**Problem description**

The profile data (from LDAP) are only generated once when the user is created in Greenlight. If the name changes (e.g. by marriage) or the email address or the LDAP mapping, then there is no way to change the profile data in Greenlight.

**Possible solution**

I think the best solution would be to sync the profile and LDAP data when a user is logging in into Greenlight.

What do you think?

All the best,

Christian

PR #1533, opened 2020-05-07
Readonly Name

mrkeksi:
User should not be able to change his name if user is managed centrally. #1518


farhatahmad:
Thanks for giving it a go, but I'm already working on this. Also requires some backend changes since users can just bypass the readonly attribute

Issue #1259, opened 2020-04-10
"You will automatically join when the meeting starts." never moves the user to the meeting when it starts


davidpesce:
**Environment:**
- Greenlight 2.5.3
- Scalelite v1.0
- BBB 2.2.3 (x3)

**Scenario:**
Anonymous user clicks the meeting link before the moderator gets there. They type in their name (press enter) and receive the message, "The meeting hasn't started yet. You will automatically join when the meeting starts." Even after the moderator starts the meeting, the user is never forwarded to the meeting room. 

The user's console will show:
```
WebSocket connection to 'wss://URL/b/cable' failed: Error during WebSocket handshake: Unexpected response code: 404
```

If the user refreshes the page, it brings them into the meeting.

farhatahmad:
Hi @davidpesce 

Can you consistently reproduce this? Sometimes ActionCable fails, which leads to the user never joining unless they refresh. That being said, there are definitely improvements that we can make on our end to make it more reliable. 

You might also want try setting `CABLE_ADAPTER=postgresql` in your `.env` file to see if it makes a difference

davidpesce:
Yeah, it seems to happen every time I try it. 

Will try that setting in the .env. Thank you!

davidpesce:
Bummer. It didn't fix the issue. I should mention that I've tried this on Chrome (latest) only.

glu000:
I have the same problem but ONLY if Greenlight ist installed on a different server than BBB. If Greenlight and BBB is on the same server it works perfectly fine. Any ideas how to debug this issue?

davidpesce:
@glu000 thats a good point to make! My greenlight is on its own instance as well. I don’t have any installed on the bbb instance. Maybe that’ll help troubleshoot the issue.

farhatahmad:
Greenlight does have a section in the nginx to handle the /cable request. Maybe there's something wrong with that? Here's the default settings we include:
```

location /b/cable {
  proxy_pass          http://127.0.0.1:5000;
  proxy_set_header    Host              $host;
  proxy_set_header    X-Forwarded-For   $proxy_add_x_forwarded_for;
  proxy_set_header    X-Forwarded-Proto $scheme;
  proxy_set_header    Upgrade           $http_upgrade;
  proxy_set_header    Connection        "Upgrade";
  proxy_http_version  1.1;
  proxy_read_timeout  6h;
  proxy_send_timeout  6h;
  client_body_timeout 6h;
[...]

PR #1577, opened 2020-05-13
fix for user stuck on waiting room after meeting starts

davidpesce:
This resolves issue #1259 where user is stuck in waiting room even after the moderator has started the meeting.


farhatahmad:
@davidpesce Finally coming around to this. I think we've solved this through some coding changes so this is no longer needed

PR #1581, opened 2020-05-13
table-responsive overflow problem

juniorojha:
table-responsive, by default, hides the overflow which is not ideal in this situation where when room names or user names are lengthier, the manage button (three vertical dots) disappears from the page.


farhatahmad:
I believe this may already be fixed in another PR (for release 2.6). Also, inline styles probably isn't the best way to correct something :)

PR #1601, opened 2020-05-15
Enhance Room OpenGraph Metadata

moschlar:
This seeks to improve the Rich Preview that most instant messaging services show when you post an url:
photo_2020-05-15_14-11-11
For a first iteration, I chose to just add the room name to the title tags but I am open for feedback on this (both in terms of the actual title string but also the technical implementation part - not a RoR expert).

I would also like to change the description string into something more appropriate for the average end user that will see this (who supposedly doesn't really care about "Greenlight [being] a simple front-end for your BigBlueButton open-source web conferencing server [...]").

Does anyone have thoughts or feedback for this?

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -14,7 +14,7 @@
 %>
 
 <!DOCTYPE html>
-<html>
+<html prefix="og: http://ogp.me/ns#">
   <head>
     <% if Rails.configuration.google_analytics %>
       <!-- Global site tag (gtag.js) - Google Analytics -->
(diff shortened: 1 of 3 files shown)


hng:
@moschlar This is a good idea, thank you. The title is a general problem with Greenlight, every page just says „BigBlueButton“ so I think it would be great to solve this for all pages, not just the meeting pages. Maybe you could also create a issue for the missing titles, so that this issue is not lost?

Code wise I would suggest to set the „- BigBlueButton“ part if title is present globally in the application layout instead of in every template, less repeation if we want to have better titles for other pages too.

moschlar:
@hng
> Code wise I would suggest to set the „- BigBlueButton“ part if title is present globally in the application layout instead of in every template, less repeation if we want to have better titles for other pages too.
You mean like this?

hng:
@moschlar 
> You mean like this?

Yes :)

I think it would be nice to set the title for every view with this PR so that this is done. I can't push to this branch.

Also this PR is still set as a draft, maybe change that?

moschlar:
@hng I rebased on current master and removed the draft status.

farhatahmad:
Starting throwing a weird 500 error after merging so I reverted this. I'm taking a slightly different approach to this in #1853 (though heavily influenced by this PR)

moschlar:
> Starting throwing a weird 500 error after merging so I reverted this. I'm taking a slightly different approach to this in #1853 (though heavily influenced by this PR)

That's  weird, I've been using this patch in our instance since I've posted this PR.
But nevermind, I'm definitely not well versed in RoR, so thanks anyway! :wink:

PR #1644, opened 2020-05-20
GRN2-301: Added more validation on profile image

### LINKED ISSUE #1559: User-based denial of service via /edit
Due to missing input validation an authenticated attacker with access to a BigBlueButton account for a limited amount of time is able to cause a denial of service condition by setting the avatar URL to https://demo.bigbluebutton.org/gl/u/logout (in the demo case).

The user is not able to use the account anymore and is logged out as soon as a login attempt is made. Unless the avatar URL is changed, the account stays inaccessible.

bigblue1

bigblue

farhatahmad:
Added validation for the profile image (will prevent all existing "invalid" images from loading)

fixes #1559

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -94,7 +94,7 @@ class SessionsController < ApplicationController
     login(user)
   end
 
-  # GET /users/logout
+  # POST /users/logout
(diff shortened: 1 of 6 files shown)


farhatahmad:
fixes #1559

PR #1649, opened 2020-05-21
GRN2-300: Changed workflow for users who can't create rooms but have shared rooms

### LINKED ISSUE #1297: Role who only see Shared Rooms
We are planning the release of BBB /Greenlight at several schools in Germany. But we don't like the way the students get the rooms (via a link in an e-mail written separately by each teacher). A role that also sees shared rooms would be great. Or replace "can't create rooms" completely (So you don't have to adjust the share menu for a room). Roles that can create rooms also have no input for room URL or ID, so if the Student has not been assigned a room, they can see an input field. #1186 

What the role should not have.
- no own room #1278
- just no possibility to create rooms
Nice to have:
- cannot see the link to share (of course, it's still in the address bar)
- cannot delete a shared room 
image

### LINKED ISSUE #1521: Problems joining shared room when no permission to create rooms is granted
I created a role in which the users assigned to this role cannot create rooms.
For this reason, I have deactivated the "Create rooms" option for the newly created role.
After assigning a user to this role, the user cannot create rooms. However, he can join
other users' rooms if he knows their ID / URL. So far so good, works!

However, if someone shares a room with users assigned to this role, the user cannot join
the shared room or any other room whose ID / URL the user knows.
If the user tries to enter the room with ID / URL, nothing happens.

Expected behavior: The user can join the shared room as a moderator.
The same behavior as for a user assigned to the standard user role.

Furthermore the shared room appears on the desk of the user and he can use and modify it like it was his own room, even though he cannot create rooms by himself.

 Greenlight. 2.5.5

farhatahmad:
fixes #1297 
fixes #1521

### DIFF
--- a/app/controllers/concerns/authenticator.rb
+++ b/app/controllers/concerns/authenticator.rb
@@ -46,8 +46,10 @@ module Authenticator
                           internal_error_url, not_found_url]
       url = if cookies[:return_to] && !dont_redirect_to.include?(cookies[:return_to])
         cookies[:return_to]
-      else
+      elsif user.role.get_permission("can_create_rooms")
         user.main_room
+      else
+        cant_create_rooms_path
       end
 
       # Delete the cookie if it exists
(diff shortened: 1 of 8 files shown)


mrkeksi:
Thanks you. :) Is there a way that people/students who don't have the right to create rooms but who have shared a room don't start the conference as moderator and don't have the right to start sessions on their own? Maybe as an extra role (is moderator)

PR #1671, opened 2020-05-26
GRN2-290: Update gems and update to Rails 5.2.4.3

farhatahmad:


aguerson:
Hi @farhatahmad,
Why did you modify the files 
app/controllers/sessions_controller.rb
config/puma.rb 
spec/controllers/password_resets_controller_spec.rb 
spec/controllers/users_controller_spec.rb 
 spec/factories.rb 
spec/rails_helper.rb 

Is it mandatory to greenlight branch 2.6 alpha ?
My greenlight on production is 2.5.6 and I only change the Gemfile and the Gemfile.lock.
Did you find some bug without you modifications in the others files ?
Regards,
Aurélien.

farhatahmad:
@aguerson The reason those files were changed was because of some new deprecation warnings caused by updating the gems. If you update Rails, you should run both rubocop and rspec to make sure things are running as expected

PR #1701, opened 2020-05-28
Puma Worker Configuration

lkiesow:
We noticed that the current default settings perform very poorly under
load. We managed to literally take down Greenlight during a larger event
when people were accessing the landing page for rooms and when doing
some tests, I was more or less able to DoS Greenlight on my own.

This patch adds a default worker configuration which significantly
improves the situation. The small, 4 core machine I was testing on could
handle about thrice the number of requests.

While the new default configuration should be reasonably well suited for
most deployments, this patch further allows users to easily configure
the worker pool on their own in the environment file.

---

This pull request is identical to #1332 which got reverted for lack of testing but should still hopefully go into Greenlight at some point.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -23,14 +23,14 @@ environment ENV.fetch("RAILS_ENV") { "development" }
 # Workers do not work on JRuby or Windows (both of which do not support
 # processes).
 #
-# workers ENV.fetch("WEB_CONCURRENCY") { 2 }
+workers ENV.fetch("WEB_CONCURRENCY") { 1 }
 
 # Use the `preload_app!` method when specifying a `workers` number.
 # This directive tells Puma to first boot the application and load code
 # before forking the application. This takes advantage of Copy On Write
 # process behavior so workers use less memory.
 #
-# preload_app!
+preload_app!
 
 # Allow puma to be restarted by `rails restart` command.
 plugin :tmp_restart
--- a/sample.env
+++ b/sample.env
@@ -268,7 +268,7 @@ CABLE_ADAPTER=postgresql
 DEFAULT_REGISTRATION=open
 
 # Preupload Presentation Storage
-# 
+#
 # By default, if Preupload Presentation is enabled for rooms, presentations are uploaded locally to ~/greenlight/storage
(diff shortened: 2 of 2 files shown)


lkiesow:
Rebased this to fix the merge conflicts.

Any chance this can get merged soonish? This has been open for quite a while and it seems like no one did find any problems with it. We have this in production at several places in different environments for about four months now. I know it got reverted to be able to do a few more tests but it has also already been two months since then.

farhatahmad:
Was planning on getting it done as part of 2.7, but due to a feature taking much longer than expected, I haven't yet been able to take a look at it. Probably won't be able to look at this until we start planning for 2.8 release

farhatahmad:
Finally had a chance to take a look over this. If possible can you set the default to 1 and then I can take a look at scaling up the default in a later version?

lkiesow:
@farhatahmad, updated the pull request. Fixed the merge conflicts. Adjusted the default to be one worker as requested. Hope this works for you. Let me know if you want any other changes.

trs80:
You need to add this as well, otherwise the "automatically join when the meeting starts" feature doens't work.
# If WEB_CONCURRENCY > 1, use postgresql(async) for CABLE_ADAPTER
# CABLE_ADAPTER=async

lkiesow:
I can add that to the environment file as documentation but that should also be the default, isn't it?
So, what about adding this:

```
# If WEB_CONCURRENCY > 1, use postgresql(async) for CABLE_ADAPTER which
# is the default for Greenlight in production mode.
# CABLE_ADAPTER=async 
```

farhatahmad:
I'm actually planning on switching the default CABLE_ADAPTER to postgres. We recommended everyone to change to Postgres back in v2.4 (I think) so everybody should be using it. I'll do it in a separate PR

farhatahmad (approved):

aguerson:
Hi guys,
Just to be sure
If I set
WEB_CONCURRENCY=2
I need to set
CABLE_ADAPTER=async

Am I right ?

farhatahmad:
@aguerson If WEB_CONCURRENCY > 1, set make sure CABLE_ADAPTER=postgres (which should be the default on most installations since v2.4 I think)

PR #1732, opened 2020-06-03
Allow the rails log level to be set in .env for production

hng:
**Problem**

Greenlight sets the rails log level to "info" for production. The log level needs to be lower for privacy reasons or set to debug for debugging.

**Solution**

This PR adds a RAILS_LOG_LEVEL to .env that can be set to the desired log level, it is only used in production and defaults to "info" as before.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -132,7 +132,7 @@ Rails.application.configure do
     "#{severity}: #{msg} \n"
   end
 
-  config.log_level = :info
+  config.log_level = ENV["RAILS_LOG_LEVEL"].present? ? ENV["RAILS_LOG_LEVEL"].to_sym : :info
 
   # Prepend all log lines with the following tags.
   config.log_tags = [:request_id]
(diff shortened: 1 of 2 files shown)


farhatahmad:
@hng Thank you for the PR. Instead of doing this as an ENV variable which would require a restart everytime, I went ahead and added a dropdown to the site settings as part of #1858 which will allow you to dynamically change the log level without a restart

hng:
@farhatahmad Thanks for picking this up! For our usecase an ENV variable is much preferred as it allows us to set this setting while deploying Greenlight instances via Ansible. Having settings (only) in the database makes this hard for us (we would need to log in to every instance and change the settings), we've had this problem also with other settings. Maybe we can find a solution for that?

hng:
@farhatahmad I think it would be also good for privacy reasons to set the log level for production to warn instead of info as it used to be. I couldn't figure out if this was changed by your PR.

PR #1774, opened 2020-06-09
Add additional Features for Room setup

einhirn:
Hi, we were asked whether we could persist some room settings, so I added this code to allow the following settings on a room:

- LocksettingsDisableMicrophone
- LocksettingsDisableWebcam
- WebcamsForModeratorOnly

Let's see what needs to be fixed up.


mrkeksi:
Works so far with the fix from Travis (remove spaces after the slash and add one below before the slash)
- is it desired that the settings are hidden by default? I mean, I think that's good, after all, the room settings are getting more and more full of settings. 
- I have a strange behaviour when I look into a room setting that has an option checked and then without clicking on the next room (without reloading) but only calling up its room setting. Then the setting is visually checked there as well. Is this understandable what I want to express?
- In the admin page the translation texts are still missing

einhirn:
>     * is it desired that the settings are hidden by default? I mean, I think that's good, after all, the room settings are getting more and more full of settings.

The "sample.env" is extended to enable the settings by default, but if you didn't change your .env they stay disabled by default.

jibon57:
@einhirn thanks for additional features. Will you please add the ability to set `maxParticipants` for room too :)

einhirn:
That's not a toggle value, I can't just copy&paste. I guess it would be beyond me, for now. You're welcome to try yourself.

jibon57:
@einhirn thanks, will try

matiasilva:
> 
> 
> @einhirn thanks, will try

To implement this you'd need to:
* Add a room setting and store that in the JSON obj in the DB
* Edit the `create` room API call with that maxParticipants value

jibon57:
@matiasilva thanks for hints. I never work on ruby nor I have experience but I think it's a good feature.

matiasilva:
Me neither :) I didn't speak any Ruby or know Rails when I first got started with BBB like 2 months ago? It's just a matter of time and practice.

farhatahmad:
Thanks for the PR @einhirn. At this time, I'm really trying to avoid adding a huge amount of configuration options as the more there are the worse the user experience is. I think the number of settings we have is already too much, so I'm trying only to add the absolute necessary. Since these toggles are already available in the BigBlueButton settings, I'm not sure they qualify under "absolute necessary"

einhirn:
Well, to expand on my initial comment, here's our use case:

Some of our lecturers want to invite participants that aren't enrolled at our university.  So they can't use our LMS but instead have created a greenlight room for their online sessions. Now, of course they can lock webcam sharing and microphone participation in BigBlueButton after starting the room, but because BBB-Sessions aren't persistent they need to do that each time. They thought that to be annoying and asked for a solution - this is what I came up with.

I could very well live with these settings being disabled by default, so every greenlight user could customize what settings are essential for their use cases. At our place, I'd think that "everyone can start" and "all join as moderator" are at least two settings that we could disable because no one here uses (or should use) them...

mrkeksi:
I understand both opinions.  I think more presets are great for all moderators who otherwise have to set their own settings for every meeting.  (Would also wish me more like: may not use private chats). 
 
I also understand that there are now slowly becoming too many settings.  On the one hand, I think it's a good idea to make the settings optional by the admin.  On the other hand I think that all settings will be turned on. 
 
That's why I said from the beginning to summarize the settings.  This is not immediately understandable but you could add a small help popup. 

https://codepen.io/MrKeksi/full/jOWLaXj

But removing settings is not a good solution.

basisbit:
a typical solution could be to only provide the typically used few settings just as done currently, but then also provide an "advanced" / additional settings subpage.

einhirn:
Hmm. Not sure about the failing rubocop and travis builds - can't figure out whether I really broke something.

farhatahmad:
At the moment, we've decided against adding any new room settings until we we make significant changes to the current structure

PR #1003, opened 2020-03-13
GRN2-278: Moved maintenance window to admin settings

etiennevvv:

PR #1775, opened 2020-06-09
Maintenance banner moved to admin site

### LINKED ISSUE #1598: [PR Question] Moved maintenance window to admin settings
Why was this #1003 PR closed? It was a super easy way to announce maintenance. can you reopen the request yourself if the creator does not respond?

mrkeksi:
close #1598

Thanks to #1003

### DIFF
--- a/app/helpers/theming_helper.rb
+++ b/app/helpers/theming_helper.rb
@@ -36,4 +36,8 @@ module ThemingHelper
   def user_color
     @settings.get_value("Primary Color") || Rails.configuration.primary_color_default
   end
+  
+  def maintenance_banner
+    @settings.get_value("Maintenance Banner")
+  end
 end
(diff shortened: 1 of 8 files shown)


app/helpers/theming_helper.rb
```diff
@@ -36,4 +36,8 @@ def privpolicy_url
   def user_color
     @settings.get_value("Primary Color") || Rails.configuration.primary_color_default
   end
+
+  def maintenance_banner
```
farhatahmad, line 40: Don't think this check is actually needed since only admins can see this page anyways

app/views/admins/components/_settings.html.erb
```diff
@@ -203,6 +203,23 @@
       </div>
     </div>
   </div>
+  <% if can_edit_maintenance_banner %>
```
farhatahmad, line 206: Same comment as above

app/helpers/theming_helper.rb
mrkeksi: i think u mean the above file, right?

farhatahmad: Oh yeah you're right. Thanks for the quick fix :)

PR #1777, opened 2020-06-10
Desgin for Manage Users Tabs

### LINKED ISSUE #943: The user filter and invite components are too close
Currently when I want to invite a user, I write his email in the search field and hit the `Invite User` button. Then I remember that those components are for two different actions.

It would be nice to rethink the way they are being displayed and make them distinct.

image

mrkeksi:
Hope this closes #943

with invite:
image

without
image

### DIFF
--- a/app/assets/stylesheets/admins.scss
+++ b/app/assets/stylesheets/admins.scss
@@ -84,10 +84,8 @@
   color: white !important;
 }
 
-.manage-users-tab {
-  &:hover {
-    cursor: pointer;
-  }
+#manage-users-nav.nav-tabs .nav-item {
+  margin-bottom: -1px;
 }
 
 #merge-account-arrow {
--- a/app/views/admins/components/_manage_users_tags.html.erb
+++ b/app/views/admins/components/_manage_users_tags.html.erb
@@ -14,20 +14,25 @@
 %>
 
 <div class="form-group mt-n3">
-  <div class="row">
-    <div class="col-12 tags">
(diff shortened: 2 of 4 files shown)


farhatahmad:
Thanks @mrkeksi. Screenshots look good. Going to pull this locally and try it out and if everything works, get it done as part of 2.7

mrkeksi:
If you filter for rolls you just don't have the invite button anymore. But I think that's acceptable, right?
also i'm thinking if the button should be round. but if i change it then you see the continuous line.

mabras:
@farhatahmad @mrkeksi 
Maybe it is off-topic, but I want to know your input on this. Do Think that `menu_buttons` (left column) should be a subnav of tabs? The right column do not have enough space for its content.

mrkeksi:
You mean with all those tables that seem so tight? Unfortunately, I can't think of a better way to do this if this menu is expanded with more submenus.
I am already thinking about tile instead of table layout.

mabras:
I mean the left column (Users, Rooms, Recording, Roles...) should be a horizontal subnav/tabs. Maybe like [this](https://getbootstrap.com/docs/4.0/components/navs/#horizontal-alignment) or [this](https://getbootstrap.com/docs/4.0/components/navs/#tabs) The subnav will be a 12-col row then the content of it in 12-col row. This will give the content more space.

mrkeksi:
I got what you wanted.  But look at this picture.  With so many menu items it just looks weird.  I could then imagine a button to fold the menu like in bbb top left. 
https://cdn.tutsplus.com/webdesign/uploads/legacy/tuts/375_responsive_menu/img/step3.png

PR #1802, opened 2020-06-16
New option to generate names of unauthenticated meeting members

frsteinb:
This adds a new global tri-state option and in case of its value "Optional" also a new per-room option. This will allow to enforce auto-generated user names for those meeting members that did not authenticate to Greenlight. The name scheme in this case is "user-XXXX", where XXXX are random capital letters.

Please note, I am not a ruby programmer and I have no clue how to setup test units. Sorry.


matiasilva:
Concept is sound. Also lowers the barrier to entry for webinar-type sessions where the name doesn't really matter.

farhatahmad:
@frsteinb Thank you for taking the time to submit a PR. While I think it definitely serves a specific purpose, I don't quite see the point of having this option available for everyone. The vast majority of use cases of BigBlueButton involve engagement between the viewer and moderator which would require identifiable names.

Also, 2 additional configuration options to save sometime a couple of keystrokes seems a little bit overkill to me.

frsteinb:
Well, I asked on the dev list, whether this would be of interest and got two out of two positive responses.

I agree that in many cases viewers are expected to be more behaved, but unfortunately, there are some basic courses on our first semester with many students regarding it as a one-way communication where they may remain anonymous. Those guys are wrong, but the need some time to recognize it. When they name themselves penis or vulva, it just brings more silly turmoil.

This might be situation typical for Germany, where student at most universities do not have to pay for their time here and do not recognize what kind of privilege this is. I am sure, our university is not the only one in Germany with such simple minded students. :-(

Regarding the pull request, of course it's up to you and I can live with whatever decision you make. Thanks a lot for this great project.

mrkeksi:
I agree with @frsteinb. In many cases we only have LDAP sources at our schools and do not allow guests as a matter of principle but in some scenarios I can well imagine that the user is not given the choice to use his own name. For example at public screenings? 
But the room setting is getting really full. In another PR the idea with fold-out advanced settings was mentioned. In itself a good thing but what do you think of @farhatahmad's idea with room types? Which are controlled by the administrator like user roles? 
https://gitlab.com/bbbatscale/bbbatscale/-/raw/master/Documentation/Screenshots/5.roomtypes.png

frsteinb:
@farhatahmad: In those "vast majority of use cases" admins could just leave the global option at its default, and everything remains as before. Only if you choose "Optional" the new option becomes available on a per-room basis.

farhatahmad:
Sorry for the long wait - coming back to this now, I'm not quite sure its a feature needed by most users as it seems to cover a very specific use case of Greenlight. I'll be closing this PR, but I will definitely point anyone asking about this to this PR in the future. Thank you for sharing your code @frsteinb

frsteinb:
I do know only the situation of two installation at TU Braunschweig. Here we do have courses with attendees that show some infantile behaviour :-/, so that some (of course, not all) organizers like and use this option.

I do not propose to make the auto-generated names the default. The default should be to leave it globally turned off. So there is no difference at all in that case.

I would like to see this option, that we really like, to be part of the standard repository, so that we do not have to fiddle with our own patch for any upcoming release.

PR #1803, opened 2020-06-16
content edit

ckbsinnogate:


matiasilva:
Your own branding belongs in your own repository. Also, the license at the top of each file must stay as it is, otherwise you're removing BBB's copyright statement. Not cool at all. If you use this publicly or redistribute this source in this form, you're violating the license and could be subject to further implications.

PR #1804, opened 2020-06-17
Add favicon header tags and 32x32 favicon.png

### LINKED ISSUE #1746: No favicon
When following the documentation with docker + nginx and the `/b/` path prefix, the favicon doesn't get loaded - it's not accessible as `//favicon.ico` at the web server root and the pages do not contain a head tag that references the correct location at `//b/favicon.ico`.

moschlar:
Closes #1746

I would add additional sizes as usually recommended (`apple-touch-icon.png` etc.), but I couldn't find good sources for hi-res images here or in the bigbluebutton repo.


farhatahmad:
Thanks for the PR @moschlar 

I think we might be better off using the Rails equivalent for favicon instead of the HTML so that we don't need to deal with the weird relative root stuff. 

See https://medium.com/@josephcardillo/how-to-add-a-favicon-to-your-rails-app-9676336f7006

app/views/layouts/application.html.erb
```diff
@@ -35,6 +35,9 @@
     <meta property="og:url" content="<%= request.base_url %>" />
     <meta property="og:image" content="<%= logo_image %>" />
 
+    <link rel="shortcut icon" type="image/x-icon" href="<%= relative_root %>/favicon.ico">
```
farhatahmad, line 38: See comment on PR

PR #1840, opened 2020-06-24
More Resilient Parsing of Booleans

lkiesow:
This patch makes parsing booleans from configuration settings more
resilient to ensure no accidental errors occur e.g. due to the use of
upper or lowercase letters.

Example: Without this patch, the following configuration would disable
Greenlight accounts which is quite unexpected behavior from a users
perspective:

    ALLOW_GREENLIGHT_ACCOUNTS=True


lkiesow:
Rebased again to the current `master`.
@farhatahmad, anything you think is missing here or any reason for not including this?

farhatahmad (approved):

PR #1296, opened 2020-04-14
Make recording of a session optional - Fix #1163

yanosz:
Fix #1163 by making recording configurable

By this PR, `recording` is among the room-features configured in the sample env. Due to its privacy implications, it has to be enabled explicitly. In addition, the administator can configure recording-options using the frontend. _Unless enabled explicitly, rooms are not recoreded._ **Note: This changes default behaviour.**

If enabled, users can configure rooms to record session. This flag is set on a per-room basis and cannot be changed while the room is active (i.e. a session is running)

PR #1851, opened 2020-06-29
GRN2-xx: Small changes/improvements to the recording settings

farhatahmad:
Changes:
  - Improved wording of strings
  - Small code quality changes
  - Removed "Active Session" warning

The reason I removed the warning is that I believe it is partially misleading/inaccurate. The warning appears if the session is running and the room owner attempts to change a setting. My issue with it is that the room setting should still be changeable even if the meeting is running, even if the change won't apply until the meeting ends and a new one is created. This is how the rest of the room settings work, so making 1 of them work differently than the others could cause some confusion

What I would rather do is maybe provide a more general warning for all settings and not just the recording setting if the room is running, but I believe that can be branched off to a different PR

### DIFF
--- a/app/assets/javascripts/room.js
+++ b/app/assets/javascripts/room.js
@@ -202,7 +202,7 @@ function updateCurrentSettings(settings_path){
     $("#room_require_moderator_approval").prop("checked", $("#room_require_moderator_approval").data("default") || settings.requireModeratorApproval)
     $("#room_anyone_can_start").prop("checked", $("#room_anyone_can_start").data("default") || settings.anyoneCanStart)
     $("#room_all_join_moderator").prop("checked", $("#room_all_join_moderator").data("default") || settings.joinModerator)
-    $("#room_recording").prop("checked", $("#room_recording").data("default") || settings.recording)
+    $("#room_recording").prop("checked", $("#room_recording").data("default") || Boolean(settings.recording))
   })
 }
 
--- a/app/assets/javascripts/room.js
+++ b/app/assets/javascripts/room.js
@@ -155,9 +155,6 @@ function showCreateRoom(target) {
     $(this).attr('style',"display:none !important")
     if($(this).children().length > 0) { $(this).children().attr('style',"display:none !important") }
   })
-
-  runningSessionWarningVisibilty(false)
-
 }
 
 function showUpdateRoom(target) {
@@ -190,9 +187,6 @@ function showUpdateRoom(target) {
     $("#create-room-access-code").text(getLocalizedString("modal.create_room.access_code_placeholder"))
     $("#room_access_code").val(null)
   }
-
-  runningSessionWarningVisibilty(false)
-
 }
 
 function showDeleteRoom(target) {
@@ -209,9 +203,6 @@ function updateCurrentSettings(settings_path){
     $("#room_anyone_can_start").prop("checked", $("#room_anyone_can_start").data("default") || settings.anyoneCanStart)
     $("#room_all_join_moderator").prop("checked", $("#room_all_join_moderator").data("default") || settings.joinModerator)
     $("#room_recording").prop("checked", $("#room_recording").data("default") || settings.recording)
-
-    runningSessionWarningVisibilty(settings.running)
-
   })
 }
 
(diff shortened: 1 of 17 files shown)


ichdasich:
Hm, i think the active session warning had a point to be there, i.e., to prevent funny expectations of users when they change a setting during a session. Furthermore (if like me) you pull the setting to display a dedicated consent form, starting a session with recordings enabled, and then changing it during the session, might break that (i.e.: Disclaimer not displayed even though the session is being recorded)

farhatahmad:
@ichdasich Sorry was still updating the description with my reasoning. The setting can still be changed, but it'll only apply once the room is ended and a new one is created (just like the other settings)

ichdasich:
Hm, so, depends on when it is written to the GL pg; If that happens immediately, my point applies. Otherwise you would have to keep state somewhere, no?

matiasilva:
Another comment I have which I'm not sure if this was touched by the changes here is the "This session is recorded" red text when the user joins. In the PR it was sandwiched between the room name and "You have been asked to join".

ichdasich:
Wasn't that taken out anyway?

mrkeksi:
I think he's trying to tell us that the clue looks too bruised?

image

Whats about a banner?
image
I don't know if it could be implemented that quickly

farhatahmad:
Yeah I'm not a huge fan of the red warning. Maybe something like this would be a better match? 

image

matiasilva:
I like this but with the color differences it's hard to associate the banner with the red join and the banner can somewhat easily go unread. Perhaps some explanatory text below the join button? Not sure.

farhatahmad:
The red join is only due to that being my local Primary Colour. I'd like to avoid too much "intrusion" or else some one will inevitably ask to make that message configurable as well

mrkeksi:
Because blue is not a alerting colour and it’s the most used Color in Greenlight whats about yellow? Alert-warning is More visibility to users then the standard one

yanosz:
Thanks for picking this up and mergint the PR. A few thought's on this PR 

* Having and adequate notice when joining rooms is afair the most debated effect of #1296. I still find it very hard to reach a consensus. IMHO customization very likely to happen; it could help to simpliy customization  w.r.t to all variants proposed here, instead of trying to find a common ground (e.g. have an input-Field in the admin-settings support html-input?).
* Personally, I placed the join notice right beside the invitation, because I think, that the user's actions referes to both. Users are invited to join a room with recording present. Unfornuatly, the default warning colors ("yellow") is quite light, thus hard to read. That's why the text is red. Due to its privacy implications, it's worth to have the user's attention. I don't like the idea of having banner, since the visual distance does not refelect the semantic closeness.
* The refactorings (naming, etc.) appear to be minor, but I'd hesitate to refactor, before having a suitable test coverage. Afair there's almost no coverage for joiner.rb and the JSON / Rails API. I'd be helpful to have a test checking the correctness of each refactoring in this commit.

farhatahmad:
So we've decided to approach this in a slightly different way. By default, everything will continue to work as before. Rooms will be created and recorded without the required consent from the joiners.

I've added a Site Setting called "Require Room Owner and Joiner Consent to Recording". If set to enabled, it displays the option in the Room Configuration tab which then displays it in the Create/Update Room Modal. 

I've also changed the room joiner message to require an explicit checkbox click to confirm that they agree/consent to their info potentially being recorded. (Still trying to find the right wording for the message)

The reason it was done this way, is so that the default behaviour isn't changed for everyone, and it allows each site admin to specify whether they want this feature or not.

ichdasich:
I like the sollution.

yanosz:
Sounds reasonable .... my master thesis is due in about two weeks. Hence I've no time for checking the sources.

ichdasich:
Good luck! If you want an independent kind-of-does-it-for-a-living review, let me know.

mrkeksi:
what about people who have shared rooms they don't get a hint. do you mean the phrase recording_present that does not appeal yet? I find it briefly and succinctly formulated.

farhatahmad:
@yanosz No worries. Best of luck!

defnull:
Is there a reason why there are two settings now? `config.require_consent_default` can only be set via site-settings (not via environment variables), and there is a `recording` flag for `config.room_features` which *can* be changed via `ROOM_FEATURES` environment variable. The second option has no effect as long as the first is not true. I'm unsure where the difference is.

PR #1873, opened 2020-07-04
Preupload translations add

mrkeksi:
Add icon because "Browse" can not be translated
Add lost translation

Are the following bugs known?
file size over 750kb when I want to start the room:
image

file size over 1mb at the upload will get:
Nginx error: 413 – Request Entity Too Large Error 
but the client_max_body_size in the /etc/nginx/site-av.. is set to the default 10mb

### DIFF
--- a/app/assets/stylesheets/rooms.scss
+++ b/app/assets/stylesheets/rooms.scss
@@ -118,3 +118,8 @@
   word-break: break-all;
   white-space: normal;
 }
+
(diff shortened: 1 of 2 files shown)


farhatahmad:
This feature Imis still a proof of concept and is therefore still not ready to be tested yet. Appreciate you jumping on it so quickly, but it'll need to wait until it gets merged into 2.7

PR #1924, opened 2020-07-18
Record title, instead of room name, in the popup

hiroshisuga:
A long record name can be truncated in the recording list. Might be better to show it in the popup.


farhatahmad:
Can you confirm that editing a recording name still works with this PR?

hiroshisuga:
Hi @farhatahmad , 
There is no problem in editing titles.

Issue #1934, opened 2020-07-21
[Feature request] (2.7) Hide recording menu and list when recording is disabled by admin


miztaka:
As an administrator, I want to disable recordings, hide recording menu and list.

In 2.7, an administrator can disable recording completely. When it is disabled, I want to hide recording menu and recording list in home page.

farhatahmad:
Fixed in 2.7

farhatahmad:
Fix was reverted in #1959

PR #1935, opened 2020-07-21
Hide recording menu and recording list when it is disabled

miztaka:
This is PR for https://github.com/bigbluebutton/greenlight/issues/1934.

### DIFF
--- a/app/views/rooms/components/_room_event.html.erb
+++ b/app/views/rooms/components/_room_event.html.erb
@@ -40,6 +40,7 @@
   </div>
 </div>
 
-<% if render_recordings %>
+<% recording = room_configuration("Room Configuration Recording") %>
+<% if render_recordings && recording != "disabled" %>
   <%= render "shared/sessions", recordings: @public_recordings, pagy: @pagy, only_public: true, user_recordings: false, title: t("room.recordings") %>
 <% end %>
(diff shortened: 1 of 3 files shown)


mrkeksi:
https://github.com/bigbluebutton/greenlight/blob/50c2070188a6f731af5d4930b59e90e54c77d703/app/views/rooms/components/_room_event.html.erb#L44

farhatahmad:
@miztaka Thanks for the PR Mitsu. @mrkeksi is correct, there is one more place that recordings could appear as well

miztaka:
@farhatahmad @mrkeksi Thank you for checking this. I changed _room_event.html.erb as well.

miztaka:
Thanks you for checking this @farhatahmad !

mrkeksi:
@miztaka Nothing changes for me at all? Does it just not work for me?

farhatahmad:
Did you set "allow room to be recorded" to `disabled` in your Room Configuration?

mrkeksi:
Of course, I switched, saved and switched again. Nothing happens (my patch-3 branch).

Also, I wonder, if it's room specific, why should "All records" be removed from the header?

miztaka:
hi @mrkeksi , this works when you disable recording in admin console, not room specific. cheers.

farhatahmad:
Reverting this in #1959. Noticed some slightly weird behaviour, but I don't want to delay 2.7 before I can fix it. I'll take a crack at this again in 2.7.1

miztaka:
What problem did you find? I'm happy to fix it.

PR #1941, opened 2020-07-24
LDAP: Add support for domain-based login

benklett:
With the new environment variable LDAP_DOMAIN. It is only used, if the
auth_method user is selected. Then the domain session controller sets
the domain field of the ldap_config which is then used in
blindsidenetworks/bn-ldap-authentication#8 to build the bindRequest
username like so: DOMAIN\USERNAME.


farhatahmad:
With the added support for OIDC, we recommend you switch your authentication provider to use an OIDC server instead. We probably won't be natively supporting LDAP and OAuth for very long

PR #1977, opened 2020-08-03
Allow uploading larger presentations

lkiesow:
This patch allows uploading presentations up to 25MB instead of the
default 1MB. This should prevent users from getting an 401 reply from
Nginx in most cases.


erpsolns:
Hi can you please give directory - folder  -  file path

lkiesow:
The file path this ends up in will depend on your installation. If you stick to the path which `bbb-install.sh` chooses, the `greenlight.nginx` would end up being at `/etc/bigbluebutton/nginx/greenlight.nginx`

farhatahmad:
Thanks for the PR @lkiesow. Any reason you wouldn't just add `client_max_body_size 25m;` to the `/b` rule above instead of creating a whole new rule? 

Also, we should probably set it to 30mb since that's the limit in BigBlueButton. There is some Javascript that checks if the file is >30mb and complains if it is

lkiesow:
I've updated the pull request and set it to 30MB.

The reason I chose to go with a separate block is that this mechanism is also a protection of Nginx for the backend, so that people cannot send huge requests to arbitrary API endpoints. This relaxes that rule just for the endpoint it's necessary for.

farhatahmad:
Ah makes sense. Just going to do a quick test, and then I'll merge it in

PR #2034, opened 2020-08-23
[security] Fix Cross-Site-Scripting Vulnerability

lkiesow:
This patch fixes a cross-site-scripting vulnerability in Greenlight which allowed users to inject code by adding scripts into their names.

I'm not quite happy about reporting the security vulnerability publically this way but I wasn't able to get in contact with you privately. So, finally, I decided this to be the best path forward since the vulnerability is probably not that severe as users need to click on the script in the current user interface for it to be included and activated.

Still, it would be great if you could provide and document a way to reliably report security issues privately. For example by [adding a security policy to this GitHub repository](https://docs.github.com/en/github/managing-security-vulnerabilities/adding-a-security-policy-to-your-repository).

### DIFF
--- a/app/assets/javascripts/room.js
+++ b/app/assets/javascripts/room.js
@@ -122,11 +122,20 @@ $(document).on('turbolinks:load', function(){
         listItem.setAttribute('class', 'list-group-item text-left not-saved add-access');
         listItem.setAttribute("data-uid", uid)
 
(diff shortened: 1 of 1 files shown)


farhatahmad:
> I'm not quite happy about reporting the security vulnerability publically this way but I wasn't able to get in contact with you privately. 

No worries. I appreciate you taking the time. For future cases, please send an email to `security@bigbluebutton.org` for all matters related to security

PR #2091, opened 2020-09-17
Added new rake tasks for create room and for create and share room

coderDem:
I habe added new rake task to create and create and share a room.
Since we needed this for add data to the database.

### DIFF
--- a/lib/tasks/room.rake
+++ b/lib/tasks/room.rake
@@ -47,4 +47,69 @@ namespace :room do
       end
     end
   end
+
+  desc "Creates a user room"
+  task :create, [:roomname, :email, :room_settings] => :environment do |_task, args|
+    u = {
+        roomname: args[:roomname],
+        email: args[:email],
+        room_settings: args[:room_settings]
+    }
+    user = User.find_by(email: u[:email] )
+    unless user != nil
+      puts "User : #{u[:email]} not found"
+      exit
+    end
+
+    userId = user.id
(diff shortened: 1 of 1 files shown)


coderDem:
Hello,

I am in an loop 

```
lib/tasks/room.rake:59:5: C: Style/NegatedUnless: Favor if over unless for negative conditions.
    unless !user.nil? ...
    ^^^^^^^^^^^^^^^^^
lib/tasks/room.rake:88:5: C: Style/NegatedUnless: Favor if over unless for negative conditions.
    unless !user.nil? ...
    ^^^^^^^^^^^^^^^^^
```

if I fix this with unless a get

```
lib/tasks/room.rake:59:5: C: Style/NegatedUnless: Favor unless over if for negative conditions.
    unless !user.nil? ...
    ^^^^^^^^^^^^^^^^^
lib/tasks/room.rake:88:5: C: Style/NegatedUnless: Favor unless over if for negative conditions.
    unless !user.nil? ...
    ^^^^^^^^^^^^^^^^^
```

Don't know how to get out of this.

hng:
@coderDem These are rubocop linting remarks. What it wants you to change I think:

„unless !user.nil?“ should be changed to „if user.nil?“ And „if user != nil“ should be changed to „if !user.nil?“

farhatahmad:
Going to close this as I believe these rake tasks are for a very specific use case. For the majority of users, they won't need rake tasks to create rooms. I'll point anyone looking for this feature to this PR. Thank you for sharing @coderDem

PR #2103, opened 2020-09-20
Enabling data URI as user's avatar

hiroshisuga:
As BBB team starts to work on user's avatar images, it's time to start on the GR side.
https://github.com/bigbluebutton/bigbluebutton/pull/10458

As a starter, I would like to enable using data URI for user's avatar image. By this users can use whatever images they want even if the images are not on the internet (ie stored in local PC). 

This PR suggest just a small change. User has to convert the image on an internet service such as https://ezgif.com/image-to-datauri . In the ideal world we want to drag and drop an image on the setting form, convert image to base64 format (using Base64.encode64?), and use the data URI for avatar images. 

avatarURI

### DIFF
--- a/app/views/shared/_header.html.erb
+++ b/app/views/shared/_header.html.erb
@@ -40,7 +40,7 @@
 
           <div class="dropdown">
             <a href="#" class="nav-link pr-0" data-toggle="dropdown">
-              <% if current_user.image.blank? || !valid_url?(current_user.image) %>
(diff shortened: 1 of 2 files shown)


farhatahmad:
Sorry for the long wait - coming back to this now. I believe this PR offers a bandaid fix over a much bigger problem. Now that we can store files locally (as a result of the Preupload Presentation feature), we should be storing all images locally, instead of using a URL. Going to close this, and hopefully in the near future, replace it with a PR to allow local image storage

PR #2107, opened 2020-09-21
Allow customization of test email recipient

warp3r:
Hi there,

When setting up SMTP through Microsoft email servers, we get the following error message:

```
Checking environment: Passed
Checking Connection: Passed
Checking Secret: Passed
Checking SMTP connection: Failed
Error connecting to SMTP - 501 5.1.5 Recipient address reserved by RFC 2606 [xxxxxxxx.eop-EUR02.prod.protection.outlook.com]
```

this MR allows the system integrator to specify an RFC 2606 - compliant test destination email recipient through the environment variable "SMTP_TEST_RECIPIENT"

regards

### DIFF
--- a/lib/tasks/configuration.rake
+++ b/lib/tasks/configuration.rake
@@ -49,7 +49,7 @@ def test_smtp
   authtype = ENV['SMTP_AUTH'].present? && ENV['SMTP_AUTH'] != "none" ? ENV['SMTP_AUTH'] : nil
 
   smtp.start(ENV['SMTP_DOMAIN'], user, password, authtype) do |s|
-    s.sendmail('test', ENV['SMTP_USERNAME'], ENV['SMTP_TEST_RECIPIENT'])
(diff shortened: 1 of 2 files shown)


lib/tasks/configuration.rake
```diff
@@ -49,7 +49,7 @@ def test_smtp
   authtype = ENV['SMTP_AUTH'].present? && ENV['SMTP_AUTH'] != "none" ? ENV['SMTP_AUTH'] : nil
 
   smtp.start(ENV['SMTP_DOMAIN'], user, password, authtype) do |s|
-    s.sendmail('test', ENV['SMTP_USERNAME'], 'notifications@example.com')
+    s.sendmail('test', ENV['SMTP_USERNAME'], ENV['SMTP_TEST_RECIPIENT'])
```
farhatahmad, line 52: Would be better to replace this with `ENV.fetch('SMTP_TEST_RECIPIENT', 'notifications@example.com')` so that it defaults incase `SMTP_TEST_RECIPIENT` is not set

warp3r: Ok! changed according to your recommendation

farhatahmad (approved):

Issue #2093, opened 2020-09-17
Registration form can be used to send spam


hng:
If registration is open the registration form can be used to send spam by setting the email field to the person to spam and using the name field to include the spam message and a URL.

We've got an abuse warning by our server provider because this happened to us.

farhatahmad:
Should be fixed now

PR #2115, opened 2020-09-22
Basic anti-spam measures: Added honeypot captcha field to registration form with invisible_captcha

hng:
The registration form can be used to send spam, see issue #2093. This adds a honeypot captcha field to the registration form. It's a simple anti-spam/-bot alternative to Google reCaptcha, which can not be used for privacy reasons for some Greenlight users.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -136,5 +136,7 @@ gem 'random_password'
 
 # Adds helpers for the Google reCAPTCHA API
 gem "recaptcha"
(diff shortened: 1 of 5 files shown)


farhatahmad:
Considering the vast majority of applications are currently using Google reCaptcha, I'm not quite sure an alternative method is needed. I'll close this PR, but I will point anyone looking for this to this PR. Thank you for sharing

PR #2139, opened 2020-09-29
GRN2-xx: Switched default cable adapter to postgres

farhatahmad:
Follow up to #1701


aguerson:
Hi guys,
Just to be sure
If I set
WEB_CONCURRENCY=2
I need to set 
CABLE_ADAPTER=async

Am I right ?

PR #2142, opened 2020-09-30
Enter a room by room ID

hiroshisuga:
Currently the moderator can only send a direct link to users to invite them for a meeting. However the link can unintentionally contain a line break in the middle of URL when sent by Email. This has brought many troubles to my students, who have no idea how the correct link would look like.

Similar to Zoom, it would be convenient if users can join a meeting by entering a room ID. 

enterid


kaystrobach:
Idea is superb!

farhatahmad:
While I can see the use of this, I'm not sure the implementation is the best. We've been talking about changing the home screen slightly, so we'll take a look at this sometime in the near future

PR #2199, opened 2020-10-15
Added StartUp Script and installer for Greenlight using Ruby on Rails

HannahLeaM:
Added a start up script and status routine for greenlight/bigbluebutton running localy with ruby on rails. 

This can be more than usefull if your hoster does not support docker (e.g. by blocking it in the kernel level of their VMs). By using rails to run the container and database you can use the service anyway as intended. To ease the use of this implementation I supplied a script based solution for controling the service and retreiving status information. 

Included in the main directory of the edit you will find an installation script for greenlight.sh and the cronjobs in use. After running it, the script can be used from your terminal as part of your $PATH. Also the cronjobs are added if there are non allready.

Dependencies as of now:
  - Ubuntu 16.04 LTS
  - Greenlight must be located at: /root/greenlight
  - Rails must be located at: /root/.rbenv/
  - Port 5000 must be open and used for the greenlight service

### DIFF
--- a/startup/greenlight-startup.sh
+++ b/startup/greenlight-startup.sh
@@ -1,0 +1,278 @@
+#!/bin/bash
+
+#Include NRPE-PreBuild
+. /usr/lib/nagios/plugins/utils.sh
+
+#VAR Definitiions
+RoR_PIDfile=/root/greenlight/tmp/pids/server.pid
+
+#Read User Flags
+while getopts "sHS:rh" option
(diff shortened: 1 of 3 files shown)


farhatahmad:
Thanks for sharing the PR. There are mutliple reasons as to why we chose the Docker approach and therefore will probably not merge this in. I will direct anyone asking for a similar feature to this PR

PR #2201, opened 2020-10-15
Polish translation improvement

MKjanek32:
Updated Polish localization - added new strings that were not translated and corrected language errors.


farhatahmad:
Hi,

Thank you for taking the time to update theses translations.

PR #1334, opened 2020-04-18
Add SAML authentication

yol:
This is a basic reimplementation of #185 on current master. (related FR: #1253)

Now I know that it is not going to be merged probably, but I wanted to make this available to other interested parties so it can at least be pulled into individual installations.

Issue #1253, opened 2020-04-09
[Feature Request] Support SAML2 for authentication


svoeth:
I know this is a duplicate of #180 , but since that issue has been marked closed without merging into production i'd like to reopen this feature request.

HaSHi93110:
I'm so interested for this support

farhatahmad:
There was a PR to add shibboleth support that I had closed for a variety of reasons. If you have a customized version of Greenlight, you could try pull the changes and seeing if you're able to use it to configure SAML. 

https://github.com/bigbluebutton/greenlight/pull/1138

yanosz:
In general, one could thing of implementing authentication in the webserver, whereas greenlight does solely authorization.

Servers such as apache2 support a plethora of authentications options, that basically result in having a REMOTE_USER header parameter.

I guess, that evaluating REMOTE_USER for authentication is a rather simple implementation from greenlight’s point of view, thus keeping the maintaince effort rather low.

einhirn:
Just a bump, so interested parties get notified. There's a patch available that implements some kind of SAML in #1334

farhatahmad:
Going to tag this as `won't implement` for now. For those looking for SAML auth, we recommend you pull the PR #1334 locally and use that.

PR #2270, opened 2020-11-09
Add SAML authentication with secure checks

RomanKosovnenko:
This is a reimplementation of #185 that is based on #1334 on the current master. (related FR: #1253)

The main differences: 
- It supports security checks like a request/response signature validation.
- Support of the Single Sign On & Single Log Out
- Roles mapping
- Redirection after log out

This implementation works well in our project. And we would like to make it available to others. Even if it will not be merged.


farhatahmad:
See https://github.com/bigbluebutton/greenlight/pull/1334#issuecomment-781671709

PR #2281, opened 2020-11-12
fix: health_check SMTP check when no auth required

jgribonvald:
The health_check controller implemented for the SMTP check try to authenticate in all case even when no auth is set and required.

### DIFF
--- a/app/controllers/health_check_controller.rb
+++ b/app/controllers/health_check_controller.rb
@@ -65,8 +65,12 @@ class HealthCheckController < ApplicationController
       smtp.enable_starttls_auto if smtp.respond_to?(:enable_starttls_auto)
     end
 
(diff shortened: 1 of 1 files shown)


farhatahmad (approved):

farhatahmad:
@jgribonvald Can you rebase this to master please?

jgribonvald:
@farhatahmad done

lonesomewalker:
Seems to be still an issue :-/
Have SMTP server which requires no auth because whitelisted IP.
Unfortunately no "forgot password link" is shown -> no emails can be sent.

jgribonvald:
@lonesomewalker I'm not on the latest version so I can't check, but this is working well at least on 2.9.2 version when you set only the domain param, other properties should not be set.

PR #2282, opened 2020-11-13
update to rails 6

mabras:


farhatahmad:
This is probably something we'll want to do on our own as it's basically impossible to verify your changes through just looking at the PR.

PR #2292, opened 2020-11-16
Update to Ruby 2.7

lkiesow:
The currently used Ruby 2.5 is pretty old and is nearing its end of
life [1]. This patch updates the used Ruby version to the current
stable version.

[1] https://www.ruby-lang.org/en/downloads/branches/


farhatahmad:
This should probably change the the version in the `.ruby-version` file aswell

lkiesow:
Done.

mabras:
Maybe you should alson edit `.gitlab-ci.yml` to use the new version.

lkiesow:
I deliberately did not change any CI things since this technically does not change the minimum Ruby version required to run Greenlight. Also, `.gitlab-ci.yml` doesn't actually seem to be used any longer. But then, you could have made the argument for `.travis.yml`

mabras:
@farhatahmad, We have been useing `2.7` With GL for month now. It is error free. How about we include this in `2.7.13`?

PR #2293, opened 2020-11-16
Changed logo to env variable

sebastianberm:
At the moment, the default logo is hardcoded in Greenlight. I want this to be more variable, and I want to be able to set this via the .env file.

This PR is a small modification, to allow the default value to be fed from the .env file.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -137,7 +137,7 @@ module Greenlight
     # DEFAULTS
 
     # Default branding image if the user does not specify one
-    config.branding_image_default = "https://raw.githubusercontent.com/bigbluebutton/greenlight/master/app/assets/images/logo_with_text.png"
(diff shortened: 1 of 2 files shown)


config/application.rb
```diff
@@ -137,7 +137,11 @@ class Application < Rails::Application
     # DEFAULTS
 
     # Default branding image if the user does not specify one
-    config.branding_image_default = "https://raw.githubusercontent.com/bigbluebutton/greenlight/master/app/assets/images/logo_with_text.png"
+    if ENV["DEFAULT_LOGO"].nil?
```
farhatahmad, line 140: You can replace the whole if statement with `config.branding_image_default = ENV["DEFAULT_LOGO"].presence || "https://raw.githubusercontent.com/bigbluebutton/greenlight/master/app/assets/images/logo_with_text.png"
`

sebastianberm: That's correct, however, Rubocop doesn't like the length of that particular line, so I split it up into smaller statements.

farhatahmad: I don't believe rubocop checks the `application.rb`. Tested it locally and rubocop seems to be okay with it

farhatahmad:
See https://github.com/bigbluebutton/greenlight/pull/2369#issuecomment-747782259

PR #2308, opened 2020-11-24
Use the user image for BBB avatar

hiroshisuga:
With this small modification, Greenlight uses the user image URL for the avatar image in BBB meetings. 

This is only possible when the PR 10458 (https://github.com/bigbluebutton/bigbluebutton/pull/10458) is merged to BBB. This has not yet merged to the official release, so you need to do it by yourself.

I submit this PR just for impatient people;-)  This is related to #2103 .

I confirmed that this PR does not affect the BBB's behaviour even without the BBB PR 10458.
avat


farhatahmad:
Going to set this onhold until the feature gets merge into BigBlueButton

rzo1:
This offers the possibility to add **ultra large** images as the input is **not** sanitized or controlled by Greenlight. This is also discussed [in the related BBB issue / pr discussion](https://github.com/bigbluebutton/bigbluebutton/pull/10458) by the following comment of the PR author:

> You need a frontend app like greenlight to sanitize this type of input.

Better approach would be IMHO:

- Upload an avatar in greenlight to local storage (as it is done with presentations)
- Provide a config option to specify validation parameters (size, width, height, etc.) and validate them in Greenlight
- Provide a config option to disable this behaviour at all (or only for specific roles)

hiroshisuga:
> This offers the possibility to add **ultra large** images as the input is **not** sanitized or controlled by Greenlight. This is also discussed [in the related BBB issue / pr discussion](https://github.com/bigbluebutton/bigbluebutton/pull/10458) by the following comment of the PR author:

Exactly.

> * Upload an avatar in greenlight to local storage (as it is done with presentations)

This needs a relatively large change in BBB, and I do not expect it to be done soon.

> * Provide a config option to specify validation parameters (size, width, height, etc.) and validate them in Greenlight

For now this would be the realistic option. In fact, we need this right now, because a user can still put a link to an ultra large avatar image (which will be transferred to every meeting participant at the invitation screen) even in the current form of greenlight. Do I understand correctly?

farhatahmad:
Hi,

To help us better test and review community PRs, we have now implemented a **mandatory** PR template. We would really appreciate it if you were to update the description of this PR to match the template below. 

For more information, please see: https://github.com/bigbluebutton/greenlight#source-code--contributing

The template must be in the following format (it is automatically applied to all new PRs):
```
<!--- 
IMPORTANT
This template is mandatory for all Pull Requests.
Please follow the template to ensure your Pull Request is reviewed.
-->

<!--- Provide a general summary of your changes in the Title above -->

## Description
<!--- Describe your changes in detail -->

## Testing Steps
<!--- Please describe in detail how to test your changes. -->

<!--- Please include screenshots of ALL visual changes. -->

```

hiroshisuga:
Update with the new PR template





## Description
With this small modification, Greenlight uses the user image URL for the avatar image in BBB meetings. In the current development code, BBB has already enabled the avatar display in the chat window.

## Testing Steps
1. set avatar image in GL.
2. start the meeting
3. see avatar in the user panel of BBB

100097512-f335d000-2e9f-11eb-9424-32ec722d3458

app/controllers/concerns/bbb_server.rb
```diff
@@ -54,6 +54,9 @@ def join_path(room, name, options = {}, uid = nil)
     join_opts = {}
     join_opts[:userID] = uid if uid
     join_opts[:join_via_html5] = true
+    if current_user
+      join_opts[:avatarURL] = current_user.image
```
farhatahmad, line 58: This should be `join_opts[:avatarURL] = current_user.image if current_user.image.present?`, that way, we only set the option if the user has an image set

farhatahmad:
Thank you for the quick change @hiroshisuga. Just some discussion around the large avatar. We've now added a timeout on the Greenlight side to prevent a bad image from crashing the page. I'm not quite sure how Greenlight can actually validate any of this info without changing the images to local storage. I will check with the BigBlueButton devs to see if they can make any changes on their end

app/controllers/concerns/bbb_server.rb
hiroshisuga: Yes, it's a good idea to make sure that nothing unexpected will happen in the future.

hiroshisuga: Didn't pass the rubocop check, complaining that current_user undefined..
Should I do something like:
`join_opts[:avatarURL] = current_user.image if (defined? current_user) && current_user.image.present?`

hiroshisuga: I reverted it to 'join_opts[:avatarURL] = current_user.image if current_user && current_user.image.present?'
This should be fine even if Rubocop complains.

app/controllers/concerns/bbb_server.rb
```diff
@@ -54,6 +54,7 @@ def join_path(room, name, options = {}, uid = nil)
     join_opts = {}
     join_opts[:userID] = uid if uid
     join_opts[:join_via_html5] = true
+    join_opts[:avatarURL] = current_user.image if current_user && current_user.image.present?
```
farhatahmad, line 57: You can do this in 1 if statement: `if current_user&.image.present?`

farhatahmad:
The reason the tests are failing is because `bbb_server.rb` does not have access to `current_user`. The solution is to see where `join_path` is called and add the options there. In this case, that would be `rooms_controller.rb` and `joiner.rb`.

hiroshisuga:
Hi @farhatahmad , I believe now everything is fine. Thank you for the respect to my (very small piece of) contribution, even if you could have done it by yourself!

app/controllers/concerns/joiner.rb
```diff
@@ -60,11 +60,11 @@ def join_room(opts)
       opts[:mute_on_start] = room_setting_with_config("muteOnStart")
```
farhatahmad, line 61: Can you just set the avatar here? `opts[:avatarURL] = ....` 

That way you won't need to make any changes to `bbb_server.rb`

app/controllers/rooms_controller.rb
```diff
@@ -180,6 +180,7 @@ def start
     opts[:mute_on_start] = room_setting_with_config("muteOnStart")
     opts[:require_moderator_approval] = room_setting_with_config("requireModeratorApproval")
     opts[:record] = record_meeting
+    opts[:avatarURL] = current_user.image
```
hiroshisuga, line 184: @farhatahmad , do we need to check here if current_user.image is present? It seems unnecessary to me.

app/controllers/concerns/joiner.rb
```diff
@@ -60,10 +60,10 @@ def join_room(opts)
...
 
       if current_user
+        opts[:avatarURL] = current_user.image
```
hiroshisuga, line 63: @farhatahmad , do we need to check if current_user.image is present here? It seems unnecessary to me.

hiroshisuga:
I changed the PR according to your suggestion. Do the changes make sense?

app/controllers/concerns/joiner.rb
hiroshisuga: > That way you won't need to make any changes to bbb_server.rb

No change at all? I thought we need to set join_opts[:avatarURL] anyway. Please look at my corrections and let me know if they make sense to you.

farhatahmad:
@hiroshisuga Code looks good. My one concern is with the issue described above by @rzo1. I think we may need to tackle that first before going ahead and merging it

ichdasich:
With the upstream part in BBB now being there, iirc, this might be something to follow up on. I still agree with @farhatahmad's assertion, though, that this needs some pre-filtering, so nobody throws a 4GB image at BBB instances.

hiroshisuga:
Yes, we need something to prevent the upload of super huge picture. As far as I remember, somebody proposed a more generic function to upload data to GL, not just referring to a URL. 

However, even in the current implementation, we can filter the image URL by checking the size of the picture. I wonder if it's demanded just for a workaround...

farhatahmad:
Closing in favour of https://github.com/bigbluebutton/greenlight/pull/2860

PR #2315, opened 2020-11-28
Allow Searching for Rooms

lkiesow:
If a user has a lot of rooms, finding the correct one can be somewhat
annoying and it would be great to be able to search for or filter the
rooms in the room list.

This patch adds a very simple search functionality for this. The search
bar is hidden as long as a user has not more than six (two rows in
desktop mode) rooms. If the number of rooms exceeds this limit, a search
field is shown to quickly filter the list.

#### Search bar in action

greenlight-search

#### Search bar appears when you have more than 6 rooms

greenlight-search-7

#### No search with ≤ 6 rooms

greenlight-no-search-2

### DIFF
--- a/app/assets/javascripts/room.js
+++ b/app/assets/javascripts/room.js
@@ -347,3 +347,16 @@ function checkIfAutoJoin() {
     $("#room-join").click()
   }
 }
+
+function filterRooms() {
+  const search_term = document.getElementById('room-search').value.toLowerCase(),
+        create_room = document.getElementById('create-room-block').parentNode,
+        container = document.getElementById('room_block_container');
+
+  container.childNodes.forEach(r => {
+    let text = r.innerText;
+    if (r != create_room && text !== undefined) {
+      r.style.display = (text.toLowerCase().indexOf(search_term) < 0) ? 'none' : 'block';
+    }
+  })
+}
--- a/app/assets/javascripts/room.js
+++ b/app/assets/javascripts/room.js
@@ -142,6 +142,9 @@ $(document).on('turbolinks:load', function(){
     $("#remove-presentation").click(function(data) {
       removePreuploadPresentation($(this).data("remove"))
     })
+
+    // trigger initial room filter
+    filterRooms();
   }
 });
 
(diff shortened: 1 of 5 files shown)


config/locales/de_DE.yml
```diff
@@ -578,6 +578,7 @@ de_DE:
     shared_access_success: Raum erfolgreich geteilt
     shared_access_error: Beim Teilen des Raums ist ein Fehler aufgetreten
     start: Starten
+    search: Nach Raum suchen…
```
mabras, line 581: This auto-generated by Transifex service.

lkiesow: That's true. I assumed that the languages will always be overwritten so that modifications like this won't be a problem while this might help for testing if translations actually work. I can remove this if you want me to?

farhatahmad: Yeah - please remove it. Makes merging translations late easier

app/assets/javascripts/room.js
```diff
@@ -347,3 +347,16 @@ function checkIfAutoJoin() {
     $("#room-join").click()
   }
 }
+
+function filterRooms() {
```
farhatahmad, line 351: Might be kind of picky on my part - but I would prefer if this was written using the jQuery syntax, since that is how most of the application's javascript is written

app/views/rooms/show.html.erb
```diff
@@ -82,6 +82,21 @@
       </div>
     </div>
 
+    <% if current_user.ordered_rooms.length > 6 %>
+      <div class="input-icon invite-link-input">
+        <span class="input-icon-addon">
+          <i class="fas fa-search"></i>
+        </span>
+        <input id="room-search"
```
farhatahmad, line 90: This can all be in one line

app/views/rooms/show.html.erb
```diff
@@ -82,6 +82,21 @@
...
+               type="text"
+               placeholder="<%= t("room.search") %>"
+               class="form-control w-100"
+               onChange="filterRooms()"
+               onKeyUp="filterRooms()"
+               style="margin: 50px 0 25px 0">
```
farhatahmad, line 96: Might want to avoid inline styles and use the bootstrap classes instead

farhatahmad:
Thanks for the PR @lkiesow. Not to nit-pick too much, but for the sake of consistency, I would prefer if the search bar was the same style as the other search bars we use

lkiesow:
> I would prefer if the search bar was the same style as the other search bars we use

No problem. You mean the style you have at the bottom of this picture (the recordings search):

greenlight-search-7

Do you want the bar to be placed at the same location I put it but just styled differently? Or do you want it smaller and placed in the top-right corner (to the right of Tokyo and above start in the picture above)?

farhatahmad:
I think the location is good. Just make it more narrow, and styled the same way as the recordings search

lkiesow:
This should hopefully fix the issues.

farhatahmad (approved):

farhatahmad:
Much better - Thank you @lkiesow

PR #2353, opened 2020-12-09
update name on omniauth

### LINKED ISSUE #2347: Update information from LDAP
If `cn` (or `displayName`) of a user is changed in LDAP Greenlight does not pick up these changes.

How do I force Greenlight to update these changes?

Kind regards
Timm

Davka:
Currently the name is not updated when logging in, which is problematic if the name has changed in the LDAP.

The pull request solves the problem

fixes #2347


farhatahmad:
A permanent option is not the solution for this. It used to be like this, before everyone complained and it was switched to only take the information the first time. Ideally there needs to be a ENV variable that specifies whether to update the info everytime or just the first time

PR #2369, opened 2020-12-15
Being able to set colors from env file

sebastianberm:
At the moment, we're doing experiments with automatic deployments of Greenlight.
We want to be able to give users the possibility to get into an immediately fully configured instance.

This will require us to be able to setup the coloring scheme as well from the env file.

The old/normal functions will not be impacted by this, as the value is only overwritten if it's present.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -140,13 +140,13 @@ module Greenlight
     config.branding_image_default = "https://raw.githubusercontent.com/bigbluebutton/greenlight/master/app/assets/images/logo_with_text.png"
 
     # Default primary color if the user does not specify one
-    config.primary_color_default = "#467fcf"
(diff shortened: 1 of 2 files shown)


farhatahmad:
On second thought, I'm not quite sure this is the direction we want to go with Greenlight. Adding tons off ENV variables the system admin the change the default variables seems slightly overkill considering theres a pretty simple work around to it. After setting up the system, you can run the following line of code to set the value to whatever you want through the terminal:

`docker exec greenlight-v2 bundle exec rails c Setting.first.features.find_or_create_by!(name: "FEATURE_NAME").update(value: "FEATURE_VALUE", enabled: true`

For `FEATURE_NAME`, set any of the features you'd like to change ex: "Branding Image" "Primary Color" "Primary Color Lighten" "Primary Color Darken"

For `FEATURE_VALUE` set it to the value you would like

PR #2400, opened 2021-01-04
Use more recent ruby docker base image

defnull:
The Dockerfile is currently pinned to Ruby 2.5.1, but the most recent patch release is 2.5.8. Docker deployments of greenlight might miss some important security fixes that way.

Gitlab CI/CD tests are actually run on `ruby:2.5` not `ruby:2.5.1-apline`. Tests and actual base image should be the same, if possible.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ruby:2.5.1-alpine AS base
+FROM ruby:2.5-alpine AS base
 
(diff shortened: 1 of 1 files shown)


defnull:
Okay, I'll close this in favour of #2292. But if `.gitlab-ci.yaml` is not used anymore, it should be removed, no?

Issue #1452, opened 2020-04-29
Support Moderator PIN


dranjor:
Currently in greenlight rooms can be protected by a pin. Guests who know this key can enter.

BigBlueButton-API additionally supports a "moderator password".

<attendeePW>jExISZJITFRW</attendeePW>
<moderatorPW>WeaBtgHgDZV</moderatorPW>

It would be nice if greenlight would support this, too. Guests who know this key will be promoted to moderator role.

RomainNakedCat:
After we deployed BBB, I directly received this request from one user.

JGallot:
+1 for this implementation

K1L0BYTE:
+1 [Join as a moderator with a moderator password for a specific room]

zechmeister:
@farhatahmad I am going to implement support for moderator passwords for my application. Of course, I'd like it to become a part of the this repo as well (as there seems to be some kind of demand for it). 

However, I am not sure about the best approach. Should I:
a) always show the option to create an moderator password,
b) hide it in the room specific options or 
c) also put it in the general admins/room_configuration?

kaystrobach:
Use case as a secretary i want to prepare rooms in order to allow my team to have virtual meetups. I need to get 2 URLs, one for the moderator and another one for the guests, that will join the meeting

zechmeister:
@kaystrobach with this PR #https://github.com/bigbluebutton/greenlight/pull/2413 you have one URL, but moderators can use a special access code in order to join the meeting with moderator privileges.

kaystrobach:
That’s fine for me as well.

To reduce the brute force risk, i’d argue to provide different links for moderators and guests and use 2 pin codes.

zechmeister:
I agree, security could easily be improved. 

However, I am still waiting for a Review/ any Feedback on the PR. At this time, I am not sure the maintainer want to support this kind of feature?

kaystrobach:
@ffdixon can you give us your opinion please. Thank uou.

ffdixon:
Thanks for your pull request @zechmeister.  

I'll leave it to @farhatahmad for the review.  At the moment we are trying to do frequent (and smallish) updates to Greenlight to fix any performance and security issues that arise.   We want to settle it down as it's starting to be used quite broadly. 

There is also bit of a backlog of features, so without making any commitment at the moment, we promise to review your pull request and give feedback.

farhatahmad:
This is done in v2.8.1

PR #2413, opened 2021-01-08
Add optional moderator codes

zechmeister:
Sometimes, you want to grant users moderator rights, without them creating accounts and the admins individually sharing the room access (The Feature was also requested in this issue: https://github.com/bigbluebutton/greenlight/issues/1452). A simple solution is to use moderator codes – similar to room access codes, but for moderators instead of _regular_ users. Bigbluebutton already supports this, with this PR greenlight does as well! ✨

**By default, the feature is disabled in the settings:** 
image

**The usage of the moderator codes is very similar to the regular access codes:**
image

image

**Rooms can still be joined through the same form. If an moderator code is entered, the user joins as an moderator and starts the room if necessary:**
image

I had to touch quit a lot of places in order to implement the feature. So I am super excited for your feedback!


farhatahmad:
Hi @zechmeister,

We're looking at implementing this now, and this PR is fantastically done. There are some small changes I'd like to make, but it's probably easiest for me to merge as is and make the changes myself. Thank you for sharing your code!

jgribonvald:
Hi it's a good feature ! 
Some in my mind a few improvement in the UI should be made. 
When you are invited to join should be improved, entering the moderator code isn't really well positioned. I think there should be a button like "I want to be/connect as moderator" that will open a modal to request the code, or maybe the user want to connect only as moderator directly like we do when a password is required to join...

PR #2445, opened 2021-01-19
Increase Database pool size

symptog:
The required pool size is calculated by `RAILS_MAX_THREADS + #Action
Cable-Workers` (default: `5 + 4 = 9`)
The default pool size of 5 is to low for all required connections.


schrd:
This should be merged. It fixes the problem of slow requests (which are in fact waiting for a free database connection) with 100% CPU load on ruby process.

basisbit (approved):
@symptog thank you. We use this config successfully already for months. Should be merged.

kaystrobach:
Works for us as well ™️

jfederico (approved):

jfederico:
Thanks @symptog

PR #2475, opened 2021-01-27
GRN2-340: Added reCAPTCHA to reset password if enabled

farhatahmad:


lonesomewalker:
Since reCAPTCHA has some issues with GDPR, are there any alternatives? :-|

basisbit:
Apparently there exists a `recaptcha_enabled` setting. @farhatahmad please add some documentation for this and please also add a link/note to the documentation to the release announcement.
In the future, please keep in mind that the majority of the world can not legally use or does not want to use Google services.

farhatahmad:
The info related to it can be found in the `sample.env`:
```
# To enable reCaptcha on the user sign up, define these 2 keys
# You can obtain these keys by registering your domain using the following url:
#
#  https://www.google.com/recaptcha/admin
#
RECAPTCHA_SITE_KEY=
RECAPTCHA_SECRET_KEY=
```

While I agree that using Google here is not ideal, I'm really struggling to find a good, strong alternative to it. I'm open to suggestions if anyone has any ideas

basisbit:
What is the goal of this change?

basisbit:
Could some [proof-of-work mechanism](https://github.com/FriendlyCaptcha/friendly-pow) as well as Nginx based rate limiting be used for Greenlight instances which don't have the reCaptcha API keys set?

farhatahmad:
> Could some proof-of-work mechanism as well as Nginx based rate limiting be used for Greenlight instances which don't have the reCaptcha API keys set?

Yep. This adds another layer of (optional) security at the application level. We always recommend system admins configure their own security rules/protocols to increase the security of the entire deployment

PR #2545, opened 2021-02-20
Added undocumented parameter

kaihen:
We found a undocumented parameter for the `.env` that we would like to share with the community. We are using the parameter for our external installation of greenlight. It's just a little bit hardning/security through obscurity.


farhatahmad:
I'm not quite sure this is an "undocumented" variable. Were trying to avoid adding a ton of variables to a file that we believe already has too many different options. People are free to use this, though I dont imagine it'll get merged

Issue #2568, opened 2021-03-11
User permission for recording


jgribonvald:
On previous release (2.7.8 at least) it was available to define a recording option for some roles only and to give the rigth of recording to some peoples only by activating the consent to true and the recodring to optional. In this case the recording option on room wasn't shown if the user didn't have the role "can_manage_rooms_recordings".

Now the role "can_manage_rooms_recordings" isn't applied when it's set to false.

farhatahmad:
That role has to do with managing the rooms and recordings on the administrator side - not the user side. Im not aware of any changes to that functionality since then, but that wasn't the intended use of it anyways

jgribonvald:
@farhatahmad Thanks for your response.
Ok I see, but do you think there could be a possibility to add a role for that ? We would like that admins set a role "can_do_rooms_recording" to users who can make a recording on their rooms. I can make a pull request for that but I need to know if it would be accepted or not for the feature idea.

farhatahmad:
@jgribonvald coming back to this now, if you're willing to make a PR for it, I'd be happy to review it and potentially merge it if all looks good

jgribonvald:
@farhatahmad the Pull request was here: https://github.com/bigbluebutton/greenlight/pull/2592

jgribonvald:
@farhatahmad rebase and feature available done !

jgribonvald:
PR merged

PR #2592, opened 2021-03-19
Feat: Add user permission for recording

jgribonvald:
Add a role permission 'can_launch_record' to users to set the ability to run recording. By default everybody can record, but we can set to the user role without perm to record and create a new role for those who can.

solve issue #2568 

Please review my code before to watch if I forget something, it's the first time that I modify something around role_permissions and do a database change so I'm not totaly sure to do all needed things. But at least it's working for me.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -180,7 +180,7 @@ class ApplicationController < ActionController::Base
   end
   helper_method :shared_access_allowed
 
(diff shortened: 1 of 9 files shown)


farhatahmad:
This isn't really an ideal solution to this problem. The total solution requires something much more complex imo. See my comment here: https://github.com/bigbluebutton/greenlight/issues/2606#issuecomment-814475191

PR #2602, opened 2021-03-23
GRN2-369: Add IP Address to logs

farhatahmad:


pajowu:
First of all: Thank you for mainting greenlight, a software I often use an really enjoy.

Second: Could you please explain the reasoning behind this change a bit more? This poses major problems for BBB-providers in the european union, which need to comply with gdpr and bbb-providers everywhere that aim to minimize collected data.

PR #2610, opened 2021-03-25
log_with_text1 duplicate image

liviraj:
my local chages


liviraj:
logo with check is changed

PR #2614, opened 2021-03-25
GRN2-XX: Updated main script for CI and Dockerfile

jfederico:


Dockerfile
```diff
@@ -19,6 +19,8 @@ RUN apk update \
     && apk upgrade \
     && apk add --update --no-cache $BUILD_PACKAGES $DEV_PACKAGES $RUBY_PACKAGES
 
+RUN gem install mimemagic -v '0.3.8' --source 'https://rubygems.org/'
```
farhatahmad, line 22: Do we need to manually reinstall the gem? It should be installed from the Gemfile (through the `bundle install`)

jfederico: May be not, that was the message I got from the builder, (literally, copy paste) but that was before adding the dependence to alpine

Issue #1645, opened 2020-05-20
Ability to have a static voicebridge PIN from within Greenlight.


lekker-xr:
Since all issues I am aware of about getting a static telephone PIN in greenlight are closed now, but someone mentioned an idea of how to achieve this without problems with the load balancer, I (re)open the request here, referencing back to issues [#1412](https://github.com/bigbluebutton/greenlight/issues/1412) and [#1572](https://github.com/bigbluebutton/greenlight/issues/1572)

Here is the idea on how to achieve this, posted originally [here in #1572](https://github.com/bigbluebutton/greenlight/issues/1572#issuecomment-629841741):

> Another approach could be creating the 5 digit numeric PIN as a hash of the 9 digit alphanumeric room name. This would be predictable and should be unique across loadbalanced instances, right?

farhatahmad:
This issue is not with the pin number creation, but rather the actual phone number. It is possible, but not without some pretty big changes to Scalelite

See https://github.com/bigbluebutton/greenlight/issues/1412#issuecomment-622579506

lekker-xr:
> This issue is not with the pin number creation, but rather the actual phone number. It is possible, but not without some pretty big changes to Scalelite

But the PIN-issue is _also_ there. And originally the issue #1412 was about the PIN.

We use a server without scalelite and already tried the greenligt-fork which has the static PIN functionality -- but it seemed not to work (I amnot the admin, and the admin has not much time and gave up on a failing `docker compose`).

And we have a static phone number (defined already by our phone-network-provider) but a dynamic PIN.

And a static PIN would be much more barrier-free to those who cannot have internet at the time of conference start.

denzs:
I wanted to open a new issue, but it seems there are more than enough issues covering the voiceBridge topic so i will add my thought here..

denzs:
## Situation

It is not possible to use VoIP-dialin via telephone when running bbb in a clustered environment.
Reason is that the voiceBridge is not predictable.
This is because voiceBridge is not set by greenlight and gets shuffled by the bbb-node everytime the room is created.
The dialNumber cant be predicted because it is more or less bound to the node at the moment and cant effectively be assigned via API.
But it could be set to the same number on every node. So it would be predictable.

I want to show a scenario how you can route SIP-calls into your bbb cluster.

Using static dialNumber(s) and predictable voiceBridge solves this problem!

All you need is a static voiceBridge set by Greenlight..

## Approach

[...]

Issue #1572, opened 2020-05-12
Feature request: Make the telephone-PIN static per room. [Enhancement]


lekker-xr:
This is a re-post of the \[[bigbluebutton feature request number 9482](https://github.com/bigbluebutton/bigbluebutton/issues/9482)\] as a greenlight **feature request**, since it seems that the server already can handle this:

**Is your feature request related to a problem? Please describe.**  
When having only the telephone available, it is difficult to join a conference, because the telephone PIN get's dynamically created when a conference is started. It cannot be shared straig after room creation and configuration.

**Describe the solution you'd like**  
The possibility to have a fixed telephone PIN, easiest to equal the online room access code (if one is configured), which can be pre-shared when people get invited to the meeting.

**Describe alternatives you've considered**    
Phoning friends to quickly log in and retrieve the PIN for me, asking other participants to SMS the pin, ...

**Additional context**  
As it is now, it excludes people who do _not_ have (easy) internet access when the conference starts.

wolmeh:
@lekker-xr Take a look at closed issue #1412 with same the topic. There is a greenlight fork on github by davidpesce which support setting a static voicebridge PIN. Because of compatibility problems with the load balancer scalelite the static PIN feature is not part of greenlight master branch.

wolmeh:
By the way @davidpesce: I tested your latest greenlight fork but when a user starts a conference there is a internal server error code 500 and the conference does not start. Log:

> Error during failsafe response: Couldn't find Room with [WHERE "rooms"."deleted" = $1 AND "rooms"."uid" = $2]
>   /usr/src/app/vendor/bundle/ruby/2.5.0/gems/activerecord-5.2.3/lib/active_record/relation/finder_methods.rb:343:in `raise_record_not_found_exception!'
>   /usr/src/app/vendor/bundle/ruby/2.5.0/gems/activerecord-5.2.3/lib/active_record/relation/finder_methods.rb:109:in `take!'
>   /usr/src/app/vendor/bundle/ruby/2.5.0/gems/activerecord-5.2.3/lib/active_record/relation/finder_methods.rb:89:in `find_by!'
>   /usr/src/app/app/controllers/rooms_controller.rb:303:in `find_room'
>   /usr/src/app/vendor/bundle/ruby/2.5.0/gems/activesupport-5.2.3/lib/active_support/callbacks.rb:426:in `block in make_lambda'
[...]

PR #2739, opened 2021-05-29
SIP support with static PIN for each room

### LINKED ISSUE #1412: Ability to set static voicebridge PIN per conference / conference room in greenlight
see [#8371](https://github.com/bigbluebutton/bigbluebutton/issues/8371) in bigbluebutton issue section.

It should be possible to start a conference with static voicebridge PIN by bbb API: https://bbb.EXAMPLE.tld/bigbluebutton/api/create?name=EXAMPLE&meetingID=EXAMPLE&attendeePW=EXAMPLE&moderatorPW=EXAMPLE&voiceBridge=EXAMPLE&checksum=EXAMPLE

But it is not possible to set a static PIN in greenlight. 

The option to set a static voicebridge PIN which is automatically identical with conference room code would be even better.

The ability to set a static voicebridge PIN is important for conference invitations (so that you can tell the PIN before the day of the conference).

Kind regards

tisoft:
Create a static PIN for each room, to be able to send the SIP dial in data to participants before the meeting is started.

## Description

The SIP PIN is generated from the room name, so it stays the same, even if the sessions is closed. The feature is only enabled, if the voice bridge phone number is set. If it is not set, the current behavior of generating a new SIP PIN for each new session is kept.

## Testing Steps

1. set the `VOICE_BRIDGE_PHONE_NUMBER` environment variable to your SIP telephone number. This value will be displayed in each room.
2. setup the BBB SIP Integration, if you use Scalelite, you can follow the instructions from https://github.com/MBcom/bbb-clustersip.git to setup a central SIP gateway for your BBB cluster
3. Greenlight will generate automatically a static conference PIN for each room and shows it next to the telephone number

image

This PR is an extraction of https://github.com/MBcom/enforce-authentication-greenlight from @MBcom.

Fixes #1412, #1572, #1645


farhatahmad:
Thank you for submitting this PR. Go to mark is an `on hold` until Scalelite natively supports SIP

ThomasAppDev:
We would love to take advantage of the change. Is it already foreseeable if and when it will be merged?

ThomasAppDev:
> Thank you for submitting this PR. Go to mark is an `on hold` until Scalelite natively supports SIP

As the changes also work completly without scalelite, is it necessary to wait until scalelite to support SIP?

ThomasAppDev:
Will the feature be available in Greenlightv2 or v3? Can we already deploy it at our system and when will it be in the official release?

Thank you!

farhatahmad:
This won't be merged into v2, as there will only be security updates to v2. v3 will implement something similar to this in a future release

PR #2764, opened 2021-06-07
Features user settings PR

galupa:
This PR should enable user settings features.
What it does basically create a user specific settings so that user can create personalized setting when they join meetings.
It basically use settings on https://docs.bigbluebutton.org/2.2/customize.html#passing-custom-parameters-to-the-client-on-join.

## Description

What we change on the code:
1. We modify the create room and join room functionality to include the user join setting.
2. We create a new entity on database UserSettings that contains the user join setting, we also implement the functionality that enable users to modify the user settings, including with UI template, so there will be a new menu on the user profile.
3. We add a new setting on the .env file to setup which user join settings from https://docs.bigbluebutton.org/2.2/customize.html#passing-custom-parameters-to-the-client-on-join that we want to include.

Note: we currently specify this to only enable the skip check audio, skip video preview, auto share webcam and listen only mode. To enable more, just add more user join settings on the .env file and some additional entry on locales also might be needed

## Testing Steps

To test this PR:
1. migrate the database so that user settings table is created on the database
2. update the .env file to enable which user settings to enable, also the default values. example:
```
JOIN_SETTINGS_FEATURES=userdata-bbb_skip_check_audio,userdata-bbb_skip_video_preview,userdata-bbb_auto_share_webcam,userdata-bbb_listen_only_mode
JOIN_SETTINGS_DEFAULT_VALUES=false,false,false,true
```
3. run Greenlight. Login to greenlight. Go to profile --> settings. change some settings and try to join a room.

Note: we also added some unit tests

Screenshot from 2021-06-08 10-13-03

Please let me know if more details are needed


farhatahmad:
Hi @galupa, thank you for resubmitting this PR with more detailed information. I'll take a look at this PR as part of the next major release

app/controllers/concerns/bbb_server.rb
```diff
@@ -54,6 +54,9 @@ def join_path(room, name, options = {}, uid = nil)
     join_opts = {}
     join_opts[:userID] = uid if uid
     join_opts[:join_via_html5] = true
+    options.select { |a, _| a.to_s.starts_with?("userdata-bbb") }.each do |k, v|
+      join_opts[k] = v
+    end
```
farhatahmad, line 59: Can you use filter here instead of select?

app/controllers/rooms_controller.rb
```diff
@@ -180,6 +180,11 @@ def start
     opts[:mute_on_start] = room_setting_with_config("muteOnStart")
     opts[:require_moderator_approval] = room_setting_with_config("requireModeratorApproval")
     opts[:record] = record_meeting
+    opts[:voice_bridge] = @room_settings["voiceBridgePin"]
```
farhatahmad, line 183: This needs to be removed

app/controllers/users_controller.rb
```diff
@@ -77,6 +79,30 @@ def change_password
     redirect_to edit_user_path unless current_user.greenlight_account?
   end
 
+  # GET /u/:user_uid/join_settings
+  def join_settings
+    redirect_to root_path unless current_user
+  end
+
+  # POST /u/:user_uid/join_settings
+  def update_settings
```
farhatahmad, line 88: I would rename this to `update_join_settings`

app/controllers/concerns/bbb_server.rb
farhatahmad: Can we also use descriptive variable names instead of `a` `k` and `v`?

app/controllers/users_controller.rb
```diff
@@ -77,6 +79,30 @@ def change_password
...
+    return redirect_to root_path unless current_user
+    settings_params = params.require(:user).permit(@join_settings)
+    @user.update_all_user_settings(settings_params)
+
+    # Notify the user that their account has been updated.
+    return redirect_back fallback_location: root_path, flash: { success: I18n.t("info_update_success") } if
+      @user.save
+
+    # redirect_to change_password_path
```
farhatahmad, line 97: Invalid comment

sample.env
```diff
@@ -184,6 +184,21 @@ RELATIVE_URL_ROOT=/b
 #   recording: Sessions are recorded
 ROOM_FEATURES=mute-on-join,require-moderator-approval,anyone-can-start,all-join-moderator,recording
 
+# Specify which room join settings you would like the users to configure on join / enter room session
```
farhatahmad, line 187: I'm not sure we'll want this configurable at the ENV level. I would prefer to have them at the role level, though I'll need to spend some more time thinking about it

farhatahmad (changes requested):

app/controllers/concerns/bbb_server.rb
galupa: Ah thanks for checking

sample.env
galupa: Ah thanks, let us know if you want to move it on other level, we will change it

galupa: Hi @farhatahmad , is there any update on this? Have you decide where we should put the configuration? Or maybe I misunderstand something?

app/controllers/concerns/bbb_server.rb
galupa: I have make changes for this please review it

app/controllers/rooms_controller.rb
galupa: I have make changes for this please review it

app/controllers/users_controller.rb
galupa: I have make changes for this please review it

farhatahmad:
Can you rebase and resolve the conflicts please? Then I'll take another stab at reviewing this

galupa:
Thanks, just rebased this one, please check it again when you have time

PR #2783, opened 2021-06-15
Accessibility fix generate access code

galupa:
This PR should enable keyboard users to be able to generate or reset access codes. It also enable users with visual impairment to know what the generated code is.

## Description

The changes on this PR are as follows:
1. The dice icon on room settings to generate access code is accessible by keyboard, so the user is able to use tabs to go the icon and can use space or enter to generate the access code
2. The trash icon also accessible by the keyboard user, user can use tabs to access the icon and can use space or enter to reset / disable the access code
3. Also add some aria label on the icons so screen reader can see it
4. also make the generated code notified by screen reader.
5. Implement above also for the moderator code

## Testing Steps

To test this:
1. Go to home page, select any room and go to room settings
2. After the modal pop up, users should be able to go to generate access code icon / button by using only tabs / keyboard.
3. User also can go to the trash icon / button to remove the access code
4. Users can use enter / space to press the button / icon
5. Those steps also applied for the moderator code

Not much changes on UI, but the buttons / icon should be able to be selected by only using keyboard
Screenshot from 2021-06-15 11-57-53
Screenshot from 2021-06-15 11-57-42

### DIFF
--- a/app/assets/javascripts/room.js
+++ b/app/assets/javascripts/room.js
@@ -50,6 +50,9 @@ $(document).on('turbolinks:load', function(){
       showDeleteRoom(this)
     })
 
+    // For keyboard users to be able to generate access code
+    generateAccessCodeAccessibility()
+
     $('.selectpicker').selectpicker({
       liveSearchPlaceholder: getLocalizedString('javascript.search.start')
     });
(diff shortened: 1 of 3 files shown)


farhatahmad:
Thanks for all the PRs @galupa. I'll get through them all sometime this week

app/assets/javascripts/room.js
```diff
@@ -50,6 +50,36 @@ $(document).on('turbolinks:load', function(){
       showDeleteRoom(this)
     })
 
+    // For keyboard users to be able to generate access code
```
farhatahmad, line 53: Same as for the other PR - can this be moved into a separate function?

galupa: Sure, thanks for reviewing the codes, I have made the changes, please check it again when you have time

farhatahmad (approved):

PR #2784, opened 2021-06-15
Manage access accessibility

galupa:
This is for accessibility fix on Share Room Access pop up modal.
This PR enable screen reader to be able to read the search results when the user is searching other user.

This also include fix for when user search on the modal and press escape, the modal is not closing, it will only close / hide the search autocomplete input. Currently when user press escape when searching user, it will close the entire modal.

## Description

Basically there are 2 changes on this:
1. Adding some aria properties on the manage user modal (search user). Since the search results element is generated by Javascript library, we add the changes on the room.js
2. Disable propagating escape key action when searching user. Currently when we search user on manage user, it will exit / cancel the search and close the modal, this changes should only exit / cancel the search.

## Testing Steps

To test this, screen reader probably needed, we usually uses Ubuntu buit in screen reader or orca.
1. Go to Greenlight dashboard, choose room and go to manage user
2. Manage user modal is shown and try to search any users
3. Press escape on keyboard
4. The modal should not closed but the search should be canceled / exited.

There is no noticeable changes on the UI


app/assets/javascripts/room.js
```diff
@@ -46,6 +46,36 @@ $(document).on('turbolinks:load', function(){
       showUpdateRoom(this)
     })
 
+    // share room pop up accessibility
```
farhatahmad, line 49: Just one comment - can we create a function at the bottom with this code called `manageAccessAccessibility` and put all of this code there? Then we could replace all this code with a simple function call

galupa: Thank you for reviewing, I have updated the code, please check it again if you have time.

PR #2787, opened 2021-06-17
Fixed voiceBridge (5-digit number) for phone call

miriyusifli:
## Description
If voiceBridge number is not specified then BigBlueButton will assign a random unused number for the meeting. Hence, voiceBridge number is changing for each meeting even the room is not changed. Each time new voiceBridge number should be sent to participants. This pull request allows keeping voiceBridge number the same for all meetings of the room. A random unused number is set for the room and it is the same for all the meetings.

By default, the feature is disabled. To enable the feature, "phone-call" should be added to "ROOM_FEATURES" in the .env file.

## Testing Steps

1. Enable phone-call feature on BigBlueButton to allow users to join the meeting via phone call.
1. Add "phone-call" to "ROOM_FEATURES" parameter in the .env file to enable the feature.
2. Open room creation modal, then you will see auto-assigned voiceBridge number (phone-call code)
3. Start the room
4. On the left side, you will see the phone number and fixed voiceBridge number (phone-call code).

phone-call


farhatahmad:
Thank you for submitting this MR. What's the difference between this and #2739 ?

miriyusifli:
Hi @farhatahmad, differences between MRs:
 - This MR doesn't show the phone number and PIN to the outside. They are shown in the room on BigBlueButton side.
 - This MR allows regenerating (changing ) the PIN for the room.

farhatahmad:
Ah I see. I'll mark this `on hold` for the same reason as the other PR

PR #2802, opened 2021-06-24
Fix permission denied bug

MiKDev:
Fixing this issue:

ERROR: for greenlight-v2  Cannot start service app: OCI runtime create failed: container_linux.go:367: starting container process caused: exec: "bin/start": permission denied: unknown


Dockerfile
```diff
@@ -61,5 +61,9 @@ EXPOSE 80
 ARG version_code
 ENV VERSION_CODE=$version_code
 
+# Set executable permission to start file
+USER root
+RUN chmod +x bin/start
```
farhatahmad, line 66: Hmm. Can we change back to the default user after making this change?

MiKDev: Probably yes, You can change user after set +x permission to start file.

this solution based on following link.
https://willmurphyscode.net/2017/08/16/troubleshooting-a-basic-docker-issue-getting-a-script-into-the-container/

if you read this link at the end of article he used following configs.

USER root
RUN chmod +x /start.sh

USER redis
CMD /start.sh

but in my tests, I didn't change the user.

farhatahmad: Hmm ideally, we would do the same thing here. I'd prefer to avoid running all other commands as root

MiKDev: I removed USER root and result was the same, You can check it.

farhatahmad (approved):

PR #2814, opened 2021-06-29
Accessibility fix for remove user access

galupa:
Hi, This PR should enable keyboard user (user that visually impaired) to remove user access from the Share Room Access modal.
By using `tabs` key, the user will be able to get to `x` remove shared user.

## Description

It is just a small changes and all changes are on share room access modal:
1. (room.js) Change `x` (remove shared user) from <span> to <button>, so that it can be accessible using keyboards and have enter key and space key action by default
2. (locales) add remove_user so for aria-label, so screen reader can read it.

## Testing Steps

To test it:
1. Go to shared room access modal
2. If there is no user, add a user
3. User (you) should be able to remove the user that just added by using only keyboard, using tabs and enter / space key.

Not much changes, but if you look closely the `x` button is select able by keyboard now
Screenshot from 2021-06-29 15-31-27


farhatahmad:
I'm seeing weird padding/margin around the button - can you take a look at it?

image

galupa:
Thank you @farhatahmad for reviewing this, I will check the codes based on your comments

galupa:
Sorry I have been a bit busy last week, but I have fix the UI issue for this PR, please check again when you have time @farhatahmad

PR #2836, opened 2021-07-09
Open graph displaying incorrect language bug

hadicheaito1:
## Description

Created a function that checks whether room exists, and calls user_locale with right params in order to determine user language.
Calling determine_locale from application.html.erb to set og:locale meta data

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -95,19 +95,17 @@ class ApplicationController < ActionController::Base
   end
 
   # Determines proper locale to be used by calling user_locale with params based on if room owner exists
-  def determine_locale(room)
-    room.nil? ? user_locale(nil) : user_locale(room.owner)
-  end
-  helper_method :determine_locale
-
-  # Sets the appropriate locale.
-  def user_locale(user = current_user)
-    locale = if user && user.language != 'default'
+  def determine_locale(user)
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -94,6 +94,12 @@ def migration_error?
     render :migration_error, status: 500 unless ENV["DB_MIGRATE_FAILED"].blank?
   end
 
+  # Determines proper locale to be used by calling user_locale with params based on if room owner exists
+  def determine_locale(room)
+    room.nil? ? user_locale(nil) : user_locale(room.owner)
```
farhatahmad, line 99: I would set use this instead here
```
if user && user.language != 'default'
      user.language
    else
      Rails.configuration.default_locale.presence || http_accept_language.language_region_compatible_from(I18n.available_locales)
    end
```

and then update `user_locale` to use the new method. Then, once you've made that change, you should actually be able to call `determine_locale(@room&.owner)` and that should work because of https://stackoverflow.com/questions/36812647/what-does-ampersand-dot-mean-in-ruby#:~:text=It%20is%20called%20the%20Safe,the%20try%20method%20in%20Rails.&text=This%20sends%20the%20my_method%20message%20to%20my_object%20.

farhatahmad (approved):

PR #2860, opened 2021-07-30
Use the user image for BBB avatar #2 (Limit image size)

hiroshisuga:
## Description
This is basically a modified version of #2308, passing the user image of GL to BBB for displaying it as the avatar. However this version limits the size of the avatar so that BBB can avoid being crashed by somebody sending a huge image.

The maximum size can be set by the parameter AVATAR_SIZE_LIMIT in .env file. 

This PR is sent as a workaround until GL implements a more sophisticated system for uploading images.

## Testing Steps
1. Start a meeting with the user image set on GL.
2. See the avatar on BBB.
3. Another user joins the meeting with the user image set on GL
4. See his avatar as well.


hiroshisuga:
I have fixed an error from Rubocop. I forgot to check if the URL is valid.

ffdixon:
Thanks @hiroshisuga for this.

sample.env
```diff
@@ -356,3 +356,6 @@ DEFAULT_REGISTRATION=open
 # For details, see: https://github.com/puma/puma#clustered-mode
 # Default: 1
 #WEB_CONCURRENCY=1
+
+# Avatar image size limit (bytes)
+AVATAR_SIZE_LIMIT=50000
```
farhatahmad, line 361: Lets up this limit to 100000 bytes instead

app/controllers/concerns/bbb_server.rb
```diff
@@ -54,6 +54,7 @@ def join_path(room, name, options = {}, uid = nil)
     join_opts = {}
     join_opts[:userID] = uid if uid
     join_opts[:join_via_html5] = true
+    join_opts[:avatarURL] = options[:avatarURL] if options[:avatarURL].present? && options[:avatarURL_size] < ENV['AVATAR_SIZE_LIMIT'].to_i
```
farhatahmad, line 57: Can we move the logic for validating the image size into the controller instead? 

```
join_opts[:avatarURL] = options[:avatarURL] if options[:avatarURL].present?
```

app/controllers/concerns/joiner.rb
```diff
@@ -18,6 +18,7 @@
 
 module Joiner
   extend ActiveSupport::Concern
+  include ApplicationHelper
```
farhatahmad, line 21: Helpers should only be used in views, so this can be removed (see below)

app/controllers/rooms_controller.rb
```diff
@@ -21,6 +21,7 @@ class RoomsController < ApplicationController
   include Recorder
   include Joiner
   include Populator
+  include ApplicationHelper
```
farhatahmad, line 24: This can be removed (see below)

app/helpers/application_helper.rb
```diff
@@ -114,6 +114,15 @@ def valid_url?(input)
     false
   end
 
+  def url_size(url)
```
farhatahmad, line 117: Can we move this to a private method in the `rooms_controller` instead?

app/controllers/rooms_controller.rb
```diff
@@ -180,6 +181,8 @@ def start
     opts[:mute_on_start] = room_setting_with_config("muteOnStart")
     opts[:require_moderator_approval] = room_setting_with_config("requireModeratorApproval")
     opts[:record] = record_meeting
+    opts[:avatarURL_size] = url_size(current_user.image)
+    opts[:avatarURL] = current_user.image
```
farhatahmad, line 185: I would move the size logic into here 
```
opts[:avatarURL] = current_user.image if current_user.image.present? && valid_avatar?(current_user.image)
```
(see below for `valid_avatar?`)

app/helpers/application_helper.rb
farhatahmad: I would also replace this with something less general.

```
  def valid_avatar?(url)
    return false if URI.regexp.match(url).nil?
    uri = URI(url)
    http = Net::HTTP.new(uri.host, uri.port)
    http.use_ssl = true if uri.scheme == 'https'
    response = http.request_head(uri)
    return response['content-length'].to_i < Rails.configuration.max_avatar_size
  end
```

farhatahmad: Also, avoid accessing the `ENV` variable directly and set it in `config/application.rb` instead

farhatahmad (changes requested):

app/controllers/concerns/joiner.rb
```diff
@@ -60,6 +61,8 @@ def join_room(opts)
       opts[:mute_on_start] = room_setting_with_config("muteOnStart")
 
       if current_user
+        opts[:avatarURL_size] = url_size(current_user.image);
+        opts[:avatarURL] = current_user.image
```
farhatahmad, line 65: I would move the size logic into here

```
opts[:avatarURL] = current_user.image if current_user.image.present? && valid_avatar?(current_user.image)
```

(see below for valid_avatar?)

hiroshisuga:
@farhatahmad , thanks for the great suggestion. I modified the PR accordingly. In my place it is working as expected. Would be great if you can test it as well.

sample.env
```diff
@@ -356,3 +356,6 @@ DEFAULT_REGISTRATION=open
 # For details, see: https://github.com/puma/puma#clustered-mode
 # Default: 1
 #WEB_CONCURRENCY=1
+
+# Avatar image size limit (bytes)
+AVATAR_SIZE_LIMIT=100000
```
hiroshisuga, line 361: @farhatahmad , would you prefer MAX_AVATAR_SIZE here?

hiroshisuga:
I modified the code a bit. It's working as expected.

A notification to users might be necessary in the user profile setting. Or we may have to limit the size of the profile picture as well.

farhatahmad:
This is fantastic. Thank you @hiroshisuga !

farhatahmad (approved):

PR #2868, opened 2021-08-05
Run puma with nonroot user

l00ptr:
## Description
In this MR we make puma / greenlight running under a dedicated user

## Testing Steps
Within the directory :
```
docker build .
docker-compose up -d
```
Then when everything is running, we can list the process runing insde the greenlight container :
```
docker exec -ti greenlight-v2 ps auxf
```
```
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
greenli+   144  0.0  0.0   1668   912 pts/0    Rs+  14:21   0:00 ps auxf
greenli+     1  0.6  2.1 243588 168068 ?       Ssl  14:14   0:02 puma 4.3.8 (tcp
greenli+   120  0.1  2.3 296900 188216 ?       Sl   14:14   0:00 puma: cluster w
```


fireba11:
@l00ptr how will update work? we'll need to run a chown for the persistent data one way or another

l00ptr:
Hi,

That's a good question. I think we don't need to chown any data... For what i know there are no persistent data with Greenlight container (all the persistent information are inside the PostgreSQL container). So we can simply remove the Greenlight container and get the new one... but maybe I am wrong.

fireba11:
greenlight by default (as in using the install script) got two volumes defined in the compose:
    volumes:
      - ./log:/usr/src/app/log
      - ./storage:/usr/src/app/storage

l00ptr:
Yes indeed how did i miss them :'( 

Thx for your attention... So i guess we probably need to add some info documentation about this change.

PR #2888, opened 2021-08-30
Sound notification when meeting started

miztaka:
## Description

As a guest user for a meeting,
I want to be notified by sound when moderator starts the meeting
So I prevent from not be aware of meeting start.

When a guest tries to join the meeting before moderator starts,
you have to wait until meeting starts.
A user sometimes do another job and not aware of start of the meeting.

This tweak makes sound notification in this situation.
The implementation is a bit roundabout to avoid user consent.

## Testing Steps
1. Prepare two windows for owner and guest
2. Guest join the meeting  (meeting not started yet)
3. Owner start the meeting
4. Guest get sound notification and redirected to the meeting session

https://user-images.githubusercontent.com/4747270/131288593-d576763f-427b-4c83-8573-b111a312a7f3.mp4


farhatahmad:
Hmm this is an extremely interesting idea that I've never thought of. I'm not sure if the front-end is the best place for this though. What do you think of BigBlueButton playing a tone when you join instead?

miztaka:
@farhatahmad well, it is nice to have any sound notification only when user is waited for a meeting to start whichever in client or bbb side. Also note that html5 may have process for moderator approval, which I raised a ticket in here: 
https://github.com/bigbluebutton/bigbluebutton/issues/13097

Anyway, our company already introduced this feature and users feedback is very good.

PR #2890, opened 2021-08-31
feat: manage a global duration of sessions

jgribonvald:
## Description

This feature provide a way to set a global duration for all bbb room start.
We can define 3 variables:
* `DEFAULT_MEETING_DURATION`: the default duration of all room - 0 means unset
* `DEFAULT_CONFIGURABLE_MIN_MEETING_DURATION`: the minimum duration that can be set from the admin UI
* `DEFAULT_CONFIGURABLE_MAX_MEETING_DURATION`: the maximum duration that can be set from the admin UI.
These variables can be set from environment, and the `DEFAULT_MEETING_DURATION` can be overriden into the admin room configuration.

Also show on main page of user rooms the duration set by the admin, if the setting is set to a value upper to 0.

A topic on mailing list was open [HERE](https://groups.google.com/d/msgid/bigbluebutton-greenlight/ceecf493-49cc-0715-1308-9ad08d68b7aa%40recia.fr?utm_medium=email&utm_source=footer)
## Testing Steps

set the variables from ENVIRONMENT and also test to set the duration on the admin room configuration site and start a room. Check the param provided to BBB.

### Admin room configuration vue
Sélection_206
### User main room view
Sélection_207


jgribonvald:
Hi folks,

The next step to another Pull request would be to manage a user customisation of the room duration based on a permission. Let me know about this PR before.
Thanks

Issue #2505, opened 2021-02-03
Modifying the defaultWelcomeMessage for rooms


sebastianberm:
For example in Moodle, it's possible to change the defaultWelcomeMessage which BigBlueButton uses to start rooms.
That way it's possible to give custom instructions and so on.

My usecase is I have a lot of different meetings going on with different requirements for the WelcomeMessage, which can be set with the BBB API. I think this is the 'welcome' setting.

Would it be possible to implement a textarea which allows you to set a custom welcome message for each room?

farhatahmad:
This falls under the same problem that a bunch of other room settings fall under. There is just way too many different settings and adding them all will cause a ton of problems from the administration level. Im not sure we've found a proper solution to this yet

sebastianberm:
I understand this might be an issue. Wouldn't it be an idea to have certain defaults editable via the .env file? And allow an 'array' like listing, so people can at least set default values for these variables?

sebastianberm:
@farhatahmad : I've implemented the editable via the .env file in my PR.
This would solve the issues we have in our usecase, but still honor your remark regarding the amount of settings per room.

estalia1:
Custom welcome message as well as attaching the picture is a recurring request from our clients. The need is to:
a) separate visually one lesson from another
b) provide users with some instructions, notes and urls  

This could be separated to a new modal window as it requires some space for the rich text field and resolve the issue mentioned by farhatahmad:
 

> This falls under the same problem that a bunch of other room settings fall under. There is just way too many different settings and adding them all will cause a ton of problems from the administration level. Im not sure we've found a proper solution to this yet

PR #2892, opened 2021-09-01
Feature/defaultwelcome

sebastianberm:
Implements modifier for the defaultWelcomeMessage

## Description
As mentioned in #2505 it allows for the defaultWelcomeMessage to be set on a 'per Greenlight' instance level.
In the particular usecase, this cannot be modified on the BBB nodes itself, but this could also be an initial configuration setting for this, could be on a 'room level' later, but as mentioned in the ticket, this is a bit too much.

## Testing Steps
Copy the sample .env file, uncomment the default value and modify the value. It should be visible when starting a new session.


jgribonvald:
Really interesting !

I guess if something could be made with a I18N solution, because when having a meeting with peoples from several countries it could be really great to have a traduction on several langages!

farhatahmad:
Can this be configurable in the admin settings instead of by env variable?

PR #2906, opened 2021-09-19
GR2-xx: Social uid rake task

farhatahmad:


defnull:
What does this do? `social_uid` and `provider` are two database columns, why should someone want to store the provider name as part of the social id? This would break omniauth user logins if used with an omniauth provider.

Also, there is a `v2.10.0` label on this PR but it was merged into `release-2.9.3-beta.1`. By accident perhaps?

farhatahmad:
@defnull This is a rake task that we'll be using internally to move all users from their current login method to using a centralized KeyCloak login server

As for the tags, that is a miscommunication that I'll sort out

PR #2931, opened 2021-10-01
(fix) Room start button margin

hiroshisuga:
## Description
Remove the px-7 margin from the room start button. When a word on the button is large, the margin can break the button labeling especially with the btn-block.

## Testing Steps
1. open GL as an admin, setting the language in Japanese
3. the letters in the button was broken before this change, but now it looks good

before fix in English it's OK.
Image 2

before fix the start button broken in Japanese.
Image 0

after fix the button is OK in Japanese.
Image 1


hiroshisuga:
I have also changed the translation so that it's not too long.

farhatahmad (approved):

Issue #2946, opened 2021-10-09
[OpenID Connect] user roles supported?


beckerr-rzht:
I am looking for a way to set **user roles** when authenticating via _OpenID Connect_.
Someting similar to `LDAP_ROLE_FIELD` for authentication via _LDAP_.

beckerr-rzht:
I implemented this analogous to `LDAP_ROLE_FIELD` and opened a pull request: #2955

hadicheaito1:
Hello! Unfortunately, we will not be able to devote time into this feature/enhancement request at the moment due to our attention being directed towards Greenlight v3!
If this is something you believe that will be beneficial to v3, then we encourage you to please re-create this request once v3 is up and running.

To find out more information on Greenlight v3, please click the following link:
https://groups.google.com/g/bigbluebutton-dev/c/1nG-k_iTHjo

Thanks for your continued support!

PR #2955, opened 2021-10-13
Added OPENID_CONNECT_ROLE_FIELD

beckerr-rzht:
Authentication via OpenID Connect is extended to allow **user roles** to be specified.

## Description
An environment variable `OPENID_CONNECT_ROLE_FIELD` can be used to define which claim is used to set the user's role. The handling is similar to the LDAP authentication, where the environment variable `LDAP_ROLE_FIELD` is used.

## Testing Steps
First the OpenID server must be extended by a claim `roles`. This is then added to the scope `public` so that it is transmitted during authentication.


jfederico:
As mentioned in #2946 we need to validate this also solves some other large deployments we know are already in production, and perhaps cover the case of mapping roles and other arguments.

flyinghuman:
would be good if this feature will be merged into master.
however, it will overwrite the pre-existing roles. 
An example: an admin-user logged in via openid_connect which also has the roles in bbb and openid_connect: "rule-denied-room-creation" will then gets this role assigned and will loose admin rights.
to prevent this it is neccessary that at least admin-roles are excluded from that:
```
unless auth['info']['roles'].nil?
      roles = auth['info']['roles'].split(' ')

      role_provider = auth['provider'] == "bn_launcher" ? auth['info']['customer'] : "greenlight"
      roles.each do |role_name|
        logger.info("[DEBUG] user has role: #{role_name}")
        role = Role.find_by(provider: role_provider, name: role_name)
        user.set_role(role_name) if !role.nil? && !user.has_role?(role_name) && !user.has_role?("admin") && !user.has_role?("super_admin")
      end
    end
```
see `&& !user.has_role?("admin") && !user.has_role?("super_admin")`

also, the default `roles.split(',')` is maybe not for every provider useful. This could also be achieved via an ENV value which can be set in `.env` Settings File.

PR #2997, opened 2021-12-03
BSN-068: Increased local account password security

farhatahmad:
## Testing Steps

There are a couple cases that need to be tested:
1. Log in with an existing local greenlight account -> you should be redirected to the reset_password page 
2. Reset your password and make sure you can log in again
3. While logged in, mark your password as insecure again `User.find_by(email: ...).update(secure_password: false)`
4. Change your password via the profile password update and make sure that you are not redirected to the reset_password page next time you login
5. Create a new account, then log out. Make sure you are not redirected to the reset_password page when you login


farhatahmad:
I give up lol

farhatahmad:
Finally found the problem. Funny enough, the command `Faker::Internet.password(min_length: 8, mix_case: true, special_characters: true)` does not guarantee numbers or characters in the password. I'm appending "1aB" to all passwords to ensure that the password created is valid

jfederico (approved):

sualko:
Is it right, that every user needs to reset his password even if it meets the new password policy? If that's the case, why? During the login the application should have the plaintext password. In my opinion that would not be very user friendly. Also I would like to see a config option to specify the minimum password length and the required characters.

rottaran:
Many thanks for the efforts but I have a lot of users complaining now, that greenlight keeps them out for the password change in the last minute before important meetings. It's a transition phase yes, but was not nice to discover this from usage instead of release notes. And some configuration option to switch this feature on explicitly in a planned and prepared way would have been super helpful.

PR #3000, opened 2021-12-07
GIT-2541: Fixed inconsistency between room last_session and meeting createTime on create (Closes #2541)

### LINKED ISSUE #2541: invalid The createTime parameter submitted mismatches with the current meeting.
Since updating to greenlight version 2.7.19 which introduced the usage of the createTime parameter in the join-call we see in some cases the error:

invalid The createTime parameter submitted mismatches with the current meeting.

After searching through the logs, we found about 2700 cases of the error from about 368000 joins in the last day.

If the error happens, all the users trying to join the conference get the error.

KH-Amir-TN:
Fixed inconsistency between room last_session in GL and meeting createTime on BBB when a problem occurs during the start of an old room

## Description

BigBlueButton (BBB) recommends the providing of the meeting creation timestamp in the `createTime` property and checks its legitimacy invalidating the join requests in case of a mismatch.

Greenlight (GL) fallows the BBB recommendations and uses the room `last_session` attribute to store the timestamp of meeting creation and provides it upon join requests as a value to the `createTime` property.

Whenever, the creation process fails GL won't update the attributes of the room in question including its `last_session` and `sessions` properties.

Some cases just like the GL timeout causes the creation process to fail on GL but to continue on BBB creating this inconsistency.
Resulting in:
1. The creation of the session on the BBB server.
2. The inconsistency between GL and the BBB server.

The room session creator will observer an external error `BigBlueButton invalid secret or...` which will lead them to retry creating the session which will result upon successful create requests in:
1. GL attempting to create a new session because it doesn't know that the room has one running already (the `is_meeting_running` BBB API call marks a session as "running" only if it had at least one attendee who joined it).
2. BBB indicates that a session exists for the corresponding `meetingID` by a `DuplicateWarning` message.
3. GL receives the duplication alert and proceeds in joining the requester without updating the room attributes.
4. The join fails because of the invalid `createTime` value given by GL that corresponds to the last successful session and not the current one.
5. The end user will be redirected to the join URL in which will observe an error from BBB (This error depends on the BBB and its reverse proxy configuration, some will observe the error `invalid The createTime parameter submitted mismatches...` as an alert banner in BBB's demo page while others will simply observe it in the URL as a query parameter while others will be redirected to a fallback site...).
This will result in preventing anyone from "starting" the room according to GL definition and therefore preventing anyone from joining the session.
This happens with rooms that hosted at least one session because at the moment for backward compatibility BBB still accepts the omitting of the `createTime` which happens automatically with new rooms that has a `nil` `last_session` which explains why a lot of users created new rooms to host the sessions after facing this issue.
A user can only wait for the BBB to implicitly stop the session or an administrator has to manually stop it or empty/update the rooms `last_session` value to resolve this.

## Testing Steps

1. Created several rooms.
2. Started several sessions.
3. Raised exceptions in several parts in the creation logic.
4. Observed the results in GL and BBB server.
5. Confirmed the root cause.

## Solution
Since, after a successful create request GL will have the session data including its `createTime` and also will have access to the room `ActiveRecord` instance in question meaning that without making additional queries to the DB or requests to the BBB API a simple check for the consistency between the room's `last_session` and the `createTime` will result in updating the local value in case of any mismatch despite the cause (A duplicate request isn't an error).
In addition, in this PR the GL timeout can be configured externally through the environmental variable `TIMEOUT`.
NOTE: Configuring the timeout on GL may not seem to solve the frequent timeout issue for some deployments that uses ScaleLite which has its own timeout that requires configuring.
This solution won't solve any inconsistency between GL and ScaleLite timeout but it ensures that even in such deployments and until ScaleLite gets updated any successful create request will resolve the issue.


app/controllers/application_controller.rb
```diff
@@ -17,6 +17,7 @@
 # with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
 
 class ApplicationController < ActionController::Base
+  include ApplicationHelper
```
farhatahmad, line 20: Helpers should only contain code that is used in the frontend

config/environments/production.rb
```diff
@@ -169,4 +169,8 @@
   config.relative_url_root = ENV['RELATIVE_URL_ROOT'] || "/b" if ENV['RELATIVE_URL_ROOT'] != "/"
 
   config.hosts = ENV['SAFE_HOSTS'].presence || nil
+  # Set the timeout
+  config.timeout = ENV["TIMEOUT"] || 10
```
farhatahmad, line 173: You could just add this to `config/application.rb`

farhatahmad: I'd also use more descriptive variables than just `timeoute`

farhatahmad:
Instead of this whole change, why not just patch the gem to increase the timeout from 10 to 20 or 30? This seems like a ton of work to an easily solvable problem

KH-Amir-TN:
> Instead of this whole change, why not just patch the gem to increase the timeout from 10 to 20 or 30? This seems like a ton of work to an easily solvable problem

I believe that you are referring to the `ruby-bbb-api` gem timeout altering that was added in this PR.
If so, I avoided patching the gem because it may brake compatibility for any updates (of the gem) since it diviates its codebase from the actual gem upstream.
I tried to keep the dependencies implementation out of my scope and focus on what could be done through GL codebase.
Do you think it's better to hardcode an increase in the gems timeout?
The `ruby-bbb-api`  gem offers accessor methods on the timeout to simply alter  its value externally.
Having the timeout set by the environmental variable felt the best way to answer some of the users requests who made it clear that a fixed timeout isn't suitable for some deployments.
The extra logic was added to bound the timeout.
The end solution has nothing to do with the timeout injection but after conselting the issue it felt right to enable GL users to define their timeout without having to build a new image.
Do you want me to limit this PR to only solve the duplication issue only and see a better way to enbale timeout altering as a separate concern and maybe add it as a feature just like `Room Limit` ?

config/environments/production.rb
KH-Amir-TN: Totally agree.

app/controllers/application_controller.rb
KH-Amir-TN: You got reason.
My bad, I'll make it an `ApplicationController` method.

farhatahmad:
> > Instead of this whole change, why not just patch the gem to increase the timeout from 10 to 20 or 30? This seems like a ton of work to an easily solvable problem
> 
> I believe that you are referring to the `ruby-bbb-api` gem timeout altering that was added in this PR. If so, I avoided patching the gem (if that what you meant) because it may brake compatibility for any updates (of the gem) since it deviates its codebase from its actual upstream. I tried to keep the dependencies implementation out of my scope and focus on what could be done through GL codebase.
> 
> Do you think it's better to hardcode an increase in the gems timeout and do you mean by patching the gem is to simply doing what I've done but with a fixed timeout rather then injecting it via an environmental variable which leads to filtering, checking and bounding its value (that may led you to feel it overly complex)?
> 
> The `ruby-bbb-api` gem offers accessor methods on the timeout to simply alter its value externally. Having the timeout set by the environmental variable felt the best way to answer some of the users requests who made it clear that a fixed timeout isn't suitable for some deployments. The extra logic was added to bound the timeout. The end solution has nothing to do with the timeout injection but after consulting the issue it felt right to enable GL users to define their own timeout without having to build a new image each time and to keep simple and flexible for all users. Do you want me to limit this PR to only solve the duplication issue only and see a better way to enable timeout altering as a separate concern and maybe add it as a feature just like `Room Limit` (It's starting to feel as the right way to do it)?

I was referring to this: https://github.com/mconf/bigbluebutton-api-ruby/blob/master/lib/bigbluebutton_api.rb#L75

Our the time I've worked on Greenlight, I've learned that it's sometimes best that we just set the limit ourselves. I think we're getting close to the point where we have too many this configurable which just causes more confusion for the administrators. 

2 options here: 
1. Patch the default in the gem (we have a good relationship with mconf and can get them to release a new version if needed)
2. Just hard code a better timeout number without making it configurable by env

KH-Amir-TN:
> > > Instead of this whole change, why not just patch the gem to increase the timeout from 10 to 20 or 30? This seems like a ton of work to an easily solvable problem
> > 
> > 
> > I believe that you are referring to the `ruby-bbb-api` gem timeout altering that was added in this PR. If so, I avoided patching the gem (if that what you meant) because it may brake compatibility for any updates (of the gem) since it deviates its codebase from its actual upstream. I tried to keep the dependencies implementation out of my scope and focus on what could be done through GL codebase.
> > Do you think it's better to hardcode an increase in the gems timeout and do you mean by patching the gem is to simply doing what I've done but with a fixed timeout rather then injecting it via an environmental variable which leads to filtering, checking and bounding its value (that may led you to feel it overly complex)?
> > The `ruby-bbb-api` gem offers accessor methods on the timeout to simply alter its value externally. Having the timeout set by the environmental variable felt the best way to answer some of the users requests who made it clear that a fixed timeout isn't suitable for some deployments. The extra logic was added to bound the timeout. The end solution has nothing to do with the timeout injection but after consulting the issue it felt right to enable GL users to define their own timeout without having to build a new image each time and to keep simple and flexible for all users. Do you want me to limit this PR to only solve the duplication issue only and see a better way to enable timeout altering as a separate concern and maybe add it as a feature just like `Room Limit` (It's starting to feel as the right way to do it)?
> 
> I was referring to this: https://github.com/mconf/bigbluebutton-api-ruby/blob/master/lib/bigbluebutton_api.rb#L75
> 
> Our the time I've worked on Greenlight, I've learned that it's sometimes best that we just set the limit ourselves. I think we're getting close to the point where we have too many this configurable which just causes more confusion for the administrators.
> 
> 2 options here:
> 
> 1. Patch the default in the gem (we have a good relationship with mconf and can get them to release a new version if needed)
> 2. Just hard code a better timeout number without making it configurable by env

Alright, do I increase the timeout to 

> > > Instead of this whole change, why not just patch the gem to increase the timeout from 10 to 20 or 30? This seems like a ton of work to an easily solvable problem
> > 
> > 
> > I believe that you are referring to the `ruby-bbb-api` gem timeout altering that was added in this PR. If so, I avoided patching the gem (if that what you meant) because it may brake compatibility for any updates (of the gem) since it deviates its codebase from its actual upstream. I tried to keep the dependencies implementation out of my scope and focus on what could be done through GL codebase.
> > Do you think it's better to hardcode an increase in the gems timeout and do you mean by patching the gem is to simply doing what I've done but with a fixed timeout rather then injecting it via an environmental variable which leads to filtering, checking and bounding its value (that may led you to feel it overly complex)?
> > The `ruby-bbb-api` gem offers accessor methods on the timeout to simply alter its value externally. Having the timeout set by the environmental variable felt the best way to answer some of the users requests who made it clear that a fixed timeout isn't suitable for some deployments. The extra logic was added to bound the timeout. The end solution has nothing to do with the timeout injection but after consulting the issue it felt right to enable GL users to define their own timeout without having to build a new image each time and to keep simple and flexible for all users. Do you want me to limit this PR to only solve the duplication issue only and see a better way to enable timeout altering as a separate concern and maybe add it as a feature just like `Room Limit` (It's starting to feel as the right way to do it)?
> 
> I was referring to this: https://github.com/mconf/bigbluebutton-api-ruby/blob/master/lib/bigbluebutton_api.rb#L75
> 
> Our the time I've worked on Greenlight, I've learned that it's sometimes best that we just set the limit ourselves. I think we're getting close to the point where we have too many this configurable which just causes more confusion for the administrators.
> 
> 2 options here:
> 
> 1. Patch the default in the gem (we have a good relationship with mconf and can get them to release a new version if needed)
> 2. Just hard code a better timeout number without making it configurable by env

You got reason, I wasn't aware of the long term implications of keeping adding configurable settings to the app.
I'll increase it to 30seconds and point this out in the PR description for those who want to further change the timeout themselves.

KH-Amir-TN:
@farhatahmad requested changes that were made on Jan 15:
1. The `bbb-ruby-api` timeout is now statically set to 30sec and NOT INJECTED EXTERNALLY [CHECK HERE](https://github.com/bigbluebutton/greenlight/pull/3000/commits/9226e0107d76bb33c3aa7d9fee3600fb31fb430f#diff-ed9a79387d769bec09ba88edcc5b72dcc91424b39c7574a9c6716f9b8c634f91R25).
2. An improvement to UX was added on meeting creation by **RETRYING** the `start_session` logic on `join_path` to a maximum of 3 retries which eliminates the need for the user manual subsequent meeting start requests to fix the issue after observing it as an error on client side (caused by the timeout and/or race condition on meeting start [The operation was tested and it is thread safe but some clients will have a 500 error i.e. because of a rolled back transaction] ) and to enhance their overall experience [CHECK HERE](https://github.com/bigbluebutton/greenlight/pull/3000/commits/9226e0107d76bb33c3aa7d9fee3600fb31fb430f#diff-55d05625ee766791453434eb7cb444c3ac6e7edf475ad3fd6dac5a0b26ad46f0R54).
3. An update was made to the `bbb-ruby-api` gem to its latest version `1.8.0` because of noticing that the `BigBlueButton::Exception` inherits directly from the `Exception` class and that this was fixed in latest version [CHECK HERE](https://github.com/mconf/bigbluebutton-api-ruby/pull/42) and [HERE](https://github.com/mconf/bigbluebutton-api-ruby/releases/tag/v1.8.0).

app/controllers/concerns/bbb_server.rb
```diff
@@ -88,12 +83,19 @@ def start_session(room, options = {})
       else
         bbb_server.create_meeting(room.name, room.bbb_id, create_options)
       end
-
-      unless meeting[:messageKey] == 'duplicateWarning'
-        room.update_attributes(sessions: room.sessions + 1, last_session: DateTime.strptime(meeting[:createTime].to_s, "%Q"))
+      logger.warn "Duplicated create request" if meeting[:messageKey] == 'duplicateWarning'
```
farhatahmad, line 86: If it's a diplicate create request, shouldn't we `return logger.warn ... ` so that the rest of the code doesn't run?

KH-Amir-TN: IMHO we cannot consider a duplicated request as a problem on GL and here's why:
We'll still want the start session logic to complete successfully (with no exceptions raised and even on duplicated requests) and with the inconsistency resolved to further complete the **joining** (`join_path` always calls `start_session` before returning a `join_meeting_url`) pipeline.
Halting the session because of duplication will prevent the session starter from joining it resulting in:
1. An error on their side (For no reason I believe, since duplicated requests doesn't indicate a problem on BBB).
2. **The room won't be declared as running** even though the session started on BBB because it had no attendee.
The room session starter will still have to start it resulting in the same error causing a `dead lock` on the room state until its session gets explicitly or implicitly ended meaning that whenever an issue raises causing this inconsistency we'll still face the problem (In fact it would be worst because this time there won't be errors on client side).
A duplicated request doesn't mean an error it only indicates that such request has been made i.e. two users attempt to start the same session BBB won't consider it an error and one if precedes the other will create the session while the other will simply make a duplicate request that will return details about the meeting as expected but with an indication if useful that this isn't this request was only passive.
Considering duplication as an error in such cases won't create a `dead lock` but it would definitely give a bad UX for those who observe a flash message or gets redirected when they were able to join the session despite their duplicate request just like BBB API encourage it.

GhaziTriki: https://docs.bigbluebutton.org/dev/api.html#create

`The create call is idempotent: you can call it multiple times with the same parameters without side effects.`

So the `duplicateWarning` is safe to ignore.

KH-Amir-TN: Exactly,
Thanks @GhaziTriki ,I've been looking for the word **idempotent** it could have helped simplify my explanation xD
The useless updates to the room `last_session` and `sessions` attributes that we should avoid aren't those with the duplicated requests but the ones in which the room attributes is already consistent and that is this PR's solution in a nutshell.

farhatahmad (approved):

Gemfile.lock
```diff
@@ -8,10 +8,10 @@ GIT
 
 GIT
   remote: https://github.com/mconf/bigbluebutton-api-ruby.git
-  revision: 91dc495324a6b7e162773227ec3650f8a5b39c50
+  revision: 803c1e0680eada08740bb068956647bdd36ac7ee
```
GhaziTriki, line 11: There is a problem with commit https://github.com/mconf/bigbluebutton-api-ruby/commit/803c1e0680eada08740bb068956647bdd36ac7ee

It should be 5b803bd640e5144d32b18e039e2bc63c85c0bb4f

GhaziTriki (changes requested):
BigBlueButton version gem must be replaced otherwise the Docker image won't build.

Gemfile.lock
KH-Amir-TN: Yes, that's bizarre It's bundler who updated the `bigbluebutton-api-ruby` dependency.
On the time of creating the PR that was the gem remote revision and the CI pipeline succeeded.
There must be something with their configuration somehow `wpramio` fork got caught as a valid remote revision for the gem.
 @farhatahmad we have to be aware of this since it can happen again and could you check with mconf about it?
Meanwhile I'll reupdate the dependency and make sure that it points to a valid commit on their master branch.

KH-Amir-TN:
> BigBlueButton version gem must be replaced otherwise the Docker image won't build.

Yup, I'll reupdate the dependency and notify you.

Gemfile.lock
KH-Amir-TN: https://github.com/KH-Amir-TN/greenlight/runs/4827195801?check_suite_focus=true#step:9:569
But 5 hours ago after @farhatahmad merged on my forked repo which in turn launched the CI pipeline again failed indicating that **803c1e0** cannot be found on master branch.
https://github.com/KH-Amir-TN/greenlight/runs/4876692555?check_suite_focus=true#step:9:349

KH-Amir-TN:
@GhaziTriki the remote revision for `bigbluebutton-ruby-api` has been updated to **5b803bd** .

app/controllers/concerns/bbb_server.rb
GhaziTriki, line 86: We need the context of the log, room name and user at least if possible

app/controllers/concerns/bbb_server.rb
```diff
@@ -45,8 +45,15 @@ def get_multiple_recordings(meeting_ids)
...
   def join_path(room, name, options = {}, uid = nil)
     # Create the meeting, even if it's running
-    start_session(room, options)
-
+    retries = 0
+    begin
+      start_session(room, options)
+    rescue StandardError => e
+      retries += 1
+      logger.warn "Retrying #{retries} time(s)."
```
GhaziTriki, line 53: Same here we need a clear context for the logs.

GhaziTriki (changes requested):
Need a small improvements for the logs. It would be better to have them with the context.

app/controllers/concerns/bbb_server.rb
KH-Amir-TN: Sure,

app/controllers/concerns/bbb_server.rb
KH-Amir-TN: Sure,

PR #3003, opened 2021-12-08
GIT-2986: Fixed incorrect room recordings title (Closes #2986)

### LINKED ISSUE #2986: Quotes " are displayed incorrectly in the records section
Quotes " are displayed incorrectly in the records section

image

KH-Amir-TN:
…86).

## Description

Some BigBlueButton (BBB) configurations enforce encoding on recordings name, Greenlight (GL) logic on rendering the list of available recordings for a room is to fetch it from the BBB server, format and represent it on the view page.
there was no formatting to the recordings name to decode the encoded characters by the BBB server resulting in having the recordings of a room named 'foo<"bar">' to be presented with title 'foo& lt;& quot;bar& quot;& gt;' (whitespaces after each & were added to prevent GitHub from decoding entities automatically).

## Testing Steps

1) Created different rooms with different names containing some characters that maps to an [html entity](https://www.w3schools.com/html/html_entities.asp).
2) Started and recorded meetings on each room.
3) Checked and confirmed that the rooms with such names will have by default an incorrect recordings title.

image

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -152,4 +152,9 @@ module ApplicationHelper
   def protected_recording?(rec)
     !rec[:protected].nil?
   end
+
+  # Takes any expression and stringify it decoding any html charachter entity if found.
+  def html_entities_decode(expr)
+    CGI.unescapeHTML expr
+  end
 end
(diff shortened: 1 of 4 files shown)


Gemfile.lock
```diff
@@ -519,3 +521,6 @@ DEPENDENCIES
   uglifier (~> 4.2.0)
   web-console (~> 3.7)
   webmock (~> 3.11)
+
```
jfederico, line 524: We don't fix the version of Bundler intentionally so it can be deployed on any environment regardless of the version of bundler they may have

KH-Amir-TN: Alright, I've updated Gemfile.lock

farhatahmad:
Can we use `CGI.unescapeHTML` instead of including a whole new gem here?

KH-Amir-TN:
> Can we use `CGI.unescapeHTML` instead of including a whole new gem here?
Yes, I've just checked it and it works fine.
You got reason, I'll update the `html_entities_decode` method to use the `CGI::unescapeHTML` method, remove the `htmlentities` gem and I'll re-request your review.

KH-Amir-TN:
> Can we use `CGI.unescapeHTML` instead of including a whole new gem here?

Done ^^

farhatahmad (approved):

PR #3018, opened 2021-12-14
GIT-2207: Fixed internal error caused by overflowing pages (Fixes #2207)

### LINKED ISSUE #2207: Minor bug - error 500 - user management
Hello, 

I got error 500 when I reject last pending user on page 2. Issue is reproducible viewing url: http://<server>/b/admins?tab=pending&page=2, when count of pending users is lower than size of page (default is 25). 

    Best regards  Pavel Polacek

KH-Amir-TN:
## Description

Greenlight uses 'Pagy' gem for a simpler and efficient pagination.
[Pagy](https://ddnexus.github.io/pagy/how-to.html#gsc.tab=0) default behavior to handle overflowing pages is to raise the ['Pagy::OverflowError'](https://ddnexus.github.io/pagy/extras/overflow#gsc.tab=0) exception causing the application to render the 500 internal error page caused by the 'rescue_from' method in the ApplicationController.
Many actions in the AdminsController logic uses the 'redirect_back' to redirect the client back to were the initiated the request which leads upon banning,unbanning,undeleting,... last users in the last page to redirect the client back to an overflow page causing Pagy to raise the OverflowError exception and the controller to rescue_from it by rendering the internal error view which explains what occurred with [#2207](https://github.com/bigbluebutton/greenlight/issues/2207).
## Testing Steps

1.  Created many users.
2. Updated their role to have the 'pending' state.
3. Start declining and approving users.
4. Observed the error with approving/declining the last user in the last page other than 1.
5. Confirmed the root cause by accessing overflown pages on several areas in the application (rooms,admin pannel,recordings...).

### DIFF
--- a/config/initializers/pagy.rb
+++ b/config/initializers/pagy.rb
@@ -80,9 +80,9 @@ require 'pagy/extras/bootstrap'
 
 # Overflow extra: Allow for easy handling of overflowing pages
 # See https://ddnexus.github.io/pagy/extras/overflow
-# require 'pagy/extras/overflow'
+require 'pagy/extras/overflow'
 # Pagy::VARS[:overflow] = :empty_page    # default  (other options: :last_page and :exception)
-
+Pagy::VARS[:overflow] = :last_page
 # Trim extra: Remove the page=1 param from links
 # See https://ddnexus.github.io/pagy/extras/trim
 # require 'pagy/extras/trim'


spec/pagy_spec.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
```
farhatahmad, line 1: I'm not quite sure these tests are needed. This just seems to be testing the gem itself, which isn't what we'd want to do

KH-Amir-TN: I have added them to ensure that 'Pagy' is configured the right way.
If it's not configured in the 'last_page' overflow mode it will fail.
As for the last test I agree too.
Do you need me to remove the last test or all of the tests?

farhatahmad: I'd personally just remove them all

KH-Amir-TN: Alright, I've just removed the spec file.

farhatahmad (approved):

PR #3019, opened 2021-12-15
BSN-063: Added login attempt limiting for increase security

farhatahmad:

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -219,6 +219,19 @@ class User < ApplicationRecord
     update_attributes(main_room: room)
   end
 
+  # returns true if the user has attempted to log in too many times in the past 24 hours
+  def locked_out?
+    attempts = failed_attempts.to_i
+    within_1_day = (24.hours.ago..DateTime.now).cover?(last_failed_attempt)
+
+    return true if attempts > 5 && within_1_day
+
+    # reset the counter if the last login attempt was more than 1 day ago
+    update(failed_attempts: 0) if attempts.positive? && !within_1_day
+
+    false
+  end
+
   private
 
   # Destory a users rooms when they are removed.
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_11_28_175028) do
+ActiveRecord::Schema.define(version: 2021_12_15_001240) do
 
   create_table "active_storage_attachments", force: :cascade do |t|
     t.string "name", null: false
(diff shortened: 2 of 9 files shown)


app/models/user.rb
```diff
@@ -219,6 +219,19 @@ def create_home_room
     update_attributes(main_room: room)
   end
 
+  # returns true if the user has attempted to log in too many times in the past 24 hours
+  def locked_out?
+    attempts = failed_attempts.to_i
+    within_1_day = (24.hours.ago..DateTime.now).cover?(last_failed_attempt)
+
+    return true if attempts > 5 && within_1_day
```
basisbit, line 227: Please consider making this magic number configurable.

db/schema.rb
```diff
@@ -145,6 +145,8 @@
     t.integer "role_id"
     t.datetime "last_login"
     t.boolean "secure_password", default: false, null: false
+    t.integer "failed_attempts"
+    t.datetime "last_failed_attempt"
```
basisbit, line 149: Doesn't this cause the failed_attempts counter to just increase, if for example there is one failed attempt every 23 hours, because `last_failed_attempt` gets reset to current time again and again, while `failed_attempts` is not being reset?

Imho this implementation has bunch of corner cases where it will not behave as documented/desired, and needs a rewrite.

farhatahmad: That's why the counter is reset here: https://github.com/bigbluebutton/greenlight/pull/3019/files#diff-9802ca3c9c4cf89904fd44bc114e35ebdf2c5dd3d5b645491e2b253e1afef29bR229

> Imho this implementation has bunch of corner cases where it will not behave as documented/desired, and needs a rewrite.

If you can provide me with those edge cases, I'd be glad to look over the implementation :)

basisbit: The counter is only reset if the last failed attempt is more than 24 hours ago. So, a Denial of Service attack would be to for example generate a list of students email addresses, then lock all of their accounts using 5 invalid login requests each, then once every 23 hours send one invalid request again

basisbit: Alternatively, send one failed login per 23 hours and after 5 days it will resut in account being locked and continue being locked as long as a request keeps coming once every 23 hours

app/models/user.rb
sualko: This is a massiv attack vector for a denial of service. Please add a config option which also allows to disable this behavior.

I would prefer a throttle, with increasing time to a maximum of one hour.

db/schema.rb
sualko: @basisbit is absolutely right and it's even what the comment says:

```
# reset the counter if the last login attempt was more than 1 day ago
update(failed_attempts: 0) if attempts.positive? && !within_1_day
``` 

Edge case: User does not remember his password and tries it several times until he gets block. A couple of hours he tries it again and the timeout starts from the beginning. At the next morning he starts again, which restarts the time once again.

In my opinion the waiting time should be related to the number of failed attempts to a maximum. E.g. the waiting time in minutes could be something like `last_failed_attempt + max(60, failed_attempts * 5)`.

basisbit: @sualko fyi, I already reported that as DoS vulnerability to Blindside Networks. I hope it gets fixed soon-ish.

app/models/user.rb
basisbit: @sualko imho a proper solution, which afaik most of the bug BBB providers already have deployed, is to use for axample Nginx rate limiting in front of these critical URLs of Greenlight.

PR #3031, opened 2021-12-18
GIT-2008: Fixed recording consent circumventing (Fixes #2008).

### LINKED ISSUE #2008: Consent on join can be circumvented by changing the setting while the room is running
For 2.7, #1296 got merged. However, the merged version no longer checks if a room is running when changing the setting. This allows the following issue:
- Enable recordings for a room and start it (with recordings enabled)
- Change the setting to 'disabled'
- Users joining the room are not informed about the session being recorded

The initial implementation by @yanosz did actually check for this. Is there any chance this can also be backported? I already had users ask about whether they could use this mechanic to circumvent the recording disclaimer when I ran @yanosz version in production (where it was not possible to circumvent this).

KH-Amir-TN:
## Description

Enabling the `Require Room Owner and Joiner Consent to Recording` feature in Site settings>Settings in the Organization section will result for rooms with the setting `Allow room to be recorded` enabled to have a consent on their join page.
Any user that wants to attend one of the room sessions has to accept the consent.
Some scenarios makes this consent vulnerable to circumventing. 

**Scenario 1**

1. User enters the waiting page of a room without recording enabled.
2. Room owner (or anyone with the right permissions) allow recordings for the room and starts a session.
3. User joins the meeting without accepting the recording consent.

**Scenario 2**
1. User access the join page for a room without recording enabled.
2. Room owner (or anyone with the right permission) enables recording and start the room.
3. User joins the session without accepting the recording consent.

**Scenario 3**
1. Room owner (or anyone with the right permission) enables recording for a room and starts a session.
2. Room owner (or anyone with the right permission) disable recording for the room.
3. User access the join page and joins the room without accepting the recording consent.

## Testing Steps

All of the mentioned scenarios above.
## Solution explanation
A room will have a new attribute `is_starting` defining wither is it **starting** or **not starting**.
Upon the creating of a room session, the room will be marked as **starting** and after the creation completes (successfully or by an error) it will be marked as **not starting** after 5 seconds by a job (to wait optimistically for the room to run on the BBB server in case it started successfully) [click here](https://github.com/bigbluebutton/greenlight/pull/3031/files#diff-16e9830311185de1ace2bb01b491a8e704f74aaebeeab2a0d21a72cb090f261eR185), [here](https://github.com/bigbluebutton/greenlight/pull/3031/files#diff-5fd8ae9ffb606caa8539e32c515131ef0123e741e5086e0367c2f9f2c1acc55aR23) and [here](https://github.com/bigbluebutton/greenlight/pull/3031/files#diff-849f1eb20de54d174f13d0d120e0defbda3accb44beef5bc4d5b4cce5aa41bf7R104).
Whenever a room that has been created is running or starting, it will conserve some of its settings upon any attempt to update them.
The `conserve_settings` method will take a list of settings and conserve them dynamically in such cases (meaning that changing that list can be done upon method call and it will dynamically iterate over them and do the rest) [click here](https://github.com/bigbluebutton/greenlight/pull/3031/files#diff-16e9830311185de1ace2bb01b491a8e704f74aaebeeab2a0d21a72cb090f261eR364).
This will prevent altering some critical settings when the room is running or starting on server side answering **Scenario 3**.
For UI consistency, any html element in the `create room modal` having the attribute  `data-disable-while-running` will be automatically disabled when the room is running or starting and it have the room state added to its settings that will be sent by the `room_settings` action upon fetching them on client side [click here](https://github.com/bigbluebutton/greenlight/pull/3031/files#diff-c4e7070b195c1385d4ca5eb639cb607adf62b2549358bc394b3cbdfb49081f7cR97), [here ](https://github.com/bigbluebutton/greenlight/pull/3031/files#diff-16e9830311185de1ace2bb01b491a8e704f74aaebeeab2a0d21a72cb090f261eR324)and [here](https://github.com/bigbluebutton/greenlight/pull/3031/files#diff-b209c5610e613906c971dbe77c47e32ff878d14c288f3fa99b6fa618cc691caaR293).
Without the **starting** state a room will still be vulnerable to **Scenario 3** because of the time required by BBB to mark the room as **running**.

For **Scenario 1 & 2** :
The recording consent is now a form input, meaning that its value will be passed to the controller that will check if the consent has to be accepted and make sure that it is otherwise it redirects users back to join page with the proper alert message forcing them to accept the consent when it has too [click here](https://github.com/bigbluebutton/greenlight/pull/3031/files#diff-c021d87ae09a8d12ce7dd52561f39f0e9f2da8a65a5cd36cf4633cf050c88245R50) and [here](https://github.com/bigbluebutton/greenlight/pull/3031/files#diff-a6077b61f51c536ef23f6d8aaeef34173f99c950df1b302a574bf9ee6761f1d0R65).
Accepting the consent will not be conserved upon entering the waiting channel since the controller action checked that the request is legit and rendered the waiting page which in its turn upon starting the room will simply make an `xhr request` to join the user leading to sending it without the consent being accepted.
For that the wait page will have a hidden html element that conserves the user choice for the consent and on client side it will make sure to extract that value and send it back to the controller upon joining [click here](https://github.com/bigbluebutton/greenlight/pull/3031/files#diff-2a92b2ea72001626f1ca601602d99e0cff62dec554e86620dd91411f74f868b0R55) and [here](https://github.com/bigbluebutton/greenlight/pull/3031/files#diff-d7af96554e17bd3310cef9018b48beca66c0aee86a85378328aac157863f889fR26).


farhatahmad:
Can you leave an explanation to what your solution to the problem actually is?

KH-Amir-TN:
> Can you leave an explanation to what your solution to the problem actually is?

Sure,
Check the description.

PR #3055, opened 2022-01-04
Fix cascaded deletion of shared_access for users and rooms

### LINKED ISSUE #2444: Deleting of shared room leads to SQL error
Greenlight v2.7.15.1

Reproduce:
1. Login with LDAP account.
2. Create a room.
3. Share it with another LDAP user.
4. Try to delete the room.

Result is a red error message that users don't understand:
`Room could not be deleted (PG::ForeignKeyViolation: ERROR: update or delete on table "rooms" violates foreign key constraint "fk_rails_e557734f55" on table "shared_accesses" DETAIL: Key (id)=(1137) is still referenced from table "shared_accesses". : DELETE FROM "rooms" WHERE "rooms"."id" = $1).`

### LINKED ISSUE #3054: User with access to shared rooms can not delete their own account
Greenlight v2.10.0.3

Reproduce:
1. Set up greenlight against PostgreSQL
1. Create two accounts
2. Create room with first account
3. Share above room with second account
4. Log in to second account
5. Delete second account via Profile -> Delete Account

Expected behaviour:
Second account is deleted

Actual behaviour:
Error 500 is displayed, and in the server log, this shows up:
```
Support: Error in user deletion: PG::ForeignKeyViolation: ERROR:  update or delete on table "users" violates foreign key constraint "fk_rails_9946f3f4cf" on table "shared_accesses"
DETAIL:  Key (id)=(5) is still referenced from table "shared_accesses".
: DELETE FROM "users" WHERE "users"."id" = $1
```

This seems to be related to #2444, I'm working on a fix for both.

zuntrax:
## Description

Fixes #2444
Fixes #3054 

Deletion of `user`s and `room`s fails with a foreign key constraint violation when either are referenced in the `shared_access` table. This is resolved by marking `shared_access` for deletion in case of dependencies in the `user` and `room` models.

This occurs only when running against PostgreSQL, as default configurations of SQLite don't seem to enforce foreign keys. As the testsuite usually is executed against SQLite, we made sure that our tests also fail when deletions succeed and orphaned `shared_access` entities are left over.

This PR might be a candidate for squashing, for now, we created four separate commits for tests and fixes of both bugs.

## Testing Steps

* Checkout this branch without the last two commits, verify that two tests fail
* Checkout tip of this branch, verify that tests pass


danimo:
@farhatahmad ping? This is a serious issue for everyone who uses Postgres (i.e. not SQLite). Could you merge this and consider a maintenance release?

app/models/user.rb
```diff
@@ -28,7 +28,7 @@ class User < ApplicationRecord
   before_destroy :destroy_rooms
 
   has_many :rooms
-  has_many :shared_access
+  has_many :shared_access, dependent: :delete_all
```
farhatahmad, line 31: This should be `dependent: :destory` instead

farhatahmad:
Thanks for the quick fixes @zuntrax - not sure how I missed this the first time around

anmol1991:
How to run spec test cases ?

PR #3095, opened 2022-01-24
Require at least 2 characters for a room name

worldthumbler:
## Description

Changed form for room creation to require 2 or mare characters as the room name
Will close #3094 
## Screenshots (if appropriate):

image

PR #3096, opened 2022-01-25
GIT-3086: Fixed the sessions persistent active state after a password update (Closes #3086).

### LINKED ISSUE #3086: Resetting | Changing password active session
Any user having their password reset/changed successfully expect all of their old open sessions to get closed.
At the moment resetting/changing a password won't close any already opened sessions which means that in events where users successfully attempts to change and/or reset their password any other opened sessions on their account will remain active (with full control over the account).
We can give users the choice to close all of their opened sessions or to keep them alive, or we can make this process mandatory on password update.

KH-Amir-TN:
## Description

As described in #3086 any password update won't invalidate any open sessions on that same account.
This raises some security concerns in scenarios of which someone gains access to an open session with full control over a user account which may be closed if the legit user attempts to change their password wither by a password reset or through their edit profile section.
This PR solves the issue by introducing a way for GL to check the validity of opened sessions and update their state accordingly.
Since GL is a ruby on rails application if fallows its guidelines for managing sessions.
In rails there's two defacto standards to work with sessions: one that stores sessions data on server i.e. in a  database and it's called `ActiveRecord::SessionStore` and an other that mitigates the storing of the session to the client browser in an encrypted signed cookie `ActionDispatch::Session::CookieStore`.
Because of the low amount of data stored in sessions ,and for performance and simple management the GL application uses the cookie based store.
The solution is backward compatible with already used session storing mechanism where it only adds a new metadata on each session that the GL application will use on later requests to check for their validity.

## Solution:
The GL session data contains the authenticated `user_id` along with some other values used by the rails framework.
Now each session will have a new attribute the `activated_at` which will hold the timestamp in which this session was activated.

#### On any request:
The `current_user` method now will be in charge of checking the validation of the session by extracting its `activated_at` metadata and checking if:
1. The session was activated before the account latest password update:
->This means that session is invalid and therefore the method will reset it and kills the request-response chain by redirecting
back the requester to `root_path` with the proper flash alert message [1].
2. The session has no `activated_at` metadata meaning that it's an old active session before the upgrade took place:
-> Here for backward compatibility and for seamless integration the `current_user` will check if the account in question had a password update by comparing it to `last_pwd_update` timestamp :
   2.1 If it had then this must be an old active request that hasn't made any interaction after the upgrade and another session
     updated its state and changed the account password invalidating the session in question and the method here will do the same as [1].
   2.2 If it hadn't then this must be an old active request that hasn't made any interaction after the upgrade but still valid and the `current_user` method will default it to the timestamp of the request and further requests will be checked against 1).

Note: the `last_pwd_update` will default to `nil` and `nil.to_i` will be 0 making comparing any none  `nil` `activated_at` timestamp valid.

#### On login:
When the users authenticates them selves successfully just like their session gets updated by adding and storing their id now it will hold also the timestamp of their login which will be the same as the `last_login` attribute in their record in the database.

#### On password change:
When a connected user changes their password successfully just like updating their `secure_password` attribute they also will have a new one the `last_pwd_update` that will hold the date and time of the successful password update.
This will invalidate any upcoming requests with sessions on the same account and to keep alive the one that made the password change it will have its `activated_at` metadata updated to match the `last_pwd_update` timestamp. 

#### On password reset:
1. When an unauthenticated request with a valid reset password URL attempts to change the account password the `last_pwd_update` will be updated to the event date and time and the request `activated_at` won't be updated resulting in invalidating all upcoming requests with sessions on that account.
2. When an authenticated request on the account which the reset password will take place the `last_pwd_update` will be updated to the event date and time and the request `activated_at` will match its timestamp keeping only that session alive while terminating all others on their next interaction with the server.
2. When an authenticated request on a different account (which is a bizarre scenario) than the one which the reset password will take place on, the `last_pwd_update` for the target account will be updated to the event date and time and the request `activated_at` won't be updated (This prevents using any valid reset password URL to circumvent the session checks).

## Testing Steps

1. Build the docker image with this PR solution.
2. Run the `bigbluebutton/greenlight:v2` docker image on an empty database.
3. Created two accounts.
4. Created four different browser sessions in which three of was connected on the first account and one on a the other account.
5. Killed the running containers mimicking an upgrade on a running deployment.
6. Run the PR docker image on the existing database.
7. Reloading two browser sessions on the 1st account and confirming that it remains authenticated with an update to the session `activated_at`.
8. Changing the first account password on the first browser session and confirming that it remained alive and that the other browser session got terminated with the request redirected to the `root_path` with the right flash alert.
9. Reloading the third browser session and confirming that it gets caught as invalid despite its `nil` `activated_at` with a redirection to the `root_path` with the right flash alert.
10. Reloading the last browser session and confirming that the session still alive and its `activated_at` gets created with the request timestamp.
11. Signin two out of the three browser sessions on the first account and confirming the update of their `activated_at` attribute matching their login timestamp. 
12. Attempting to reset the password on the unauthenticated session.
13. Reloading all browser sessions and confirming the invalidation of the two browser sessions on the first account and the conserve of the `activated_at` timestamp for the other two (the unauthenticated and the one with the second account id).
14. Signing all of the three browser sessions on the first account making all of the four sessions authenticated with three of which on the first account and one on the second.
15. Attempting to reset the password for account one on one of the authenticated sessions on the same account.
16. Confirming that all of the other two sessions on the same account gets invalidated with the proper outcome (redirection and flash alter etc...) while updating the requester session `activarted_at` timestamp to match the new `last_pwd_update` and conserving the forth session data.
17. Signing all of the two browser sessions on the first account and attempting to reset the password on the forth session (which is connected on the second account).
18. Confirming that all of the three browser sessions gets invalidated while conserving the last one but without an update to its `activated_at` nor its `last_pwd_update`.


KH-Amir-TN:
The `Rubocop + RSpec` appears to be blocked for a reason.
I tried to force-push and even to create a new commit and push it.
But that doesn't seem to solve it.
Also happened with #3095.

farhatahmad (approved):

JamieSlome:
@farhatahmad - I believe that this addresses the issue initially discovered in the [report](https://www.huntr.dev/bounties/9b341840-fd3f-4a21-839f-ad1fcb422a0e/).

Can you confirm the above, and if possible also confirm the fix commit SHA against the report, and we will proceed to attach the CVE you created to the report (`CVE-2022-36029`)?

farhatahmad:
@JamieSlome That's correct - I'll update the 2 reports fixed now

SamuelWei:
How is it handled if the user is authenticated via omniauth, e.g. login via LDAP and the password is changed in LDAP?

Issue #3100, opened 2022-01-26
Password complexity should be reconsidered 


KH-Amir-TN:
Using the `secure_password` flag to distinguish users with "insecure" passwords  who in turn has to reset their passwords on the next login.

This raised some issues:
1.  Because of the solution design the flag has to get updated manually in several areas like with #3078 and #3079.
2.  Because of the default `false` value of the flag all users will have to change their password upon the next login despite its complexity which caused a bad experience on their side.

Since the flag doesn't appear to provide good use and it's error prone a better alternative would be to separate concerns by validating the password complexity with  local account `creations` and `updates` and upon authentication the password complexity will be tested:
1. If the password is secure then the upgrade will be seamless for those users.
2. If the password isn't secure then users won't login until they update their password to match the required complexity (users will get redirected instantly to the reset password page with the proper flash alert message just like with the original solution.).

With this issue getting closed and with #3082 getting merged, that should enhance the overall UX when dealing with anything that relates to passwords in Greenlight (GL).

Issue #3091, opened 2022-01-21
the new password rules do not match the actual pattern


rottaran:
The validation pattern is defined here:
https://github.com/bigbluebutton/greenlight/blob/964d74689d31c98881e8fd59d9124e2c72a19d0e/app/models/user.rb#L48

The password change formular says:
* At least 8 characters
* At least 1 lowercase character
* At least 1 uppercase character
* At least 1 number
* At least 1 symbol

https://github.com/bigbluebutton/greenlight/blob/964d74689d31c98881e8fd59d9124e2c72a19d0e/app/views/users/components/_password_security.html.erb#L27

Some users have tried to use `,` and `.` as symbol. This is not in the pattern `#?!@$%^&*-`. There are likely other symbols that users will come up with.

It might be wise to explicitly state the available symbols on the user interface or accept anything other than digits and a-zA-Z as symbol.

falkiy:
Hi, struggled and wondered today creating new users.

`root@server:~# docker exec greenlight-v2 bundle exec rake user:create["test","test3@test.de","$(pwgen 8 -scy)","admin"]
Invalid Arguments
{:password=>["is invalid"]}`

`root@server:~# docker exec greenlight-v2 bundle exec rake user:create["test","test3@test.de","$(pwgen 8 -scy)","admin"]
Invalid Arguments
{:password=>["is too short", "is invalid"]}`

`root@server:~# docker exec greenlight-v2 bundle exec rake user:create["test","test3@test.de","$(pwgen 8 -scy)","admin"]
Invalid Arguments
{:password=>["is invalid"]}`

`root@server:~# docker exec greenlight-v2 bundle exec rake user:create["test","test3@test.de","$(pwgen 8 -scy)","admin"]
Account successfully created.
Email: test3@test.de
Password: M6~Yq~O?
Role: admin`

Seems not very stable. No problem for me deploying manual and finding this issue, but maybe real issue in scripted environments (ansible and so on)?

thx && greets

lonesomewalker:
Yeah, you are absolutely right, it was a nightmare :-/

Hamid-Najafi:
same problem :(
Invalid Arguments
{:password=>["is invalid"]}

KH-Amir-TN:
Hi **everyone**,

After reading carefully all of your comments I want to ensure you that all of the issues that was caused by the latest release are under our radar and has the highest priority.
We do apologize if you had any bad experience of any sort using Greenlight. 
[...]

PR #3082, opened 2022-01-18
GIT-1861: Fixed reset password and activation dead lock on users with terms unaccepted

KH-Amir-TN:
## Description

Organizations using Greenlight (GL) may have certain policies, terms and conditions that they would require their users to accept in order to consume their platform services.
Users who don't accept these terms won't be able to access all of the platform features until they align with its values.
But a user with unaccepted terms doesn't necessary mean that they refuse them as mentioned in #1861 [Terms can change and users may be explicitly have their acceptance reset by administrators in order to force them to read the new changes and accept them.].
Users with local accounts who still haven't accepted the terms should remain able to:
1. Reset their passwords.
2. Verify their email address.
3. Edit their profile information and password.
But users won't be able to join or manage their rooms or rooms that were shared with them until they accept terms.
There were issues that prevented users with unaccepted terms from having capabilities mentioned above.
This pull request ensures the users ability to reset their password, verify their email and change their information and administrators ability to send reset password emails and update profiles for users with unaccepted terms.
This pull request also solves some other minor bugs in the upstream not directly related to #3081.

#### Password reset [FIXED][#1861]
1. User hadn't accept the terms.
2. User had forgotten their password and tries to reset it.
3. User receive the reset email.
4. User can't reset their password to login to accept the terms [dead lock].

### Account activation (email verification) [FIXED][#1861]
1. User have unaccepted terms (i.e. User signs up while no terms exist).
2. User receives the verification email.
3. Before verifying the email an administrator enables terms and conditions.
4. User can't activate their account [dead lock].

### User edit profile [FIXED][considered in same context with #1861]
1. User login with unaccepted terms.
2. User doesn't accept terms.
3. User access their edit profile space and attempt to update their information.
4. User can't update their information nor change their password.

Also:
### New email verification [FIXED][#3080]
1. A logged in active user or an administrator access the user edit profile panel.
2. An update request to the user email gets created.
3. User have their email updated without having to verify it.

### Secure password circumventing [FIXED][#3078]
1. User with old password logged in or an administrator access the user edit profile.
2. An administrative update upgrades the platform to add new password complexity policy.
3. User will have their `secure_password` set to `false` by default to force them to reset it upon the next login.
4. User or the administrator make an update on the user's information.
5. User gets their `secure_password` set to `true` and therefore they circumvent the password forced reset.

### Change password to a secure one [FIXED][#3079]
1. A logged in user with an insecure password access their edit profile space.
2. User attempts to change their password to a secure one without making a reset.
[...]

Issue #3081, opened 2022-01-18
Password complexity requirements indication and password errors on the view lacks proper localization


KH-Amir-TN:
With the new password complexity increase the password validation encapsulate a lot of constraints.
Therefore, writing these criteria back to the user was a great action and having explicit and wide errors was mandatory.
The problem with these indications is that they weren't localized and they were only statically written in English. 
This it self isn't an issue but an enhancement request but in the same context the change password area in the edit profile panel errors aren't correct and adequate with the scenario.
Having the new password with `:password` as a key will cause these inconsistencies with the error observation and the error itself.
I.e. giving a wrong Old password will indicate that the Old password is incorrect and that the new password doesn't meet complexity requirements despite its value.
[CHECK HERE](https://github.com/bigbluebutton/greenlight/blob/master/app/views/users/components/_password_security.html.erb), [HERE](https://github.com/bigbluebutton/greenlight/blob/master/app/controllers/users_controller.rb#L135) and [HERE](https://github.com/bigbluebutton/greenlight/blob/master/app/controllers/users_controller.rb#L131).

PR #3101, opened 2022-01-26
GIT-3100: Reconsidered the password complexity check design.

KH-Amir-TN:
## Description

As of the latest release of Greenlight (GL) `v2.11.x` the password complexity has increased.
The solution design used a flag (`secure_password`) to mark accounts with "secure/insecure" passwords so that the application can decide the proper action to take with each account.
Any user local account with insecure password (`secure_password == false`) won't login and gets redirected directly to `reset_password` page to update it matching the new requirements.
This raised some issues as described in #3078,#3079,#3081, #3091,#3100.
This addresses #3100 and #3091 while PR #3082 targets the others.

## Solution:
This proposition is straightforward, any local account will have its password validated against the complexity requirements upon login without the use of the `secure_password` flag which simplifies things a lot and avoid potential errors and overhead:
1. If the password is secure then the user login.
2. If the password is insecure (doesn't match the `PASSWORD_PATTERN`) the user won't login and they will get redirected to reset password page.
This will ensure that old accounts with already "secure" passwords doesn't get prompted to change theirs.

The password minimal complexity required remains the same:
1. A password has to have at least **One digit**.
2. A password has to have at least **One non-word characters (anything but `a-z, A-Z, 0-9`)**.
3. A password has to have at least **One Upper case letter**.
4. A password has to have at least **One lower case letter**.
5. A password has to be at least **8 characters in length**.

## Testing Steps

1. Ran `bigbluebutton/greenlight:v2.10.0.3` on empty database.
2. Created three accounts two of which with insecure and one with a secure password.
3. Opened three browser sessions for each account.
4. Ran this PR docker image on the existing database.
5. Attempted to change the password on one account with an insecure one:
   5.1 Confirmed the complexity validation for the new password.
   5.2 Changed the password to a secure one.
   5.3 Signed out and resigned in on same account and confirming that the user doesn't has to reset their password.
6. Signed out all of the other sessions and resigned in to confirm that:
   6.1 The account with insecure password can't login and gets redirected to the reset password page:
       6.1.1 All validation for password complexity works.
       6.1.2 After resetting the password and signing in the user logs in without having to reset it again.
       6.1.3 Signing out and resigning in without having to reset the password.
   6.2 The user with secure password logs in without having to make a reset.

Also, Tested for `user` and `admin` create rake tasks.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -21,6 +21,7 @@ require 'bbb_api'
 class User < ApplicationRecord
   include Deleteable
 
+  PASSWORD_PATTERN = /\A(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*(\W|_)).{8,}\z/
   after_create :setup_user
 
   before_save { email.try(:downcase!) }
@@ -45,7 +46,7 @@ class User < ApplicationRecord
                     format: { with: /\A[\w+\-'.]+@[a-z\d\-.]+\.[a-z]+\z/i }
 
   validates :password, length: { minimum: 8 },
-            format: /\A(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).{8,}\z/,
+            format: PASSWORD_PATTERN,
             confirmation: true,
             if: :validate_password?
 
@@ -113,6 +114,10 @@ class User < ApplicationRecord
 
       search_param = "%#{sanitize_sql_like(string)}%"
       where(search_query, search: search_param)
+    end
+
+    def secure_password?(pwd)
+      pwd.match? PASSWORD_PATTERN
     end
   end
 
--- a/app/controllers/password_resets_controller.rb
+++ b/app/controllers/password_resets_controller.rb
@@ -55,8 +55,8 @@ class PasswordResetsController < ApplicationController
     elsif @user.update_attributes(user_params)
       # Clear the user's social uid if they are switching from a social to a local account
       @user.update_attribute(:social_uid, nil) if @user.social_uid.present?
-      # Mark password as secure and deactivate the reset digest in use.
-      @user.update(reset_digest: nil, reset_sent_at: nil, secure_password: true)
+      # Deactivate the reset digest in use disabling the reset link.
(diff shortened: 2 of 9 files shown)


db/migrate/20211128175028_add_secure_password_to_users.rb
```diff
@@ -1,7 +0,0 @@
-# frozen_string_literal: true
-
-class AddSecurePasswordToUsers < ActiveRecord::Migration[5.2]
-  def change
-    add_column :users, :secure_password, :boolean, null: false, default: false
-  end
-end
```
farhatahmad, line 7: Deleting a migration will cause other issues. You should instead create a new migration that removes the column instead

app/models/user.rb
```diff
@@ -114,6 +115,10 @@ def merge_list_search(string)
       search_param = "%#{sanitize_sql_like(string)}%"
       where(search_query, search: search_param)
     end
+
+    def secure_password?(pwd)
+      PASSWORD_PATTERN === pwd
```
farhatahmad, line 120: Hmmm I've never seen `===` used to check regex. I normally would use
`pwd.match(PASSWORD_PATTERN)`

farhatahmad:
Other than the 2 issues above, this looks good. I'll merge this and look at a new release once the issues are fixed

db/migrate/20211128175028_add_secure_password_to_users.rb
KH-Amir-TN: You got reason it's wrong.
I'll change it.

app/models/user.rb
KH-Amir-TN: It's a case equality that `case` uses under the hood.
https://ruby-doc.org/core-3.0.2/Regexp.html#method-i-3D-3D-3D
Do you want me to use `#string.match`?

KH-Amir-TN: DONE

db/migrate/20211128175028_add_secure_password_to_users.rb
KH-Amir-TN: DONE

farhatahmad (approved):

PR #3176, opened 2022-03-02
Remove provider from whitelist

scouillard:
## Description

On User update, provider attribute cannot be changed.
User could bypass read-only form input and change the provider via browser.
Provider attribute is removed from whitelist in User model.

## Testing Steps

- Log into your Greenlight account and go to your user profile.
- Via the browser HTML file, remove the read-only setting and change your provider.
- Try to update the user.


farhatahmad:
Not quite the approach I was looking for. Hint: Look in the `users_controller` and follow the logic in the code

farhatahmad:
Make sure you run `rubocop -A` before pushing. See the errors here: https://github.com/bigbluebutton/greenlight/runs/5409011431?check_suite_focus=true

farhatahmad (approved):

hadicheaito1 (approved):

PR #3279, opened 2022-03-10
Preventing XSS Share Room Access

hadicheaito1:
## Description

- Changed rendering a string of html to rendering html using Javascript to properly prevent XSS in Shared Rooms modal
## Testing Steps

- Create User 1 and User 2
- Share a room owned by User 1 to User 2
- Let User 2 then change their profile name to: Rein 
- Go to share room modal and see that there is no XSS occurring anymore

As you can see, there is no more alert occurring:
image

### DIFF
--- a/app/assets/javascripts/room.js
+++ b/app/assets/javascripts/room.js
@@ -350,12 +350,6 @@ function displaySharedUsers(path) {
       spanRemove = document.createElement("span"),
       spanRemoveIcon = document.createElement("i");
 
-      listName.appendChild(spanAvatar)
-      listName.appendChild(spanName)
-      listName.appendChild(spanUid)
-      listName.appendChild(spanRemove)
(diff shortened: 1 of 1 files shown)


scouillard (approved):

app/assets/javascripts/room.js
```diff
@@ -338,18 +338,38 @@ function saveAccessChanges() {
...
+      listName = document.createElement("li"),
+      spanAvatar = document.createElement("span"),
+      spanName = document.createElement("span"),
+      spanUid = document.createElement("span"),
+      spanRemove = document.createElement("span"),
+      spanRemoveIcon = document.createElement("i");
+
+      listName.appendChild(spanAvatar)
+      listName.appendChild(spanName)
+      listName.appendChild(spanUid)
```
farhatahmad, line 355: spanUid should be a child of spanName

PR #3287, opened 2022-03-14
Add missing React librairies

scouillard:
Add React-Query for fetching/updating from/to RoR API.
Add React-Hook-Form for Forms.


farhatahmad:
You've got conflicts. You need to make sure you always fetch upstream first before branching off

PR #3300, opened 2022-03-16
GLv3-3202: Allow user to "Sign Up".

KH-Amir-TN:
Backend:
		Done:
                        - Set the API endpoint for the signup '/api/v1/users.json'.
                        - Made some changes to Rubocop.
                        - Added a request spec for the users API and prepared the tests for the signup with/without valid user data.
                        - Improved routing compatibility between SSR (Server side routing) and CSR (Client side routing).
                        - Tried defining an approach for error and request handling (Still require more work).
                        - Improved API resiliency by introducing a much stricter API:
                        -> API calls has to have an explicit `.json` format along with an `Accept: 'application/json'` header.
                        -> The TODOs were left as comments in the code.
	Frontend:
		        - Set a new project structure for the front end JavaScript code influenced by the
		        react-router-demo https://stackblitz.com/edit/github-agqlf5/ :
		            -> main.jsx: An entry point where Routes, Layouts, Global state is passed/defined.
		            -> app.jsx: The parent route component for the `/` endpoint.
		            -> components/: A directory where custom components resides.
		            -> components/styling: A directory where custom components with no business logic resides (i.e. custom buttons).
		            -> components/forms: A directory where custom form components resides.
		            -> routes: A directory where sub routes resides.
		            
		        - Created a Signup route that renders `routes/Signup.jsx` for the `/signup` endpoint.
		        - Created a `components/forms/SignupForm.jsx` form component that describes the signup form.
		        - Created a `helpers/forms/SignupFormHelpers.jsx` module to bundle the helpers for the signup form component.
		        - Created a `components/forms/FormControl.jsx` component that wraps/connects a react bootstrap form control component with the react hook form API.
		        - Created a `comonents/styling/buttons/ButtonLink.jsx` component that wraps/connects a react bootstrap button component with the react router API.
                        
                        TODO: 
                             - Use axios and react-query instead of the fetch API.
                             - Introduce a way to check for email uniqueness asynchronously from the client side.
                             - Improve the form performance.
                             - Improve API error handling. 
                             - Use typescript.
                             - Add any required tests.
                             - Configure the eslinter.

## Description

Allow users to signup in GLv3.

## Testing Steps

1. Pull the code.
3. Run any left migration `rails db:migrate`
4. Create any left database for testing `rails db:create`
5. Run the linter and tests `bundle exec rubocop --parallel && bundle exec rspec`
6. Run the rails s `rails s` or better `./bin/dev` to run the assets builders processes at once.

To test the API:
     
  To test the UI:
    1. Enter the index `/`
    2. You should have this UI: 
image

    3. For any other URL except for `"/" and "/signup"`, the UI should render this:    
image

4. Entering "/signup" or clicking the `Signup` button should render this:
image

5. Clicking the `Home` button should have the same outcome as accessing the index `/` page.
6. Fill the form, validations should run as expected.
7. Data won't get submitted only if the form is valid.
8. The result of the API call can be viewed with the dev tools along with the puma server logs and the users table in the db.


hadicheaito1 (approved):

app/models/user.rb
```diff
@@ -3,7 +3,10 @@
 class User < ApplicationRecord
   has_secure_password
 
-  validates :name, presence: true
+  validates :name, presence: true # TODO: Change into full_name or seperate first and last name.
+  # TODO: Validate email format.
   validates :email, presence: true, uniqueness: { case_sensitive: false, scope: :provider }
-  validates :provider, presence: true
+  validates :provider, presence: true # TODO: Set a default value or remove the presence validation.
```
farhatahmad, line 9: I think provider is something that will need to be explicitly set, so I don't think there's a need for a default value

spec/requests/users_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+        name: Faker::Name.name,
+        email: Faker::Internet.email,
+        password: 'Password123+',
+        password_confirmation: 'Password123+'
+      }
+    }
+  end
+
+  describe 'Signup' do
+    let(:headers) do
```
farhatahmad, line 18: This can be replaced with 1 line `let(:headers) { { 'ACCEPT' => 'application/json'}  }

config/routes.rb
```diff
@@ -4,4 +4,11 @@
   # Define your application routes per the DSL in https://guides.rubyonrails.org/routing.html
 
   root 'components#index'
+  # All the Api endpoints must be under /api/v1 and must have an extension .json.
+  namespace :api do
+    namespace :v1 do
+      post '/signup', to: 'users#create' # /api/v1/signup.json
```
farhatahmad, line 10: I think it may be better to keep the standard Rails routing concepts here. Use `resources :users, only: :create` -> POST /api/v1/users

spec/requests/users_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    context 'invalid user params' do
+      # Common expectations bundled for a DRY code.
+      def expectations(invalid_user_params)
+        expect { post '/api/v1/signup.json', params: invalid_user_params, headers: }.not_to change(User, :count)
+        expect(response.content_type).to eq('application/json; charset=utf-8')
+        expect(response).to have_http_status(:bad_request)
+        expect(JSON.parse(response.body)['errors']).not_to be_empty
+      end
+
+      it 'fails for missing params / password confirmation mismatches' do
```
farhatahmad, line 42: Can we split this into 2 tests? 1 for missing params and 1 for password_confirmation not matching

spec/requests/users_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+        invalid_user_params = { user: { password: 'something', password_confirmation: 'something_else' } }
+        expect { post '/api/v1/signup.json', params: invalid_user_params, headers: }.not_to change(User, :count)
+        expectations invalid_user_params
+      end
+
+      it 'fails for duplicated name for same provider' do
```
farhatahmad, line 48: This is already tested on the model, so no need to test it here

spec/requests/users_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    context 'invalid user params' do
+      # Common expectations bundled for a DRY code.
+      def expectations(invalid_user_params)
+        expect { post '/api/v1/signup.json', params: invalid_user_params, headers: }.not_to change(User, :count)
+        expect(response.content_type).to eq('application/json; charset=utf-8')
+        expect(response).to have_http_status(:bad_request)
+        expect(JSON.parse(response.body)['errors']).not_to be_empty
+      end
```
farhatahmad, line 40: While I like this idea, I think it's a bit overkill for now. Repeating expects is pretty standard in tests and only when it comes to 4 or 5 (or greater) tests should we move it to a separate method

app/models/user.rb
KH-Amir-TN: Alright, I'll leave the defaulting this way until we define the explicit user domain https://github.com/bigbluebutton/greenlight/pull/3300/files#diff-ef20d40ddd2dd2a254d26fd4c59655cbed6b541c5f908b81125d09848c1dfeb6R19
I left the comment there because it felt weird but it makes sense that we'll change it.

spec/requests/users_spec.rb
KH-Amir-TN: Rubocop asks to put let blocks with do;end rather then the brackets.
I'll try to inline it with do;end; 
Is this cop necessary?

config/routes.rb
KH-Amir-TN: Alright,

spec/requests/users_spec.rb
KH-Amir-TN: Thought about duplication since the validation source is the same for both scenarios (ActiveRecord).
I'll split them into two.

spec/requests/users_spec.rb
KH-Amir-TN: Makes sense, I'll remove it.

spec/requests/users_spec.rb
KH-Amir-TN: Alright, I'll embrace duplication for now and I'll leave this for a refactoring.

spec/requests/users_spec.rb
farhatahmad: Lets disable that cop then

KH-Amir-TN: I'll be more than happy to do it HH

spec/requests/users_spec.rb
KH-Amir-TN: Done

spec/requests/users_spec.rb
KH-Amir-TN: Done

app/models/user.rb
KH-Amir-TN: Done,

config/routes.rb
KH-Amir-TN: Done

spec/rails_helper.rb
```diff
@@ -66,3 +66,12 @@
...
 end
+
+# Tests constants
+# API endpoints
+API_URL_PREFIX = '/api/v1'
+API_ENDPOINTS = {
+  Users: {
+    Signup: "#{API_URL_PREFIX}/users.json" # POST /.../users.json
+  }
+}.freeze
```
farhatahmad, line 77: Not sure this file is the right place for this

farhatahmad: Can we not access the standard rails routes in tests?

spec/requests/users_spec.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Users API', type: :request do
```
farhatahmad, line 5: Shouldn't this be `RSpec.describe Api::V1::UsersController` or something like that?

spec/requests/users_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+    end
+
+    context 'invalid user params' do
+      it 'fails for missing params' do
+        invalid_user_params = { user: {} }
+        expect { post API_ENDPOINTS[:Users][:Signup], params: invalid_user_params, headers: }.not_to change(User, :count)
+        expect(response.content_type).to eq('application/json; charset=utf-8')
+        expect(response).to have_http_status(:bad_request)
+        expect(JSON.parse(response.body)['errors']).not_to be_empty
```
farhatahmad, line 35: I would actually check that a specific error is returned. Maybe omit 1 param and check the `errors` array

spec/requests/users_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+
+      it 'fails for password confirmation mismatches' do
+        invalid_user_params = valid_user_params
+        invalid_user_params[:user][:password_confirmation] = 'something_else'
+        expect { post API_ENDPOINTS[:Users][:Signup], params: invalid_user_params, headers: }.not_to change(User, :count)
+        expect(response.content_type).to eq('application/json; charset=utf-8')
+        expect(response).to have_http_status(:bad_request)
+        expect(JSON.parse(response.body)['errors']).not_to be_empty
```
farhatahmad, line 44: Same as above

farhatahmad (approved):

spec/rails_helper.rb
KH-Amir-TN: You mean we should use xxx_path route helpers instead?
I thought about it but it felt like a problem because it violates the testing in isolation concept by depending on the route helpers which may not help us spot bugs.
I'm not sure if its the right place to do that there either, I haven't gave it a lot of thought but surely it felt right to have a single source of truth for endpoints to avoid duplication and errors on the run and it should be independent from anything (i.e. the actual routes) to spot issues.

spec/requests/users_spec.rb
KH-Amir-TN: I used Rspec request api not the controller Api which is the new recommend way to make integration tests.
Symmetrically request api isn't limited to controllers but rather to a bigger context like Users Api or Rooms Api where tests for the full request/response cycle lives.
Do you want me to use the subject notation?

spec/requests/users_spec.rb
KH-Amir-TN: You got reason.
What about deep matching the user erro4s array with the errors array recieved in the response body?
I think it solves this concern?

spec/rails_helper.rb
KH-Amir-TN: Do you want me to remove it?

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      #     - Filters and validates the recieved params( throws ActionController::ParameterMissing for missed params).
+      #     - Creates and saves a new user record in Users table with the provided parameters.
+      #     - Returns json with the fallowig format : { msg: String, data: Array[serializable objects]
+      #      ,errors: Array[String] }
+      #     and status_code: 201 (created) | 400 (bad request).
+      #         msg = response_msgs[:success] and errors = [] for successful save to db with status_code = 201.
+      #         msg = response_msgs[:failed] and errors = user.errors.to_a for unsuccessfuly save to db
+      #         with status_code = 400.
+      def create
+        user = User.new({ provider: 'greenlight' }.merge(user_params)) # TMP fix for presence validation of :provider
```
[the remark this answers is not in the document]
hadicheaito1: Never mind, ignore this. I didn't notice the uniqueness constraint on the email user model :sweat_smile:

spec/rails_helper.rb
farhatahmad: In this case, I think it makes sense to just use the routes since they already exist. I don't think it necessary violates the isolation concept, it just uses the predefined routes to know where to make the request to

spec/requests/users_spec.rb
farhatahmad: I'd still explicitly check the error is what we expect it. What if the error returned here is "password confirmation doesn't match" or "email is already taken" or any other error?

Our test is specifically for missing params so we should explicitly check for that. Otherwise, this is just a test that check that those params return an error (doesn't matter which one)

app/controllers/api/v1/users_controller.rb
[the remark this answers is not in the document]
KH-Amir-TN: No worries, it happens a lot HH

spec/rails_helper.rb
KH-Amir-TN: Alright, I'll use the route helpers instead.

spec/requests/users_spec.rb
farhatahmad: Yeah I'd keep the `type: :request` but for the sake of code cleanliness, I'd rather we separate tests by controller, just to keep things easy to manage

spec/requests/users_spec.rb
KH-Amir-TN: But after giving it a second thought it feels that the word "missing" isn't logical because the checks are done by the model preset validations and has nothing to do with some parameters missing or not.

I think I should change test description to invalid parameters instead of missing because that's what it does.
And I'll make sure the errors in the response matches those of the model validation with the same values.
What do you think?

spec/requests/users_spec.rb
KH-Amir-TN: Alright, Ill change it.

spec/requests/users_spec.rb
farhatahmad: Yeah that's another option. We simply test that invalid params = fail and not get into the specifics of why or how. 

I think that makes more sense as well

KH-Amir-TN: Perfect, I'll change it either.
Thanks for your review!

spec/rails_helper.rb
KH-Amir-TN: Done

spec/requests/users_spec.rb
KH-Amir-TN: Done

spec/requests/users_spec.rb
KH-Amir-TN: Done

app/models/user.rb
```diff
@@ -2,10 +2,11 @@
 
 class User < ApplicationRecord
   has_secure_password
-
   has_many :rooms, dependent: :destroy
-
-  validates :name, presence: true
+  validates :name, presence: true # TODO: amir - Change into full_name or seperate first and last name.
```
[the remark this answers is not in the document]
KH-Amir-TN: In terms of database normalization some may consider this as not normalized because it violates the 1st normal form:
`If a relation contains a composite or multi-valued attribute, it violates the first normal form, or the relation is in first normal form if it does not contain any composite or multi-valued attribute. A relation is in first normal form if every attribute in that relation is singled valued attribute. `
https://www.geeksforgeeks.org/first-normal-form-1nf/

1. The semantic name should be descriptive -> `full_name` rather than `name`.
2. Having two attributes one for the first name and the other for the last name may simplify validations and filtering (i.e. won't have to worry about whitespaces and its implications for length and representation).
3. For database administrators to better filter data and query it along with a better use of memory (see database normalization).
But this is not a big deal in our case at least it should be name `full_name` IMHO.

app/models/user.rb
```diff
@@ -2,10 +2,11 @@
...
+  # TODO: amir - Validate email format.
   validates :email, presence: true, uniqueness: { case_sensitive: false, scope: :provider }
   validates :provider, presence: true
+  # ADDED: presence validation to ensure password confirmation to match the password value.
```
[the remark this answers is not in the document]
KH-Amir-TN: It should be removed.

spec/requests/users_controller_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+    context 'invalid user params' do
+      it 'fails for invalid values' do
+        invalid_user_params = {
+          user: { name: '', email: 'invalid', password: 'something', password_confirmation: 'something' }
+        }
+        user = User.new invalid_user_params[:user]
+        user.provider = 'greenlight'
+        expect(user).not_to be_valid
```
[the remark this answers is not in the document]
KH-Amir-TN: Makes sense I'll remove it.

config/routes.rb
```diff
@@ -4,4 +4,11 @@
   # Define your application routes per the DSL in https://guides.rubyonrails.org/routing.html
 
   root 'components#index'
+  # All the Api endpoints must be under /api/v1 and must have an extension .json.
+  namespace :api do
+    namespace :v1 do
+      resources :users, only: [:create] # POST /api/v1/users.json -> api/v1/users_controller#create
```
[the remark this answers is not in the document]
KH-Amir-TN: Alright,

spec/requests/users_controller_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        expect { post api_v1_users_path, params: invalid_user_params, headers: }.not_to change(User, :count)
+        expect(response.content_type).to eq('application/json; charset=utf-8')
+        expect(response).to have_http_status(:bad_request)
+        expect(JSON.parse(response.body)['errors']).to eq user.errors.to_a
```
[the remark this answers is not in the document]
KH-Amir-TN: Chage back to `expect(JSON.parse(response.body)['errors']).not_to be_empty` ?

app/models/user.rb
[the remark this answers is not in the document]
KH-Amir-TN: So do I leave the todo comment there?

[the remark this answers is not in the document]
farhatahmad: Yeah

spec/requests/users_controller_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      }
+    }
+  end
+
+  describe 'Signup' do
+    let(:headers) { { 'ACCEPT' => 'application/json' } }
+
+    context 'valid user params' do
+      it 'creates a user account for valid params' do
+        expect { post api_v1_users_path, params: valid_user_params, headers: }.to change(User, :count).from(0).to(1)
```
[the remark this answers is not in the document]
farhatahmad: Oh nice I didn't know that. If that's the case then just leave it as is

spec/requests/users_controller_spec.rb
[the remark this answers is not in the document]
farhatahmad: Yeah

config/routes.rb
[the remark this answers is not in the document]
KH-Amir-TN: Done,

spec/requests/users_controller_spec.rb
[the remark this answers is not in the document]
KH-Amir-TN: Done

spec/requests/users_controller_spec.rb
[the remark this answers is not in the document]
KH-Amir-TN: Done,

app/models/user.rb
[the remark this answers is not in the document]
KH-Amir-TN: Done,

app/models/user.rb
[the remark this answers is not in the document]
KH-Amir-TN: Done,

spec/requests/users_controller_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      it 'fails for password confirmation mismatches' do
+        invalid_user_params = valid_user_params
+        invalid_user_params[:user][:password_confirmation] = 'something_else'
+        user = User.new invalid_user_params[:user]
+        user.provider = 'greenlight'
+        expect(user).not_to be_valid
+        expect { post api_v1_users_path, params: invalid_user_params, headers: }.not_to change(User, :count)
+        expect(response.content_type).to eq('application/json; charset=utf-8')
+        expect(response).to have_http_status(:bad_request)
+        expect(JSON.parse(response.body)['errors']).to eq user.errors.to_a
```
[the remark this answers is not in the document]
KH-Amir-TN: Would it better to remove it entirely since we agreed that each check is done by the model itself?
I think its better to keep only two examples in the integration (controller) spec where one tests the success of the full signup operation with valid data and the other tests its fail for invalid data without specifying what invalid means here.
And add a new example in the model spec where we test the password confirmation process?

KH-Amir-TN:
Closed in favor of decoupling.

PR #3301, opened 2022-03-16
Auto create a room when a user is created

### LINKED ISSUE #3175: Allow user to share room using URL
- This allows the user to copy room url and share it with others to make it easy to join

- Front End
- Back end
  - Generate invite path for room and check if it exists when joining

farhatahmad:
## Testing Steps

1. Create a new user `user = User.create(name: "Test test", email: "testa@as.com", provider: "greenlight", password: "test", password_confirmation: "test")`
2. Ensure the user has a room `user.rooms`

### DIFF
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -9,8 +9,6 @@ class Room < ApplicationRecord
 
   before_validation :set_friendly_id, :set_meeting_id, on: :create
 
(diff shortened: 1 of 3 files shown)


farhatahmad:
uhhh on second thought we probably don't want to do this

PR #3304, opened 2022-03-18
Add Sign In Form

scouillard:
## Description

Frontend:
User enter Email and Password in Form.
Form have basic validation.
On submit, Email and Password information is sent to API via fetch POST method.

Backend:
Server checks if User exists.
Server checks if User and Password match via authenticate method.
If success, return success status, signed_in as true and store User id in session.
If error, return error status and signed_in as false.

## Testing Steps

Set
sessionUrl =  'https://jsonplaceholder.typicode.com/posts'

Convert Fetch response to JSON
.then((response) => response.json())

Add 
.then((json) => console.log(json)) 

And you shall receive a response (user email, user password) with the form inputs.

image
Server receives the POST request with the email and password parameters


app/javascript/components/sessions/SignIn.tsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react';
+import { useForm } from 'react-hook-form';
+import Form from 'react-bootstrap/Form';
+import Button from 'react-bootstrap/Button';
+
+async function createSession(sessionUser) {
+  const sessionUrl = '/api/v1/sessions';
+
+  await fetch(sessionUrl, {
```
scouillard, line 9: status: "success" will be removed, is an artefact from testing

scouillard:
TODO: 
What causes the duplication in params (see screenshot) - why is session in Parameters?? 

-> Any custom config in Rails logger? Does it output params or params + session?

Else 

-> Debug w/ byebug
-> Whitelist maybe? session_params

app/javascript/components/sessions/SignIn.tsx
```diff
@@ -0,0 +1,74 @@
...
+      <h1> Sign In </h1>
+      <Form onSubmit={handleSubmit(handleSignIn)}>
+        <Form.Group className="mb-3" controlId="signInEmail">
+          <Form.Label>Email:</Form.Label>
+          <Form.Control
+            type="email"
+            placeholder="Email"
+            {...register('email', { required: 'Email is required.' })}
+          />
+          <Form.Text className="text-danger d-block">
```
KH-Amir-TN, line 55: That should do it.
Also, we can leverage the ` <Form.Control.Feedback type="invalid">` component.

app/javascript/components/sessions/SignIn.tsx
```diff
@@ -0,0 +1,74 @@
...
+          </Form.Text>
+        </Form.Group>
+        <Form.Group className="mb-3" controlId="signInPassword">
+          <Form.Label>Password:</Form.Label>
+          <Form.Control
+            type="password"
+            placeholder="Password"
+            {...register('password', { required: 'Password is required.' })}
+          />
+          <Form.Text className="text-danger d-block">
```
KH-Amir-TN, line 66: The same in here.

app/javascript/components/sessions/SignIn.tsx
```diff
@@ -0,0 +1,74 @@
...
+  };
+
+  const { register, handleSubmit, formState: { errors } } = useForm<FormInputs>();
+
+  const handleSignIn = (sessionUser) => createSession(sessionUser);
+
+  return (
+    <div className="col-md-3 mx-auto">
+      <h1> Sign In </h1>
+      <Form onSubmit={handleSubmit(handleSignIn)}>
```
KH-Amir-TN, line 47: I think it's better to have a single source of truth so we can disable the HTML5 validations here by using the `noValidate` prop and set it to `true`.
If you like to disable the browser auto completion we can also use `autoComplete="off"`.

app/javascript/components/sessions/SignIn.tsx
```diff
@@ -0,0 +1,74 @@
...
+        email: sessionUser.email,
+        password: sessionUser.password,
+      },
+    }),
+  })
+    .then((response) => {
+      if (response.status === 204) {
+        // TODO: samuel - Need to discuss how to Authenticate the user
+        // Do Something
+      } else {
```
KH-Amir-TN, line 24: Likes this!

app/javascript/components/sessions/SignIn.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { useForm } from 'react-hook-form';
+import Form from 'react-bootstrap/Form';
+import Button from 'react-bootstrap/Button';
+
+async function createSession(sessionUser) {
+  const sessionUrl = '/api/v1/sessions';
+
+  await fetch(sessionUrl, {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
```
KH-Amir-TN, line 11: Likes this!

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+module Api
+  module V1
+    class SessionsController < ApplicationController
+      before_action :authenticate
+
+      # TODO: samuel - Bypass CSRF token for now
+      protect_from_forgery with: :exception
+
+      def create
+        user = User.find_by(email: params[:user][:email])
```
KH-Amir-TN, line 12: We can leverage the Rails parameters [API](https://api.rubyonrails.org/classes/ActionController/Parameters.html).

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class SessionsController < ApplicationController
+      before_action :authenticate
```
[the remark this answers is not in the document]
scouillard: It's a Rails method.
https://apidock.com/rails/ActiveModel/SecurePassword/InstanceMethodsOnActivation/authenticate
I assume the error is due to the User controller not existing yet.

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+        if user.present? && user.authenticate(params[:user][:password])
+          session[:user_id] = user.id
+          render json: { status: :success, signed_in: true }, status: :no_content
+        else
+          render json: { status: :error, signed_in: false }, status: :bad_request
+        end
+      end
+
+      def destroy
+        session[:user_id] = nil
```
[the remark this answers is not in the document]
scouillard: We could use reset_session as well, although isn't it deprecated?
I think this is standard.

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+      protect_from_forgery with: :exception
+
+      def create
+        user = User.find_by(email: params[:user][:email])
+
+        if user.present? && user.authenticate(params[:user][:password])
+          session[:user_id] = user.id
+          render json: { status: :success, signed_in: true }, status: :no_content
+        else
+          render json: { status: :error, signed_in: false }, status: :bad_request
```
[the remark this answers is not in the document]
scouillard: I'm not sure what you mean.
This should only cover the Authentication error?

app/controllers/api/v1/sessions_controller.rb
scouillard: Agreed

app/javascript/components/sessions/SignIn.tsx
```diff
@@ -0,0 +1,74 @@
...
+    body: JSON.stringify({
+      status: 'success',
+      user: {
+        email: sessionUser.email,
+        password: sessionUser.password,
+      },
+    }),
+  })
+    .then((response) => {
```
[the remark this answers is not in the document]
scouillard: I need to read more on the async/await pattern, but no I don't think it is necessary indeed.

app/javascript/components/sessions/SignIn.tsx
scouillard: I don't understand why we should remove front end validations and auto completion

app/controllers/api/v1/sessions_controller.rb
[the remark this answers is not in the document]
KH-Amir-TN: Oh, didn't know about its deprecation.

app/controllers/api/v1/sessions_controller.rb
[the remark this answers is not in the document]
scouillard: You are right Amir. Thank you.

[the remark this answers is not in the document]
KH-Amir-TN: You're welcome,

app/controllers/api/v1/sessions_controller.rb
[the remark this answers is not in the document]
KH-Amir-TN: No, this is how it should be IMHO!
A request with a non present user in the database should still be considered as a bad request to avoid providing further data that may be used as an attack vector.
 I like and encourage what you have done,

app/javascript/components/sessions/SignIn.tsx
KH-Amir-TN: Oh, didn't mean that we should remove front end validations.
HTML 5 by default has some basic input validations with the explicit use of some build in attributes (i.e. required) and also implicitly depending of the type of the input (i.e. an input with type="email" will automatically get validated to match the HTML 5 pattern for emails).
This has nothing to do with JavaScript, it may feel good but this can introduce some troubles because will have two sources of validations one that we made and know (the validations through Yup, or react-hook-form or custom made by our JavaScript code) and those of HTML5.

app/javascript/components/sessions/SignIn.tsx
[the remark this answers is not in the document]
KH-Amir-TN: I encourage that we use the `Async/Await` syntax for readability purposes.
It's only a syntactic sugar, it doesn't have any other added value TBH,

PR #3309, opened 2022-03-22
GLv3-3202-backend

KH-Amir-TN:
Backend:
             - Set the API endpoint for the signup '/api/v1/users.json'.
             - Made some changes to Rubocop.
             - Added a request spec for the users API and prepared the tests for the signup with/without valid user data.
             - Added a password confirmation examples for the User model.
             - Improved routing compatibility between SSR (Server side routing) and CSR (Client side routing).
             - Tried defining an approach for error and request handling.
             - Removed /ruby-build.
	     - Improved API resiliency by introducing a much stricter API:
            	* API calls has to have an explicit `.json` format along with an `Accept: 'application/json'` header.
         -> The TODOs were left as comments in the code.

## Description

Allow users to signup in GLv3.

## Testing Steps

1. Pull the code.

To test the API:


hadicheaito1 (approved):

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      #     params with the fallowing format: { user: { :name, :email, :password, :password_confirmation } }
+      # Does:
+      #     - Filters and validates the recieved params( throws ActionController::ParameterMissing for missed params).
+      #     - Creates and saves a new user record in Users table with the provided parameters.
+      #     - Returns json with the fallowig format : { msg: String, data: Array[serializable objects]
+      #      ,errors: Array[String] }
+      #     and status_code: 201 (created) | 400 (bad request).
+      #         msg = response_msgs[:success] and errors = [] for successful save to db with status_code = 201.
+      #         msg = response_msgs[:failed] and errors = user.errors.to_a for unsuccessfuly save to db
+      #         with status_code = 400.
```
farhatahmad, line 17: Let's simplify this a bit. 2 or 3 lines simply explaining what this does, the params it accepts and what it returns should be good

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      def create
+        user = User.new({ provider: 'greenlight' }.merge(user_params)) # TMP fix for presence validation of :provider
+        if user.save
+          render json: {
+            msg: Rails.configuration.response_msgs[:success],
```
farhatahmad, line 22: The more I think about this, the more I feel like this is redundant. Whats the point of this while the status exists?

spec/models/user_spec.rb
```diff
@@ -13,6 +13,31 @@
...
+    context 'password confirmation' do
+      let(:valid_user) do
+        {
+          name: Faker::Name.name,
+          email: Faker::Internet.email,
+          provider: 'greenlight',
+          password: 'Password123+',
+          password_confirmation: 'Password123+'
+        }
+      end
```
farhatahmad, line 28: No need for this since you have the user_factory

spec/models/user_spec.rb
```diff
@@ -13,6 +13,31 @@
...
+
+      it 'validate the record for confirmed password' do
+        user = described_class.new valid_user
+        expect(user).to be_valid
+      end
```
farhatahmad, line 33: This test isn't really useful

spec/models/user_spec.rb
```diff
@@ -13,6 +13,31 @@
...
+
+      it 'invalidate the record for mismatched password confirmation' do
+        user = described_class.new valid_user
+        user.password_confirmation = "#{user.password}extra"
+        expect(user).to be_invalid
+        expect(user.errors.first.details).to eq({ error: :confirmation, attribute: 'Password' })
```
farhatahmad, line 39: Use the factory here instead. 

```
user = build(:user, password_confirmation: "....")
expect(user).to be_invalid
expect(user.errors.first.details).to eq({ error: :confirmation, attribute: 'Password' })
```

app/controllers/api/v1/users_controller.rb
KH-Amir-TN: Alright,

app/controllers/api/v1/users_controller.rb
KH-Amir-TN: Just a consistent response format because of the lack of one.
It's not final and has to be rectified according to our need.
1. The msg can contain a more explicit status message that may be used on client side.
2. The data is where any data sent back to client side lives.
3. The error is where any error messages flows between the backend -> frontend.

spec/models/user_spec.rb
KH-Amir-TN: Didn't know about the `build` method.
Thank you, I'll change it.

spec/models/user_spec.rb
KH-Amir-TN: Alright,

app/controllers/api/v1/users_controller.rb
KH-Amir-TN: What can I do?

spec/models/user_spec.rb
KH-Amir-TN: Makes since after using `build`.

app/controllers/api/v1/users_controller.rb
farhatahmad: The `data` and `error` parts are good, my thought is around the `msg` part. `status` already exists so that we can return a ton of different responses based on what happened to the client side, so duplicating it here seems kind of odd

app/controllers/api/v1/users_controller.rb
KH-Amir-TN: Done,

spec/models/user_spec.rb
KH-Amir-TN: Done,

spec/models/user_spec.rb
KH-Amir-TN: Done,

spec/models/user_spec.rb
KH-Amir-TN: Done,

farhatahmad (approved):

PR #3310, opened 2022-03-22
GLv3-3202-frontend-structure

KH-Amir-TN:
Frontend:
	          - Set a new project structure for the front end JavaScript code influenced by the
	          react-router-demo https://stackblitz.com/edit/github-agqlf5/ :
	              -> main.jsx: An entry point where Routes, Layouts, Global context is provided.
	              -> app.jsx: The parent route component for the `/` endpoint.
	              -> components/: A directory where custom components resides.
	              -> components/styling: A directory where custom components with no business logic resides (i.e. custom buttons).
	              -> components/forms: A directory where custom forms and form related components resides.
	              -> routes: A directory where sub routes resides.
	          - Created a `comonents/styling/buttons/ButtonLink.jsx` component that wraps/connects a react bootstrap button component with the react router API.

                  TODO:
  		    - Create the signup form.
## Description

Defined a project structure for GLv3 frontend code.
## Testing Steps

1. Pull the code.
2. Install the dependencies `bundle install && npm|yarn install`.
3. Clean the previous assets build by running `rm app/assets/builds/*` (This won't remove .keep since it's hidden).
4. Clean the database and tmp files for a better isolation by running `rails tmp:clear && rails db:schema:cache:clear && rails db:migrate:reset`
5. Run the linter and specs `bundle exec rubocop --parallel && bundle exec rspec`
6. Run `./bin/dev` to run the assets builders processes and the Puma server all at once.

To test the API:
    1. Make a POST request to `/api/v1/users.json`.
     1.1 The request has to have an `Accept: application/json` header.
     1.2. The request body should have a JSON representation with this format: 
     `{ user: { :name, :email, :password, :password_confirmation } }` .
     -> When not specifying the `.json` format the API should render back the `components#index` so that CSR handles the request.
     -> When not specifying the `Accept` header correctly the same scenario should occur.
     -> For unknown routes in the server side the CSR should take charge of handling it.
     -> Not respecting the user data required format will return a 400 BAD request with a JSON object in the response body having an errors array with a message **Invalid or Missing parameters.**
     -> Any unpermitted values given in the user data will be filtered, only `name, email ,password,...` will pass.
 
 To test the UI:
    1.  Entering `/` should render this:    
image

    2. Entering `/signup` should render this:    
image

    3. Entering anything else should render this:
image

    4:  Clicking on the Signup button should have the same outcome as 2).

PR #3311, opened 2022-03-22
GLv3-3202-frontend-form

KH-Amir-TN:
Frontend:

	          - Created a Signup route that renders `routes/Signup.jsx` for the `/signup` endpoint.
	          - Created a `components/forms/SignupForm.jsx` form component that describes the signup form.
	          - Created a `helpers/forms/SignupFormHelpers.jsx` module to bundle the helpers for the signup form component.
	          - Created a `components/forms/FormControl.jsx` component that wraps/connects a react bootstrap form control component with its parent form context.
		  - Created a `components/forms/Form.jsx` component that wraps a react bootstrap form connecting any children with the react hook form context.
		  - Created a `components/stylings/Spinner.jsx` component which defines a react bootstrap spinner with default props.

                  TODO:
                    - Use axios and react-query instead of the fetch API.
                    - Introduce a way to check for email uniqueness asynchronously from the client side.
                    - Add and use Yup.
		    - Improve the form performance.
                    - Improve API server side errors handling.
                    - Use typescript and prop types.
                    - Add any required tests.
                    - Configure the eslinter.
		    - Add feedback for the signup operations.
## Description

Added the signup form.

## Testing Steps

1. Pull the code.

To test the API:
 
 To test the UI:
    1.  Entering `/` should render this:    
image

    2. Entering `/signup` should render this:    
image

    3. Entering anything else should render this:
image

    4.  Clicking on the Signup button should have the same outcome as 2). 
    5. Click the home button in the Signup page should render the the index page `/`.
    6. Fill the form, validations should run as expected for each field respectively and on real time:
       6.1 The full name field shouldn't be blank.
       6.2 The email field shouldn't be blank and should match this pattern `/\S+@\S+\.\S+/` (This will change in the future since it raises security concerns.).
       6.3 The password field shouldn't be blank and should have a password of at least 8 characters in length.
       6.4 The password confirmation should match exactly the password field entered value.       
image
    -> When all fields are valid the form state becomes valid the form UI should become:    
image
   -> When the form is submitting;
       - A spinner with a growing circle animation should appear next to Submit button and the button should become disabled as fallow:       
image
    7. Data won't get submitted only if the form is valid (all fields becomes valid).
    8. The result of the form submission can be viewed with the dev tools along with the puma server logs and the users table entries in the db.


KH-Amir-TN:
Pending for #3310 to be merged.

farhatahmad:
@KH-Amir-TN Can you take a look at the Sonarcloud issue?

farhatahmad (approved):

farhatahmad:
Yep :)

PR #3312, opened 2022-03-22
Created Recording Model

farhatahmad:

### DIFF
--- a/spec/factories/recording_factory.rb
+++ b/spec/factories/recording_factory.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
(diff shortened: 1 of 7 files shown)


spec/factories/recording_factory.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :recording do
+    room
```
hadicheaito1, line 5: Does this create a room using room factory? Then assigns the recording to that room?

farhatahmad: Yep

scouillard (approved):

PR #3315, opened 2022-03-24
Allow user to view their rooms

### LINKED ISSUE #3162: Allow user to view a list of all their rooms
- Front End

- Back End
  - Request to get all the rooms for the current user

hadicheaito1:
closes #3162 

## Description

- When a user goes to /rooms page, they are able to view the rooms that belong to them
- Each box contains room name, last session (to be integrated still), copy link icon (to be integrated still), and start room button (to be integrated still)
- Each card is clickable which takes you to the specifc room page rooms/:friendly_id
- RSpec test for rooms index get request

- NOTE: 
  - This is not the final look, but I tried to make the overall idea look similar to the mock ups to make it easier in the future
  - current_user needs to be implemented for the rooms/index to return dynamically and for the rspec test to work dynamically as well

## Testing Steps

Steps will change slightly when current_user is implemented
1. Pull branch and code
2. Open rails console
3. Create user with id:1
4. Create multiple rooms with user_id equal the to id of the user you created in step 3 (id:1)
5. Go to /rooms path, rooms should be visible here

image


app/javascript/components/RoomCard.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+export default function RoomCard(props) {
+  const  {id, name} = props
+  const navigate = useNavigate() 
+  const handleClick = useCallback( () => { navigate(id)}, [id] )
+
+  return (
+    <Container>
+      <Card style={{ width: '14rem' }} border='dark'>
+        <Card.Body id='rooms-card' onClick={handleClick}>
```
hadicheaito1, line 15: Since the card.body is now clickable (takes you to specific room), the Start and Copy url buttons are covered. I played around with the z-index of the start button and link button but it made no difference. 
Any ideas are appreciated here.

PR #3323, opened 2022-03-28
Added the frontend components for Show Room

farhatahmad:
## Description

Added a new page that can be accessed by going to `/rooms/room_friendly_id`. The page should display the room name and some random tab buttons. The `Start Session` button doesn't work, but the `home` button does

## Testing Steps

1. Create a room in your console
2. Get the room's friendly id
3. Go to `/rooms/FRIENDLY_ID` and ensure the room name is correct

image

### DIFF
--- a/app/javascript/components/rooms/RoomView.tsx
+++ b/app/javascript/components/rooms/RoomView.tsx
@@ -1,0 +1,45 @@
+import React, {useEffect, useState} from "react";
+import axios from "axios"
+import {Row, Col, Tabs, Tab} from "react-bootstrap";
+import ButtonLink from "../stylings/buttons/ButtonLink";
+import FeatureTabs from "./FeatureTabs";
+import {Link, useParams} from "react-router-dom";
+import {useQuery} from "react-query";
+import { Spinner } from "../stylings/Spinner"
+import {House} from "react-bootstrap-icons";
+
+export default function RoomView() {
+  const { friendly_id } = useParams()
+
+  const { isLoading, error, data: room, isFetching } = useQuery("getRoom", () =>
+    axios.get(`/api/v1/rooms/${friendly_id}.json`, {
+      headers: {
+        'Content-Type': 'application/json',
+        'Accept': 'application/json'
+      }
+    }).then(resp => resp.data.data)
+  );
+
+  if (isLoading) return <Spinner />
+
+  return (
+    <>
+      <Row>
+        <Col>
+          <Link to="/rooms">
+            <House />
+          </Link>
+        </Col>
+      </Row>
+      <Row className="my-5">
+        <Col>
+          { room.name }
+        </Col>
+        <Col>
+          <ButtonLink to="/" variant="primary" className="float-end">Start Session</ButtonLink>
+        </Col>
+      </Row>
+      <FeatureTabs />
+    </>
+  )
+}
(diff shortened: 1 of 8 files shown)


hadicheaito1 (approved):

app/javascript/app.jsx
```diff
@@ -2,16 +2,13 @@ import React from "react";
 import { Col, Container, Row } from "react-bootstrap";
 import { Outlet } from "react-router-dom";
 import ButtonLink from "./components/stylings/buttons/ButtonLink";
```
hadicheaito1, line 4: This can be removed now

app/javascript/components/rooms/RoomView.tsx
```diff
@@ -0,0 +1,45 @@
...
+export default function RoomView() {
+  const { friendly_id } = useParams()
+
+  const { isLoading, error, data, isFetching } = useQuery("getRoom", () =>
+    axios.get(`/api/v1/rooms/${friendly_id}.json`, {
+      headers: {
+        'Content-Type': 'application/json',
+        'Accept': 'application/json'
+      }
+    }).then(resp => resp.data)
```
hadicheaito1, line 20: Not sure what approach we want to use for these since I know they will be moved to another file soon, but you can do something like this (using useState) which would allow you to just access room attributes easier:
```suggestion
    }).then(function (resp) {
      setRoom(resp.data.data)
    })
```

Then you'd be able to do room.name instead of data.data.name.

But then 'data' returned from useQuery won't really be used anymore, so I am not sure what is best.

scouillard (dismissed):

app/javascript/components/rooms/RoomView.tsx
```diff
@@ -0,0 +1,45 @@
...
+import FeatureTabs from "./FeatureTabs";
+import {Link, useParams} from "react-router-dom";
+import {useQuery} from "react-query";
+import { Spinner } from "../stylings/Spinner"
+import {House} from "react-bootstrap-icons";
+
+export default function RoomView() {
+  const { friendly_id } = useParams()
+
+  const { isLoading, error, data: room, isFetching } = useQuery("getRoom", () =>
```
farhatahmad, line 14: @hadicheaito1 What do you think of this?

app/javascript/components/rooms/RoomView.tsx
```diff
@@ -0,0 +1,45 @@
...
+    axios.get(`/api/v1/rooms/${friendly_id}.json`, {
+      headers: {
+        'Content-Type': 'application/json',
+        'Accept': 'application/json'
+      }
+    }).then(resp => resp.data.data)
```
farhatahmad, line 20: It's set here

app/javascript/components/rooms/RoomView.tsx
```diff
@@ -0,0 +1,45 @@
...
+      <Row>
+        <Col>
+          <Link to="/rooms">
+            <House />
+          </Link>
+        </Col>
+      </Row>
+      <Row className="my-5">
+        <Col>
+          { room.name }
```
farhatahmad, line 36: and used here

app/javascript/components/rooms/RoomView.tsx
hadicheaito1: Yeah that looks clean :+1:

scouillard (approved):

PR #3327, opened 2022-03-28
Add Sign In

scouillard:
## Description

This PR implements 3 features/mechanics that arose from the Sign In process: 

- the sign in form (SignIn component)
- the server side api hooks for signing in (Sessions controller)
- the current user api hooks and display (CurrentUser, AuthContext component and Sessions helper)

**Sign In Form**
User enter email and password in Sign In form.
If email and password correspond to a user in the server database, user is redirected to Home page on Form submit.

**Server Side API Hooks**
Server checks if User exists.
Server checks if User and Password match via authenticate method.
If success, return success status, signed_in as true and store User id in session.
If error, return error status and signed_in as false.

**Current User**
Frontend:
If signed in, the Current User information is displayed (name, email, provider, signed_in?).
The currentUser object is accessible to any component via the useAuth() method import.
Backend:
The current_user helper method can now be accessed anywhere in server side to display current user.

## How To Access Current User in a Component

The whole app tree has been wrapped inside the AuthProvider component.

As per AuthContext.tsx (in components/sessions), every AuthProvider children will inherit the currentUser value/props. 

image

To use the currentUser value/props inside a component, import the context ` import { useAuth } from '../sessions/AuthContext';` and use the custom hook useAuth() `const currentUser = useAuth();` 

The object returned by useAuth() now contains the current_user data, as per the API hook set at /api/v1/sessions/signed_in (SessionsController).

## Testing Steps

1. Navigate to /sign_in

image

2. Fill in the email and password input as an existing User in your local database
If the authentication is successful, you will be redirected to the home component.
Also, the user information is displayed to confirm the persistence of the session.

image

## TODO:
- Clean up code

  1. createSession() api call does not belong in Form/SignIn component.
  2. fetchCurrentUser() api call does not belong in AuthContext component.
  3. review variables names with team.
  4. use Amir template for Sign In form.

- Use axios instead of fetch for api calls
- Add Alerts if Sign In is success or error
- Use Typescript


app/controllers/api/v1/sessions_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+
+      # GET /api/v1/sessions/signed_in
+      def signed_in
+        if current_user
+          render json: { current_user:, signed_in: true }
+        else
+          render json: { signed_in: false }
+        end
+      end
```
farhatahmad, line 34: Why not do this as `index`?

app/javascript/application.jsx
```diff
@@ -0,0 +1,18 @@
...
+const queryClient = new QueryClient();
+
+document.addEventListener("DOMContentLoaded", () => {
+  render(
+    <QueryClientProvider client={queryClient}>
+      <App />
+    </QueryClientProvider>,
+    document.body.appendChild(document.createElement('div'))
+  );
+});
```
farhatahmad, line 18: Why not just do this in `app.jsx`?

app/helpers/sessions_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+  # Signs In the user
+  def sign_in(user)
+    session[:user_id] = user.id
+  end
+
+  # Returns the current signed in User (if any)
+  def current_user
+    @current_user ||= User.find_by(id: session[:user_id])
+  end
+end
```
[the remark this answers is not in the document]
scouillard: Agreed - helpers will be removed.

app/controllers/api/v1/sessions_controller.rb
scouillard: Indeed, we could make use of the default route.

app/javascript/application.jsx
scouillard: This file has no use and will be deleted.

app/helpers/sessions_helper.rb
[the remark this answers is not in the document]
scouillard: There is no /api/v1/ scope for ApplicationController at the moment.
current_user could reside in SessionsController for now.
Could also be in UsersController.

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+      # DELETE /api/v1/sessions/:id
+      def destroy
+        session[:user_id] = nil
+        render json: { signed_in: false }, status: :ok
+      end
+
+      # Returns the current signed in User (if any)
+      def current_user
+        @current_user ||= User.find_by(id: session[:user_id])
+      end
```
farhatahmad, line 39: Let's move this to the `ApplicationController` so it's accessible across all controllers

scouillard: Right.

app/helpers/sessions_helper.rb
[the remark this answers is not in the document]
hadicheaito1: current_user method should work if it is moved to application_controller (just tried it locally) as a public method.
The reason we need it in application_controller is because current_user needs to be accessible throughout the application.

farhatahmad (approved):

PR #3345, opened 2022-03-31
Add Sign Out

scouillard:
## Description

If user is signed in, user can now sign out.

**Frontend**
Link to (Sign In/Sign Out) will switch depending on user status.
Anchor has onClick event attribute to trigger DELETE request to /api/v1/sessions/sign_out

**Backend**
/sign_out route was created (instead of using Rails delete default route) to avoid having to use session[:user_id]
session[user_id] and @current_user is set to nil on request.

## Testing Steps

Sign In.

image

Sign Out.

image


scouillard:
Should we add /sign_in route as we (seem to) have to do it anyways for /sign_out ?
I think it is more explicit than using default sessions#index.
Or maybe there is a better way to do the Routing?

hadicheaito1:
> Should we add /sign_in route as we (seem to) have to do it anyways for /sign_out ? I think it is more explicit than using default sessions#index. Or maybe there is a better way to do the Routing?

I think using index when possible is following REST more, so I don't think it's a bad idea to keep it the way it is. The others can verify though :stuck_out_tongue:

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -43,6 +43,12 @@ def session_params
       def sign_in(user)
         session[:user_id] = user.id
       end
+
+      # Signs Out the user
+      def sign_out
+        session[:user_id] = nil
+        @current_user = nil
```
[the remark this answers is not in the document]
scouillard: You are right - the behaviour does not change when it is removed.

Is it me or is there some Rails magic going on?
Shouldn't the right side be ignored if @current_user isn't nil?

image

scouillard:
Link to Sign Up should also be removed from CurrentUser/Navbar if User is signed in.

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -43,6 +43,11 @@ def session_params
       def sign_in(user)
         session[:user_id] = user.id
       end
+
+      # Signs Out the user
+      def sign_out
+        session[:user_id] = nil
+      end
```
farhatahmad, line 50: No need to break this off into its own function

farhatahmad, line 50: No need to break this off into its own function

scouillard: Done

app/assets/stylesheets/application.bootstrap.scss
```diff
@@ -1,4 +1,4 @@
-$primary: #1a454a;
+$primary: #043c54;
```
hadicheaito1, line 1: In other applications, we have the primary set to #1A454A

app/controllers/api/v1/sessions_controller.rb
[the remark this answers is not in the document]
KH-Amir-TN: @scouillard I think I get the reason behind your confusing.
The `@current_user` is an instance variable and it stays alive only for the whole `request/response` lifecycle meaning that with each new request a new instance variable will get created.
Upon initial interaction and since `@current_user` will be `nil` the instruction `User.find_by id:...` will get executed fetching the current user data from the db according to its stored id in the session hash.
After this and until a response is issued back the instance variable `@current_user` will always hold that value.
Once a response is issued back Rails garbage collects all created instances for that request and all of the controller instances with their properties will get lost.
Upon the next request the whole process will repeat meaning after a successful signout operation which resets the session data for the client the upcoming requests will send back no user id and each controller instance will have a `nil` `@current_user`.

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -6,7 +6,7 @@ class SessionsController < ApplicationController
       # TODO: samuel - Bypass CSRF token for now
       skip_before_action :verify_authenticity_token
 
-      # GET /api/v1/sessions/signed_in
+      # GET /api/v1/sessions
       def index
         if current_user
           render json: { current_user:, signed_in: true }
```
KH-Amir-TN, line 12: Let's remember to filer out any sensitive data later on.

config/routes.rb
```diff
@@ -6,7 +6,11 @@
   # All the Api endpoints must be under /api/v1 and must have an extension .json.
   namespace :api do
     namespace :v1 do
-      resources :sessions, only: %i[index create destroy]
+      resources :sessions, only: %i[index create] do
+        collection do
+          delete 'sign_out', to: 'sessions#destroy'
```
KH-Amir-TN, line 11: Is `resources :sessions, only: %i[index create destroy]` isn't suiting your needs?

KH-Amir-TN (dismissed):

config/routes.rb
scouillard: Seems standard for destroying a session.
Default URI for destroy is /api/v1/sessions/:id
I figured we would rather not be passing the session id around.

app/assets/stylesheets/application.bootstrap.scss
hadicheaito1: Oh I see. I'm not sure if we should be using that or the BN "official" colour.
@farhatahmad What do you think?

farhatahmad: Leave it as #1a454a for now - we'll need to change it to blue later anyways

app/controllers/api/v1/sessions_controller.rb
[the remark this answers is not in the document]
scouillard: Thank you Amir.
I wasn't thinking in terms of request/response.

hadicheaito1 (approved):

config/routes.rb
KH-Amir-TN: You got reason,

app/javascript/components/user/CurrentUser.jsx
```diff
@@ -28,8 +34,15 @@ export default function CurrentUser() {
...
         { currentUser?.signed_in.toString() }
       </span>
-      <Link to="/signin">Sign In</Link>
-      <Link to="/signup">Sign Up</Link>
+
+      { currentUser?.signed_in
+        ? <Link to="/" onClick={signOut}>Sign Out</Link>
+        : (
+          <>
+            <Link to="/sign_in">Sign In</Link>
```
farhatahmad, line 42: `signin`

config/routes.rb
farhatahmad, line 11: I would change this to `signout`

farhatahmad, line 11: I would change this to `signout`

farhatahmad (approved):

PR #3349, opened 2022-04-01
Refactor Auth/Sign In

scouillard:
## Description

Auth / Sign In process has been refactored.
- Axios is being used instead of Fetch.
- Custom hooks now reside in /hooks/

## Testing Steps

Pull the code and everything should be working as before.


farhatahmad:
I think this can be closed. Most of these changes were already done in my refactor

PR #3355, opened 2022-04-04
Backend: Added the room session start feature.

KH-Amir-TN:
## Description

Backend: Added the room session start feature.
## Testing Steps

1. Pull the code.

### DIFF
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -4,7 +4,7 @@ module Api
   module V1
     class RoomsController < ApplicationController
       skip_before_action :verify_authenticity_token # TODO: amir - Revisit this.
-      before_action :find_room, only: :show
+      before_action :find_room, only: %i[show start]
 
       # GET /api/v1/rooms.json
       # Returns: { data: Array[serializable objects(rooms)] , errors: Array[String] }
@@ -18,6 +18,30 @@ module Api
 
       def show
         render_json data: @room, status: :ok
+      end
+
+      # POST /api/v1/rooms/:friendly_id/start.json
+      # Returns: { data: Array[serializable objects] , errors: Array[String] }
+      # Does: Starts the Room meeting and joins in the meeting starter.
+
+      def start
+        # TODO: amir - Check the legitimately of the action.
+        bbb_api = BigBlueButtonApi.new
+        meeting_starter = current_user ? "user(id):#{current_user.id}" : 'unauthenticated user'
+        options = { logoutURL: request.headers['Referer'] || root_url }
+        retries = 0
+        begin
+          logger.info "Starting meeting for room(friendly_id):#{@room.friendly_id} by #{meeting_starter}."
+          join_url = bbb_api.start_meeting room: @room, meeting_starter: current_user, options: options
+          logger.info "meeting successfully started for room(friendly_id):#{@room.friendly_id} by #{meeting_starter}."
+
+          render_json data: { join_url: }, status: :created
+        rescue BigBlueButton::BigBlueButtonException => e
+          retries += 1
+          logger.info "Retrying meeting start for room(friendly_id):#{@room.friendly_id} because of error(key): #{e.key} #{retries} times..."
+          retry unless retries >= 3
+          raise e
+        end
       end
 
       private
--- a/app/services/big_blue_button_api.rb
+++ b/app/services/big_blue_button_api.rb
@@ -12,6 +12,19 @@ class BigBlueButtonApi
     @bbb_server ||= BigBlueButton::BigBlueButtonApi.new(bbb_endpoint, bbb_secret, '1.8')
   end
 
+  # Start a meeting for a specific room and returns the join URL.
+  def start_meeting(room:, meeting_starter:, options: {})
+    # TODO: amir - Revisit this.
+    create_options = default_create_opts.merge(options)
+    join_options = { join_via_html5: true } # TODO: amir - Revisit this (createTime,...).
+
+    user_name = meeting_starter&.name || 'Someone'
+    password = (meeting_starter && create_options[:moderatorPW]) || create_options[:attendeePW]
+
+    bbb_server.create_meeting room.name, room.friendly_id, create_options
+    bbb_server.join_meeting_url room.friendly_id, user_name, password, join_options
+  end
+
   private
 
   def bbb_endpoint
@@ -20,5 +33,21 @@ class BigBlueButtonApi
 
   def bbb_secret
     ENV.fetch 'BIGBLUEBUTTON_SECRET', '8cd8ef52e8e101574e400365b55e11a6'
+  end
+
+  def default_create_opts
+    {
+      # TODO: amir - revisit this.
+      record: true,
+      logoutURL: 'http://localhost',
+      moderatorPW: 'mp',
+      attendeePW: 'ap',
+      moderatorOnlyMessage: 'Welcome Moderator',
+      muteOnStart: false,
+      guestPolicy: 'ALWAYS_ACCEPT',
+      'meta_gl-v3-listed': 'public',
+      'meta_bbb-origin-version': 3,
+      'meta_bbb-origin': 'Greenlight'
+    }
   end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -12,7 +12,9 @@ Rails.application.routes.draw do
         end
       end
       resources :users, only: [:create]
-      resources :rooms, only: %i[show index], param: :friendly_id
+      resources :rooms, only: %i[show index], param: :friendly_id do
+        post '/start', to: 'rooms#start', as: :start_meeting, on: :member
+      end
     end
   end
   match '*path', to: 'components#index', via: :all # Enable CSR for full fledged http requests.
--- a/spec/services/big_blue_button_api_spec.rb
+++ b/spec/services/big_blue_button_api_spec.rb
@@ -5,6 +5,18 @@ require 'bigbluebutton_api'
 
 describe BigBlueButtonApi, type: :service do
   let(:bbb_service) { described_class.new }
+  let(:default_create_opts) do
+    {
+      moderatorPW: 'mp',
+      attendeePW: 'ap'
+    }
+  end
+
+  let(:default_join_opts) do
+    {
+      join_via_html5: true
+    }
+  end
 
   before do
     ENV['BIGBLUEBUTTON_ENDPOINT'] = 'http://test.com/bigbluebutton/api'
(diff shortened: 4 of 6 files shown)


config/routes.rb
```diff
@@ -1,14 +1,15 @@
 # frozen_string_literal: true
 
 Rails.application.routes.draw do
-  root 'components#index'
+  root 'components#index', via: :all
```
hadicheaito1, line 4: This is just telling us that it can use any type of request (get, post,...), but it doesn't really do anything right? WIth or without ```via: :all``` it works the same

config/routes.rb
```diff
@@ -1,14 +1,15 @@
...
 
   # All the Api endpoints must be under /api/v1 and must have an extension .json.
   namespace :api do
     namespace :v1 do
       resources :sessions, only: %i[index create destroy]
       resources :users, only: [:create]
       resources :rooms, only: %i[show index], param: :friendly_id
+      post '/room/:friendly_id/start', to: 'rooms#start', as: :start_session
```
hadicheaito1, line 12: I believe it should be ```friendlyId``` after Ahmad's most recent commit, in order to keep consistency.
@farhatahmad correct me if I'm wrong

KH-Amir-TN: You'd like to keep consistency between the frontend and the backend codebase?
I don't mind it but in rails by default we usually use the `snake_case` and not the `camel case`.

config/routes.rb
KH-Amir-TN: It doesn't have that much of added value, I've just included it to complete all possible routes even for `POST/PUT/...` requests on the `root_path` to avoid having `Routing Error::No route matches [...] "/"`

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -26,6 +26,31 @@ def show
...
+      def start
+        # TODO: amir - Check the legitimately of the action.
+        bbb_api = BigBlueButtonApi.new
+        room_id = @room.friendly_id
+        session_starter = current_user ? "user(id):#{current_user.id}" : 'unauthenticated user'
+        options = { logoutURL: request.headers['Referer'] || root_url }
+        retries = 0
+        begin
+          logger.info "Starting session for room(friendly_id):#{room_id} by #{session_starter}."
+          join_url = bbb_api.start_session room: @room, session_starter: current_user, options: options
```
farhatahmad, line 42: I would just do it in one line 
```
BigBlueButtonApi.new.start_session room: ....
```

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -26,6 +26,31 @@ def show
...
       end
 
+      # POST /api/v1/room/:friendly_id/start.json
+      # Returns: { data: Array[serializable objects] , errors: Array[String] }
+      # Does: Starts the Room session and joins in the room starter.
+
+      def start
+        # TODO: amir - Check the legitimately of the action.
+        bbb_api = BigBlueButtonApi.new
+        room_id = @room.friendly_id
```
farhatahmad, line 36: No need for this - I would just keep referring to it as `@room.friendly_id`

app/services/big_blue_button_api.rb
```diff
@@ -12,13 +12,57 @@ def bbb_server
     @bbb_server ||= BigBlueButton::BigBlueButtonApi.new(bbb_endpoint, bbb_secret, '1.8')
   end
 
+  # Start a session for a specific room and returns the join URL.
+  def start_session(room:, session_starter:, options: {})
+    # TODO: amir - Revisit this.
+    create_options = default_create_opts.merge(options)
+    join_options = default_join_opts.merge(create_options) # TODO: amir - Revisit this (createTime,...).
+
+    meeting_id = room.friendly_id
```
farhatahmad, line 21: I'd avoid this too - adds confusion around naming. Just set `meeting_id: room.friendly_id`

app/services/big_blue_button_api.rb
```diff
@@ -12,13 +12,57 @@ def bbb_server
...
+    user_name = session_starter&.name || 'Someone'
+    password = (session_starter && create_options[:moderatorPW]) || create_options[:attendeePW]
```
farhatahmad, line 23: We don't actually need the password anymore - I think there's a new param called `role` set to either `viewer` or `moderator`

app/services/big_blue_button_api.rb
```diff
@@ -12,13 +12,57 @@ def bbb_server
...
+
+    create_session meeting_name: room.name, meeting_id:, options: create_options
+    get_session_join_url meeting_id:, user_name:, password:, options: join_options
+  end
+
   private
 
+  def create_session(meeting_name:, meeting_id:, options: {}, modules: nil)
+    bbb_server.create_meeting meeting_name, meeting_id, options, modules
+  end
```
farhatahmad, line 33: I wouldn't break this off into another method - just add more complexity

app/services/big_blue_button_api.rb
```diff
@@ -12,13 +12,57 @@ def bbb_server
...
+
+  def get_session_join_url(meeting_id:, user_name:, password:, options: {})
+    bbb_server.join_meeting_url meeting_id, user_name, password, options
+  end
```
farhatahmad, line 37: Same as above

app/services/big_blue_button_api.rb
```diff
@@ -12,13 +12,57 @@ def bbb_server
...
+      guestPolicy: 'ALWAYS_ACCEPT',
+      'meta_gl-v3-listed': 'public',
+      'meta_bbb-origin-version': 3,
+      'meta_bbb-origin': 'Greenlight'
+    }
+  end
+
+  def default_join_opts
+    {
+      join_via_html5: true
```
farhatahmad, line 65: I believe this can be dropped

config/routes.rb
farhatahmad: Yeah we should keep it as `snake_case` in Rails but `camelCase` in React

spec/services/big_blue_button_api_spec.rb
```diff
@@ -5,6 +5,26 @@
...
+      'meta_gl-v3-listed': 'public',
+      'meta_bbb-origin-version': 3,
+      'meta_bbb-origin': 'Greenlight'
+    }
+  end
+
+  let(:default_join_opts) do
+    {
+      join_via_html5: true
+    }
```
farhatahmad, line 26: Why do we need to define these? Are they not set in the service?

config/routes.rb
```diff
@@ -9,6 +9,7 @@
...
```
farhatahmad, line 12: There's a cleaner way to do this - https://guides.rubyonrails.org/routing.html#adding-more-restful-actions

app/controllers/api/v1/rooms_controller.rb
KH-Amir-TN: That means creating new instance for each retry (including creating new `BigBlueButton::BigBlueButtonApi instance`) plus it further increases the line length and decreases readability.
Do you like me to change it?

app/controllers/api/v1/rooms_controller.rb
KH-Amir-TN: Makes sense, will do it

app/services/big_blue_button_api.rb
KH-Amir-TN: Alright, I'll change it

app/services/big_blue_button_api.rb
KH-Amir-TN: Didn't know about it, I'll check the BBB API and use the new approach.

app/services/big_blue_button_api.rb
KH-Amir-TN: Alright, I'll remove it

app/services/big_blue_button_api.rb
KH-Amir-TN: Alright, I'll remove it

app/services/big_blue_button_api.rb
KH-Amir-TN: The `default_join_opts` ?

config/routes.rb
KH-Amir-TN: Alright, I'll check it.

KH-Amir-TN: Thank you!

spec/services/big_blue_button_api_spec.rb
KH-Amir-TN: For isolation purposes.
Do you need me to remove it?

app/controllers/api/v1/rooms_controller.rb
KH-Amir-TN: DONE

app/services/big_blue_button_api.rb
KH-Amir-TN: DONE

app/services/big_blue_button_api.rb
KH-Amir-TN: DONE

app/services/big_blue_button_api.rb
KH-Amir-TN: DONE

app/services/big_blue_button_api.rb
KH-Amir-TN: DONE

config/routes.rb
KH-Amir-TN: DONE

app/services/big_blue_button_api.rb
KH-Amir-TN: Unfortunately, it appears that there's some problems with using `role` parameter.
I'll show you the issue, for now I'll keep it this way until we figure out what to do.

app/controllers/api/v1/rooms_controller.rb
farhatahmad: No that's a fair point. Lets leave it as is

spec/services/big_blue_button_api_spec.rb
farhatahmad: Yeah . `start_session` calls the private method, so it's fine to test them both together

farhatahmad:
To avoid confusion, lets refer to BigBlueButton "sessions" as "meetings" instead. Sessions are used to refer to users signing in and out

spec/services/big_blue_button_api_spec.rb
KH-Amir-TN: Besides that, still need the default `hashes` for the method call assertion with the current setting this  allows for duplication and make tests dependent on the underlying implementation [LINE 63](https://github.com/bigbluebutton/greenlight/pull/3355/files#diff-63e7617af1a7aeb70cbf8dcc62e3bc240d1d99ae2f3490d0b293b3710f4660d8R63).

1. We can make the methods `public` in such case so that we use them directly for assertion and improve the tests quality without having to type them statically or to keep the `let` helpers in place plus having to synchronize them with the actual returned values.
OR, 
2. We can just make the memoized helper `default_create_opts` something much simpler since it's a mock we don't need it to store all legit BBB parameters or to be a copy of the the actual `#default_create_opts` and keep the mocking.
I'll go with `2)` until I get your suggestions.
What do you think?

config/routes.rb
```diff
@@ -12,7 +12,9 @@
         end
       end
       resources :users, only: [:create]
-      resources :rooms, only: %i[show index], param: :friendly_id
+      resources :rooms, only: %i[show index], param: :friendly_id do
+        post '/start', to: 'rooms#start', as: :start_session, on: :member
```
farhatahmad, line 16: I would drop the `as: :start_session` here

farhatahmad (approved):

PR #3358, opened 2022-04-04
Ran eslint for the first time

farhatahmad:

### DIFF
--- a/app/javascript/components/rooms/Room.jsx
+++ b/app/javascript/components/rooms/Room.jsx
@@ -1,18 +1,17 @@
-import React from "react";
-import axios from "axios"
-import {Col, Row} from "react-bootstrap";
-import ButtonLink from "../stylings/buttons/ButtonLink";
-import FeatureTabs from "./FeatureTabs";
-import {Link, useParams} from "react-router-dom";
-import {Spinner} from "../stylings/Spinner"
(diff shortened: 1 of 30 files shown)


hadicheaito1 (approved):

app/javascript/components/rooms/Room.jsx
```diff
@@ -1,18 +1,17 @@
...
+import { Link, useParams } from 'react-router-dom';
+import { House } from 'react-bootstrap-icons';
+import ButtonLink from '../stylings/buttons/ButtonLink';
+import FeatureTabs from './FeatureTabs';
+import Spinner from '../stylings/Spinner';
+import useRoom from '../../hooks/queries/rooms/useRoom';
 
 export default function Room() {
-  const { friendly_id } = useParams()
+  const { friendlyId } = useParams();
```
KH-Amir-TN, line 11: This introduces a bug.
Either: 
1. We change it to `friendly_id: friendlyId` 
2. We use `friendly_id` (Which is absurd.).
3. Change the route param in the room route path attribute in `main.jsx` to `path='/rooms/:friendlyId'`.

farhatahmad: https://github.com/bigbluebutton/greenlight/pull/3364

PR #3367, opened 2022-04-05
More refactoring to try and get eslint to pass

farhatahmad:


hadicheaito1 (approved):

app/javascript/components/stylings/Spinner.jsx
```diff
@@ -1,15 +1,14 @@
...
-export default function Spinner(props) {
+export default function Spinner() {
   return (
     <BootstrapSpinner
       as="span"
       animation="grow"
       size="sm"
       role="status"
       aria-hidden="true"
-      {...props}
```
KH-Amir-TN, line 12: This will introduce a bug in the `UI` once we need to overwrite the default attributes.

PR #3371, opened 2022-04-06
CVE-2015-9284: potential issue with omniauth accepting get requests

jfederico:
## Description

We have a long standing issue open due [CVE-2015-9284](https://nvd.nist.gov/vuln/detail/CVE-2015-9284) related to a vulnerability in omniauth.

And although may not be critical for this application, this PR mitigates the risk as recommended [here](https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284) 

## Testing Steps

After the update, accessing any of the authentication providers enabled from a browser should result in a 404.

E.g.
https://yourserver/auth/google


farhatahmad (approved):

defnull:
The `/signin` route triggers a redirect to `/auth/:provider` if exactly one omniauth provider is configured. Redirects are GET requests. This PR breaks this flow and thus most larger greenlight instances out there.

defnull:
Also, the CVE seems to only affect applications were multiple social logins can be attached to a single user account. This is not the case for Greenlight.

PR #3373, opened 2022-04-06
Rooms front end test

hadicheaito1:
## Description

- Front end test for Rooms page, using Jest
## Testing Steps

- Pull code
- run ```jest -u```
- Ensure all tests pass

image

### DIFF
--- a/app/javascript/__tests__/rooms/Rooms.test.jsx
+++ b/app/javascript/__tests__/rooms/Rooms.test.jsx
@@ -1,0 +1,54 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import Rooms from '../../components/rooms/Rooms';
+import RoomsList from '../../components/rooms/RoomsList';
+import RoomsTabs from '../../components/rooms/RoomsTabs';
+import useRooms from '../../hooks/queries/rooms/__mocks__/useRooms';
+
+jest.mock('react-router-dom');
+jest.mock('../../hooks/queries/rooms/useRooms');
+
+const rooms = {
+  data:
+    [
+      {
+        name: 'Home Room',
+        friendly_id: 'dummy_id1',
+      },
+      {
+        name: 'Room 1',
+        friendly_id: 'dummy_id2',
+      },
+      {
+        name: 'Room 2',
+        friendly_id: 'dummy_id3',
+      },
+    ],
+};
+
+describe('Rooms (useRooms query test)', () => {
+  test('Rooms from query match expected rooms', () => {
+    const json = useRooms();
+    expect(json).toMatchObject(rooms);
+  });
+});
+
+describe('Rooms (Snapshot Tests)', () => {
+  it('Rooms renders rooms', () => {
+    const roomsComponent = renderer.create(<Rooms />);
+    const json = roomsComponent.toJSON();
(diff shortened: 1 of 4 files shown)


scouillard:
It seems like every aria-label, aria-control and id are dynamically rendered on every test run, which make the snapshot test fail every time.

I think one way to solve the issue would be to expect any() for the listed properties above.
https://jestjs.io/docs/snapshot-testing#property-matchers

Or to hard code the properties.
https://stackoverflow.com/questions/53354341/react-jest-snapshot-testing-with-dynamic-id

scouillard:
Running `jest -u` will re-generate snapshot artifacts for all failing snapshot tests, and should be used only when there is change in the DOM/layout.

app/javascript/__tests__/rooms/Rooms.test.jsx
```diff
@@ -0,0 +1,54 @@
...
+      {
+        name: 'Room 2',
+        friendly_id: 'dummy_id3',
+      },
+    ],
+};
+
+describe('Rooms (useRooms query test)', () => {
+  test('Rooms from query match expected rooms', () => {
+    const json = useRooms();
```
KH-Amir-TN, line 31: We are testing a mock here. 
To test queries we should mock the API calls but use the queries original implementation.

app/javascript/__tests__/rooms/Rooms.test.jsx
```diff
@@ -0,0 +1,54 @@
...
+import renderer from 'react-test-renderer';
+import Rooms from '../../components/rooms/Rooms';
+import RoomsList from '../../components/rooms/RoomsList';
+import RoomsTabs from '../../components/rooms/RoomsTabs';
+import useRooms from '../../hooks/queries/rooms/__mocks__/useRooms';
+
+jest.mock('react-router-dom');
+jest.mock('../../hooks/queries/rooms/useRooms');
+
+const rooms = {
```
KH-Amir-TN, line 11: I think this is not useful since it's in place only to test a mock.

KH-Amir-TN (changes requested):

KH-Amir-TN:
> It seems like every aria-label, aria-control and id are dynamically rendered on every test run, which make the snapshot test fail every time.
> 
> I think one way to solve the issue would be to expect any() for the listed properties above via Property Matchers. https://jestjs.io/docs/snapshot-testing#property-matchers
> 
> Or to hard code the properties directly in the prop. https://stackoverflow.com/questions/53354341/react-jest-snapshot-testing-with-dynamic-id
> 
> I've been trying both solutions so far without success.

You had this problem when pulling this PR or on one that you're working on?

scouillard:
> > It seems like every aria-label, aria-control and id are dynamically rendered on every test run, which make the snapshot test fail every time.
> > I think one way to solve the issue would be to expect any() for the listed properties above via Property Matchers. https://jestjs.io/docs/snapshot-testing#property-matchers
> > Or to hard code the properties directly in the prop. https://stackoverflow.com/questions/53354341/react-jest-snapshot-testing-with-dynamic-id
> > I've been trying both solutions so far without success.
> 
> You had this problem when pulling this PR or on one that you're working on?

On this PR.

PR #3377, opened 2022-04-07
Fixed issue with ButtonLink props

farhatahmad:

### DIFF
--- a/app/javascript/components/stylings/buttons/ButtonLink.jsx
+++ b/app/javascript/components/stylings/buttons/ButtonLink.jsx
@@ -30,5 +30,5 @@ ButtonLink.propTypes = {
   to: PropTypes.string.isRequired,
   className: PropTypes.string,
   variant: PropTypes.string,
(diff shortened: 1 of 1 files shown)


app/javascript/components/stylings/buttons/ButtonLink.jsx
```diff
@@ -5,16 +5,30 @@ import PropTypes from 'prop-types';
 
 export default function ButtonLink(props) {
   const navigate = useNavigate();
-  const { to } = props;
+  const {
+    to, className, variant, children,
+  } = props;
```
hadicheaito1, line 10: Not big, but Isn't it cleaner if it's one line? ```const { to, className, variant, children } = props;```

farhatahmad: It's an eslint rule

hadicheaito1 (approved):

scouillard (approved):

PR #3366, opened 2022-04-05
Sign In and Sign Up forms are now styled according to mock ups.

scouillard:
## Description

Add styling to forms.
Add rooms, forms, and custom helper files to store the CSS.

## Testing Steps

Make sure Sign In and Sign Up works as intended/as it was working before.

## Screenshots (if appropriate):

image

image

## TODO

Refactor the Sign In form with @KH-Amir-TN helpers/custom components.
Find a solution for Checkbox input control.

PR #3380, opened 2022-04-07
Add sign wrapper

scouillard:
## Description

Add Wrapper for Sign Up and Sign In form since they are (almost) the same.
Rename SignIn to Signin for consistency. 

Initially had a component for the "Have an Account?" message at the bottom of the (3) forms presented in the markups (sing in, sign up and join session).

It felt wrong - it is now passing a "form" prop to dictate the "Have an account" message.

The wrapper is made specially for the Sign In and Sign Up forms, but it could probably be adapted for Join Session form.


scouillard:
Depends on #3366

farhatahmad:
I'm not quite sure this makes sense to join the 2 into a `wrapper` here. I understand that they're both similar in the way they look - but thats what the bootstrap components are from. I think this actually introduces more complexity - rather than just duplicating a small amount of code and keeping them separate

scouillard:
> 



> 

I was on the fence for the Wrapper.
I thought it made sense, more so that it uses the FormHelpers we are already using.
At worst, I used this opportunity to experiment with React.

farhatahmad:
@hadicheaito1 @KH-Amir-TN Thoughts?

KH-Amir-TN:
I want to make sure that I fully understood your idea @scouillard .
So the wrapper component is meant to encapsulate the signin/sinup forms wrapping design like the card, etc...
In a consistent way?
If that's the case then I like what you're doing except that you can directly do the wrapping in the Form component which creates the react hook form context.

scouillard:
> I want to make sure that I fully understood your idea @scouillard . So the wrapper component is meant to encapsulate the signin/sinup forms wrapping design like the card, etc... In a consistent way? If that's the case then I like what you're doing except that you can directly do the wrapping in the Form component which creates the react hook form context.

Yep @KH-Amir-TN 
Doing it in the Form component would make the Form component unusable for other forms. Unless we know they all use the same design. 

Right now we have the Sign Up and Sign In form, which are the same design.
And the Join Session form, for which the Wrapper would require some tweaking to fit it in.

KH-Amir-TN:
I see what you mean,
You can create a component and call it CardForm or something relevant and do the wrapping their or keep it as is.

hadicheaito1:
I do find it makes it a bit more complicated than necessary in this case. Not with or against the idea though.

farhatahmad:
I spent some time thinking about this - I find this overcomplicates the views for these 2 pages. A bit of duplication is okay (imo) as long as it makes sense, which I think is the case here

PR #3385, opened 2022-04-12
Back-end: Recordings table

hadicheaito1:
## Description

- Created Recordings controller
- Added attributes to Recording table
- Added get recordings route
- Supports get request to get all recordings for the current user's rooms
- Created Formats model
- Rpsec testing


KH-Amir-TN (dismissed):

db/migrate/20220411162947_add_users_to_recordings.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddUsersToRecordings < ActiveRecord::Migration[7.0]
+  def change
+    add_column :recordings, :users, :integer
```
farhatahmad, line 5: You can merge all the `add_column` migrations into one

db/migrate/20220411192122_create_formats.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateFormats < ActiveRecord::Migration[7.0]
+  def change
+    create_table :formats do |t|
+      t.belongs_to :recording, foreign_key: true
+      t.string :type
+      t.string :url
```
farhatahmad, line 8: I would add `null: false`

db/migrate/20220411194428_rename_type_to_recording_type.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class RenameTypeToRecordingType < ActiveRecord::Migration[7.0]
+  def change
+    rename_column :formats, :type, :recording_type
+  end
+end
```
farhatahmad, line 7: I would drop this - you're creating the table from scratch, so you should just do it right the first time

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+      # GET /api/v1/recordings.json
+      # Returns: { data: Array[serializable objects(recordings)] , errors: Array[String] }
+      # Does: Returns all the Recordings from all the current user's rooms
+      def index
+        # Gets the id's of the rooms that belong to the current user
+        room_ids = Room.where(user_id: current_user&.id).map(&:id)
+        # Gets the recordings that belong to current user
+        recordings = Recording.where(room_id: room_ids)
+
+        render json: {
```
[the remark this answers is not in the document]
hadicheaito1: This one has an :includes as well, would need to change the render_json function a bit to make this work. Not sure if it should be part of this PR or another one when I clean things up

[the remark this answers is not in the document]
farhatahmad: You can extend it as part of this PR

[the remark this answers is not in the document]
hadicheaito1: Any ideas on what default value would be best for includes: param in render_json function?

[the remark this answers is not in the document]
farhatahmad: Does nil work?

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+  module V1
+    class RecordingsController < ApplicationController
+      # GET /api/v1/recordings.json
+      # Returns: { data: Array[serializable objects(recordings)] , errors: Array[String] }
+      # Does: Returns all the Recordings from all the current user's rooms
+      def index
+        # Gets the id's of the rooms that belong to the current user
+        room_ids = Room.where(user_id: current_user&.id).map(&:id)
+        # Gets the recordings that belong to current user
+        recordings = Recording.where(room_id: room_ids)
```
farhatahmad, line 13: If you add a `has_many :recordings through: :rooms` to `User`, you can just do `current_user.recordings` here

db/migrate/20220412180711_add_length_visibility_users_to_recordings.rb
```diff
@@ -0,0 +1,7 @@
+class AddLengthVisibilityUsersToRecordings < ActiveRecord::Migration[7.0]
+  def change
+    add_column :recordings, :visibility, :string
+    add_column :recordings, :length, :integer
+    add_column :recordings, :users, :integer
```
farhatahmad, line 5: This needs validation as well

app/models/format.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class Format < ApplicationRecord
+  belongs_to :recording
+end
```
farhatahmad, line 5: Dont forget to add the validation here and create a spec file

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+      # GET /api/v1/recordings.json
+      # Returns: { data: Array[serializable objects(recordings)] , errors: Array[String] }
+      # Does: Returns all the Recordings from all the current user's rooms
+      def index
+        recordings = current_user.recordings
+
+        render json: {
+          data: recordings,
+          errors: []
+        }, status: :ok, include: :formats
```
farhatahmad, line 15: You should use `render_json` here

spec/controllers/recordings_controller_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  let(:user) { create(:user) }
+
+  before do
+    request.headers['ACCEPT'] = 'application/json'
+    session[:user_id] = user.id
+  end
+
+  # TODO: Hadi - Make test work with current_user
+  describe 'index' do
+    it 'ids of recordings in response are matching recording ids that belong to current_user rooms' do
```
farhatahmad, line 15: The name of this doesnt doesnt really make sense. "it ids of recordings..."

Should instead be something like `it 'returns recordings ids that belong to current_user'`

spec/controllers/recordings_controller_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      recordings = create_list(:recording, 6)
+      create_list(:room, 5, user:, recordings:)
+      get :index
+      expect(response).to have_http_status(:ok)
+      response_recording_ids = JSON.parse(response.body)['data'].map { |recording| recording['id'] }
+      expect(response_recording_ids).to eq(recordings.pluck(:id))
+    end
+
+    it 'no recordings for any of the current_user rooms should return empty list' do
```
farhatahmad, line 24: Same as above

spec/models/format_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+require 'rails_helper'
+
+RSpec.describe Format, type: :model do
+  describe 'validations' do
+    subject { create(:format) }
+
+    it { is_expected.to belong_to(:recording) }
+    it { is_expected.to validate_presence_of(:recording_type) }
+    it { is_expected.to validate_presence_of(:url) }
```
farhatahmad, line 11: Don't forget to add these for the recording_spec.rb too

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+module Api
+  module V1
+    class RecordingsController < ApplicationController
+      # GET /api/v1/recordings.json
+      # Returns: { data: Array[serializable objects(recordings)] , errors: Array[String] }
+      # Does: Returns all the Recordings from all the current user's rooms
+      def index
+        recordings = current_user.recordings
+
+        render json: { data: recordings, status: :ok, include: :formats }
```
farhatahmad, line 12: This is still using `render json` - you should be using the `render_json` method instead

app/controllers/application_controller.rb
```diff
@@ -35,11 +35,11 @@ def current_user
     @current_user ||= User.find_by(id: session[:user_id])
   end
 
-  def render_json(data: {}, errors: [], status: :ok)
+  def render_json(data: {}, errors: [], status: :ok, include: nil)
```
farhatahmad, line 38: This is the method you should be using

farhatahmad (approved):

PR #3387, opened 2022-04-12
Backend: Added create room capability for authenticated users.

KH-Amir-TN:
## Description

Added create room capability for authenticated users.
## Testing Steps

1. Pull the code.

### DIFF
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -23,7 +23,6 @@ module Api
       # POST /api/v1/rooms/:friendly_id/start.json
       # Returns: { data: Array[serializable objects] , errors: Array[String] }
       # Does: Starts the Room meeting and joins in the meeting starter.
-
       def start
         # TODO: amir - Check the legitimately of the action.
         bbb_api = BigBlueButtonApi.new
@@ -44,10 +43,24 @@ module Api
         end
       end
 
+      # POST /api/v1/rooms.json
+      # Returns: { data: Array[serializable objects] , errors: Array[String] }
+      # Does: creates a room for the authenticated user.
+      def create
+        # TODO: amir - ensure accessibility for unauthenticated requests only.
+        room = Room.create!(room_params.merge(user_id: current_user.id))
+        logger.info "room(friendly_id):#{room.friendly_id} created for user(id):#{current_user.id}"
+        render_json status: :created
+      end
+
       private
 
       def find_room
         @room = Room.find_by!(friendly_id: params[:friendly_id])
+      end
+
+      def room_params
+        params.require(:room).permit(:name)
       end
     end
   end
--- a/spec/controllers/rooms_controller_spec.rb
+++ b/spec/controllers/rooms_controller_spec.rb
@@ -11,7 +11,7 @@ RSpec.describe Api::V1::RoomsController, type: :controller do
   end
 
   # TODO: Hadi - Make test work with current_user
-  describe 'index' do
+  describe '#index' do
     it 'ids of rooms in response are matching room ids that belong to current_user' do
       rooms = create_list(:room, 5, user:)
       get :index
@@ -36,7 +36,7 @@ RSpec.describe Api::V1::RoomsController, type: :controller do
       expect(JSON.parse(response.body)['data']['id']).to eq(room.id)
     end
 
-    it 'returns not_found if the room doesnt exist' do
+    it 'returns :not_found if the room doesnt exist' do
       get :show, params: { friendly_id: 'invalid_friendly_id' }
       expect(response).to have_http_status(:not_found)
       expect(JSON.parse(response.body)['data']).to be_empty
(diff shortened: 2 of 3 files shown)


app/controllers/api/v1/rooms_controller.rb
```diff
@@ -5,6 +5,7 @@ module V1
     class RoomsController < ApplicationController
       skip_before_action :verify_authenticity_token # TODO: amir - Revisit this.
       before_action :find_room, only: %i[show start]
+      before_action :ensure_authenticated, only: :create
```
farhatahmad, line 8: I would leave this out for now - we should have a separate PR implementing this correctly across all controllers

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -44,11 +44,24 @@ def start
         end
       end
 
+      # POST /api/v1/rooms.json
+      # Returns: { data: Array[serializable objects] , errors: Array[String] }
+      # Does: creates a room for the authenticated user.
+      def create
+        room = Room.new(room_params.merge(user_id: current_user.id)) and room.save!
```
farhatahmad, line 51: This syntax is weird. Remember that keeping things simple makes life much easier for us

```
Room.create!(room_params.merge(user_id: current_user.id))
```

app/controllers/application_controller.rb
```diff
@@ -42,6 +42,10 @@ def render_json(data: {}, errors: [], status: :ok)
     }, status:
   end
 
+  def ensure_authenticated
+    render_json status: :unauthorized and return unless current_user
```
farhatahmad, line 46: Lets leave this for another PR (and avoid the `and` unless necessary)

spec/controllers/rooms_controller_spec.rb
```diff
@@ -59,9 +59,29 @@
...
+  describe '#create' do
+    let(:room_params) do
+      {
+        room: { name: Faker::Science.science }
+      }
+    end
+
+    it 'creates a room for the authenticated user' do
+      session[:user_id] = user.id
+      expect { post :create, params: room_params }.to change { user.reload.rooms.count }.from(0).to(1)
```
farhatahmad, line 77: Is the `reload` necessary here?

app/controllers/api/v1/rooms_controller.rb
KH-Amir-TN: You got reason, 
It's funny though that I used it at first but don't why I've changed it.

app/controllers/application_controller.rb
KH-Amir-TN: Sure,
Does the `render` method kills the request/response chain, if so then the `and return` is absurd unlike with the `redirect_to`.

spec/controllers/rooms_controller_spec.rb
KH-Amir-TN: After checking, It appears that it is not required.
The `count` method seems to be making its own `SQL query`.

app/controllers/api/v1/rooms_controller.rb
KH-Amir-TN: You want me to remove it?

app/controllers/application_controller.rb
KH-Amir-TN: Render doesn't stop the request/response chain, to avoid a full `ìf...else....end` block while using the one liner syntax using the `and return` to kill the chain is recommended.
https://guides.rubyonrails.org/layouts_and_rendering.html#avoiding-double-render-errors

farhatahmad:
```
def ensure_authenticated
  return unless current_user
  return render_json status: :unauthorized  
end
```

app/controllers/api/v1/rooms_controller.rb
farhatahmad, line 8: Don't forget to remove this

KH-Amir-TN: Do you want me to call `#ensure_authenticated` manually in the `#create`?

KH-Amir-TN: That will make it:
`
def create
  return ensure_authenticated
  # Rest of the code...
end
`

farhatahmad: Nono let's drop the ensure authenticated completely from this. 

None of our other actions are secured yet, so it's better if we do them one shot

KH-Amir-TN: Alright, I'll remove the check entirely.

spec/controllers/rooms_controller_spec.rb
```diff
@@ -59,9 +59,29 @@
...
+      }
+    end
+
+    it 'creates a room for the authenticated user' do
+      session[:user_id] = user.id
+      expect { post :create, params: room_params }.to change { user.rooms.count }.from(0).to(1)
+      expect(response).to have_http_status(:created)
+    end
+
+    it 'returns :unauthorized for unauthenticated create requests' do
```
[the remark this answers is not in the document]
KH-Amir-TN: Yup, I'll remove it.
Thank you!

hadicheaito1 (approved):

farhatahmad (approved):

PR #3394, opened 2022-04-14
Create room feature

KH-Amir-TN:
TODO:
    - Add the capability to define the room name.

## Description

This PR enables creating a room for an authenticated user through GL-v3 app.
The authenticated user access the rooms area.
In the rooms area the user click the `+ new room` button and a new room will get created with a name of `Room` and a prefix of the timestamp of the creation event on client side.
**Note**: Made some minor changes to the RoomCard component styling to fix the card size in an expected way.
## Testing Steps

> 1. Pull the code.
> 4. Clean the database and tmp files for a better isolation by running `rails tmp:clear && rails db:schema:cache:clear && rails db:migrate:reset`
> 5. Run the linter and specs `bundle exec rubocop --parallel && bundle exec rspec`

1. Login with a pre created account.
2. Access the `/rooms` page.
3. Click on the `+ New Room` button
4. Confirm that:
  4.1  A new room card gets created as the last one in the rooms list.
  4.2 The new room card will be a placeholder card having:
     4.2.1 The same size as the other room cards.
     4.2.2 The same consistent format (the button , the horizontal rule, the icons, the text placeholders actual placement ).
     4.2.3 No effect once clicked or hovered over.
     4.2.4 Having no effect if its `start` button gets clicked.
  4.3 The `+ New Room` button gets disabled and a spinner with growing animation gets displayed inside it.
  4.4 Once the `+ New Room` animation stops, the spinner disappears and the button becomes enabled (clickable):
     4.4.1 The new room placeholder becomes a valid room card with the room's data matching those in the db.
     4.4.2 The new room card will be clickable and have highlighting effects once hovered over.
     4.4.3 Clicking on the new room card will have the expected outcome as any other room card. 
 -> Any error that pops during the process should revert the rooms list that the page had before committing the action of creating the new room along with a full re-fetch to the cache (rooms list) after reverting.
-> To better visualize the animations kindly consider throttling the network speed in the Dev tools.

image

image

image

image

image


farhatahmad:
Needs to be rebased please

KH-Amir-TN:
> Needs to be rebased please

Rebased on #c1953323

scouillard (dismissed):

app/javascript/components/rooms/RoomCard.jsx
```diff
@@ -15,15 +15,15 @@ export default function RoomCard(props) {
   const { handleStartMeeting, isLoading: startMeetingIsLoading } = useStartMeeting(friendlyId);
 
   return (
-    <Card className="rooms-card" border="light">
-      <Card.Body className="room-card-top pb-0" onClick={handleClick}>
+    <Card className="rooms-card" style={{ height: '19rem', overflow: 'hidden' }} border="light">
```
farhatahmad, line 18: I'm not sure inline styles is the way to go here. I'd rather some other pattern

app/javascript/components/rooms/RoomsList.jsx
```diff
@@ -1,14 +1,18 @@
...
 import useRooms from '../../hooks/queries/rooms/useRooms';
-import ButtonLink from '../shared/stylings/buttons/ButtonLink';
+import useCreateRoom from '../../hooks/mutations/rooms/useCreateRoom';
+import RoomPlaceHolder from './RoomPlaceHolder';
 
 export default function RoomsList() {
   const { isLoading, data: rooms } = useRooms();
-  if (isLoading) return <Spinner />;
+  const newRoomData = { name: `Room ${Date.now()}` }; // TODO: amir - change this.
+  const { handleCreateRoom, isLoading: createRoomIsLoading } = useCreateRoom(newRoomData);
```
farhatahmad, line 14: I don't think renaming `isLoading` is necessary here

app/javascript/components/rooms/RoomPlaceHolder.jsx
```diff
@@ -0,0 +1,29 @@
...
+        <Placeholder as={Card.Title} animation="glow">
+          <Placeholder xs={12} size="sm" />
+        </Placeholder>
+        <Placeholder as={Card.Text} animation="glow">
+          <Placeholder xs={7} size="xs" /> <Placeholder xs={4} size="xs" /> <Placeholder xs={3} size="xs" />
+        </Placeholder>
+      </Card.Body>
+      <Card.Body style={{ maxHeight: '7rem', overflow: 'hidden' }} className="pt-0">
+        <Placeholder as={Card.Text} animation="glow">
+          <Placeholder xs={3} size="xs" /> <Placeholder xs={2} size="xs" />
```
farhatahmad, line 21: Is placeholder required here? Is there a way we can do this across all fields in the app without manually specifying each one?

app/javascript/components/rooms/RoomsList.jsx
KH-Amir-TN: I noticed that we'll have several different `isLoading` states and thought about making renaming them a practice even when it's not necessary.

app/javascript/components/rooms/RoomCard.jsx
KH-Amir-TN: Sure, if you approve the changes it's ready to get changed into a seperate css file.

app/javascript/components/rooms/RoomPlaceHolder.jsx
KH-Amir-TN: I don't think I get  exactly what you mean?

app/javascript/components/rooms/RoomsList.jsx
farhatahmad: Hmm thats a good point - in that case, I would also rename the `isLoading` in `useRooms`

app/javascript/components/rooms/RoomPlaceHolder.jsx
KH-Amir-TN: Alright, but what should I set in place of the actual room card?
I'm doing an optimistic update and the `RoomPlaceHolder` component was made to replace the actual room until it gets created?
As for making a generic placeholder that fits all components, I don't know how we can do that.
When working on this, the cleanest way that came to my mind is the fallowing where each component will have its mock.

farhatahmad (approved):

PR #3395, opened 2022-04-14
Backend: Add User Update Name Form

scouillard:
## Description

Add Form to allow User to change it's own name.

## Testing Steps

Navigate to the user profile via /user or via the header dropdown.
Choose a new name and click on Update.

image

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -21,6 +21,15 @@ module Api
         end
       end
 
+      def update
+        user = User.find(params[:id])
+        if user.update(user_params)
+          render_json
+        else
+          render_json errors: user.errors.to_a, status: :bad_request
+        end
+      end
+
       private
 
       def user_params
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -24,7 +24,7 @@ module Api
       def update
         user = User.find(params[:id])
         if user.update(user_params)
-          render_json
+          render_json status: :ok
         else
           render_json errors: user.errors.to_a, status: :bad_request
         end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -8,5 +8,5 @@ class User < ApplicationRecord
   # TODO: amir - Validate email format.
   validates :email, presence: true, uniqueness: { case_sensitive: false, scope: :provider }
   validates :provider, presence: true
-  validates :password_confirmation, presence: true
+  validates :password_confirmation, presence: true, on: :create
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -11,7 +11,7 @@ Rails.application.routes.draw do
           delete 'signout', to: 'sessions#destroy'
         end
       end
-      resources :users, only: [:create]
+      resources :users, only: %i[create update]
       resources :rooms, only: %i[show index create destroy], param: :friendly_id do
         post '/start', to: 'rooms#start', as: :start_meeting, on: :member
       end
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::V1::UsersController, type: :controller do
+  before do
+    request.headers['ACCEPT'] = 'application/json'
+  end
+
+  # TODO - samuel: does this belong in Controller spec? Or Model?
+  describe 'POST users#create' do
+    it 'adds greenlight as provider on creation' do
+      user = create(:user)
+      expect(user).to have_attributes(provider: 'greenlight')
+    end
+  end
+
+  describe 'POST users#update' do
+    it 'allows user to edit his name' do
+      user = create(:user)
+      patch :update, params: { id: user.id, user: { name: 'New Name' } }
+      Rails.logger.debug "Reason of rollback: #{user.errors.full_messages}"
+      expect(response).to have_http_status(:ok)
+      expect(user.name).to eql('New Name')
+    end
+  end
+end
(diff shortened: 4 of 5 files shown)


farhatahmad:
Let's split this into 2 PRs - the backend one should have some tests with it too

app/models/user.rb
```diff
@@ -7,5 +7,5 @@ class User < ApplicationRecord
   # TODO: amir - Validate email format.
   validates :email, presence: true, uniqueness: { case_sensitive: false, scope: :provider }
   validates :provider, presence: true
-  validates :password_confirmation, presence: true
+  validates :password_confirmation, presence: true, on: :create
```
hadicheaito1, line 10: Could you explain a bit why this needed to be added?

KH-Amir-TN: Will we be having a separate page to change the password like with GLv2?

app/controllers/api/v1/users_controller.rb
```diff
@@ -21,6 +21,15 @@ def create
         end
       end
 
+      def update
+        user = User.find(params[:id])
+        if user.update(user_params)
+          render_json status: :ok
```
KH-Amir-TN, line 27: To further simplify this, render_json by default returns render json: { data: [], errors: [] }, status: :ok, include: nil

KH-Amir-TN (approved):

app/models/user.rb
KH-Amir-TN: You got reason,

app/controllers/api/v1/users_controller.rb
farhatahmad: I would still keep the `status: :ok` so that its clear that this is a success

farhatahmad: Also - shouldn't we return the updated user?

farhatahmad: Thats fine for now - i'll look into it later

spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+require 'rails_helper'
+
+RSpec.describe Api::V1::UsersController, type: :controller do
+  before do
+    request.headers['ACCEPT'] = 'application/json'
+  end
+
+  describe 'POST users#update' do
+    it 'allows user to update the attributes' do
```
farhatahmad, line 11: `it 'updates the users attributes'`

spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      user = create(:user)
+      patch :update, params: { id: user.id, user: { name: 'New Name', email: 'newemail@email.com' } }
```
farhatahmad, line 13: I would create a var at the top

```
updated_params = {
  name: "New Name",
  email: "newemail@gmail.com"
}
```

and then use that in the `eq` below

spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      user.reload
+      expect(response).to have_http_status(:ok)
+      expect(user.name).to eql('New Name')
```
farhatahmad, line 16: Let's use `eq` instead of `eql` for consistency

spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      expect(user.email).to eql('newemail@email.com')
+    end
+
+    it 'does an error when user update with invalid name' do
```
farhatahmad, line 20: `it 'returns an error if the user update fails'`

spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      user = create(:user)
+      patch :update, params: { id: user.id, user: { name: nil } }
+      user.reload
+      expect(response).to have_http_status(:bad_request)
```
farhatahmad, line 24: I'd also add a test to make sure the name is still the same value

spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+
+    it 'does an error when user update with invalid email' do
+      user = create(:user)
+      patch :update, params: { id: user.id, user: { email: nil } }
+      user.reload
+      expect(response).to have_http_status(:bad_request)
+    end
```
farhatahmad, line 32: No need for this test. You only need to test one case where it fails

config/routes.rb
```diff
@@ -11,7 +11,7 @@
           delete 'signout', to: 'sessions#destroy'
         end
       end
-      resources :users, only: [:create]
+      resources :users, only: %i[create update destroy]
```
farhatahmad, line 14: Why `destroy`?

scouillard: It shouldn't be there.. I am working on the Delete User feature in parallel.

spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::V1::UsersController, type: :controller do
+  updated_params = {
+    name: 'New Name',
+    email: 'newemail@gmail.com'
+  }
```
farhatahmad, line 9: These aren't needed in all of the tests, so you should only put them in the block thats needed

spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    request.headers['ACCEPT'] = 'application/json'
+  end
+
+  describe 'POST users#update' do
+    it 'updates the users attributes' do
+      user = create(:user)
+      patch :update, params: { id: user.id, user: updated_params }
+      user.reload
+      expect(response).to have_http_status(:ok)
+      expect(user.name).to eq('New Name')
```
farhatahmad, line 21: expect(user.name).to eq(updated_params.name)

spec/controllers/users_controller_spec.rb
farhatahmad: ^ Still needs to be done

scouillard: NVM, I will just add line 34 to the test above, as you suggested.

farhatahmad (approved):

PR #3397, opened 2022-04-14
Cleaned up a bunch of styling and fixed the nav tabs and background

farhatahmad:
## Description

Still needs some touch ups, but that will be done in a later PR (once more components are on the page)

image


scouillard:
Should we make /signin (the Sign In component) as Root?

app/javascript/App.jsx
```diff
@@ -2,11 +2,14 @@ import React from 'react';
...
 import Header from './components/shared/Header';
+import { useAuth } from './contexts/auth/AuthProvider';
 
 export default function App() {
+  const currentUser = useAuth();
+
   return (
     <>
-      <Header />
+      {currentUser?.signed_in && <Header /> }
```
KH-Amir-TN, line 12: That will introduce the fallowing bug:
>Users won't be able to login since they aren't authenticated, they won't see the HEADER

app/assets/stylesheets/application.bootstrap.scss
```diff
@@ -1,10 +1,16 @@
...
+@import 'variables';
 @import 'bootstrap/scss/bootstrap';
 
 @import 'helpers';
 @import 'rooms';
 
+html,
 body {
-  background-color: whitesmoke;
+  height: 100%;
```
KH-Amir-TN, line 9: What is the use of `height: 100%` here?

KH-Amir-TN (changes requested):

app/javascript/App.jsx
hadicheaito1: True

farhatahmad: Yeah we'll need to them to the index for now. I'll add it to this pr

app/assets/stylesheets/application.bootstrap.scss
farhatahmad: Make the body take the full height of the browser, so that we can later on make the background take the full height as well

KH-Amir-TN: Alright, thanks thought that by default the body occupies 100% of the page.

app/javascript/App.jsx
farhatahmad: Added a new homepage component with the 2 buttons. Once we get the mockup for the home page, we can make it look proper

hadicheaito1 (dismissed):

hadicheaito1 (approved):

hadicheaito1 (approved):

PR #3401, opened 2022-04-19
Backend: Designed the models for room settings and meeting options.

KH-Amir-TN:
## Description

- Created `meeting_option` model which will hold all the the GL room meeting options.
- Created `room_meeting_option` join model for the `room` and `meeting_option` model.
~- Added some attributes to the `room` table.~
- Created the required migrations and tests.
## Testing Steps

> 1. Pull the code.
> 4. Clean the database and tmp files for a better isolation by running `rails tmp:clear && rails db:schema:cache:clear && rails db:migrate:reset`
> 5. Run the linter and specs `bundle exec rubocop --parallel && bundle exec rspec`

image

### DIFF
--- a/app/models/meeting_option.rb
+++ b/app/models/meeting_option.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class MeetingOption < ApplicationRecord
+  has_many :room_meeting_options, dependent: :restrict_with_exception
+
+  validates :name, presence: true, uniqueness: true
+  validates :default_value, presence: true
+end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -3,6 +3,7 @@
 class Room < ApplicationRecord
   belongs_to :user
   has_many :recordings, dependent: :destroy
+  has_many :room_meeting_options, dependent: :destroy
 
   validates :name, presence: true
   validates :friendly_id, presence: true, uniqueness: true
--- a/db/migrate/20220418171035_create_meeting_options.rb
+++ b/db/migrate/20220418171035_create_meeting_options.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateMeetingOptions < ActiveRecord::Migration[7.0]
+  def change
+    create_table :meeting_options do |t|
+      t.string :name, index: { unique: true }
+      t.string :default_value
+
+      t.timestamps
+    end
+  end
+end
--- a/app/models/room_meeting_option.rb
+++ b/app/models/room_meeting_option.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class RoomMeetingOption < ApplicationRecord
+  belongs_to :room
+  belongs_to :meeting_option
+
+  delegate :name, to: :meeting_option
+end
--- a/db/migrate/20220418171152_create_room_meeting_options.rb
+++ b/db/migrate/20220418171152_create_room_meeting_options.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateRoomMeetingOptions < ActiveRecord::Migration[7.0]
+  def change
+    create_table :room_meeting_options do |t|
+      t.belongs_to :room, foreign_key: true
+      t.belongs_to :meeting_option, foreign_key: true
+
+      t.string :value
+      t.timestamps
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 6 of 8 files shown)


KH-Amir-TN:
Rebased on 87196bd4e441140a27d439ce977254c78ac2b36b

scouillard:
Why does room_meeting_options has a "value" column?
Isn't it strictly for linking Rooms and Meeting Options tables?

scouillard:
I suppose the plan is to store the Meeting settings as strings in the value column?
We could also make a boolean column for every setting in Meeting Options?
I think that is how I would have made it.
Would you please explain why you choose to store the settings as strings?

KH-Amir-TN:
> Why does room_meeting_options has a "value" column? Isn't it strictly for linking Rooms and Meeting Options tables?

A **many-to-many** relation results in the creation of a new entity/table.
 https://www.tutorialspoint.com/dbms/er_diagram_representation.htm
In ER (Entity Relationship) modeling an instance of the new entity(table) holds any values related to the association along with bounds to the different dimensions (entities).
See how in the example the relationship **OPOS** have a attribute **QTY** resulting from the association of many **customers** to many **articles**:
image

So each **room** has many **meeting options** and each registered **meeting option** has many **rooms** which results in forming the entity `room_meeting_options` that links a **room** to a **meeting option**.
So the actual meeting option for a specific room is the `room_meeting_option` record not the `meeting_option `and therefore each meeting option that a room selects needs its value defined in the association level (it's created only because of the association).
The **meeting option** entity is a place to store all **meeting options** (muteOnStart, joinOnStart, ...) that any **room** can have and the _value_ attribute in that level corresponds to the _default value_ of the option where the actual value for that option for a specific **room** resides in the `room_meeting_option` record for that **meeting option** and that **room**.

scouillard:
Why do we need a value column if we already have the meeting_option_id key in the joining table?

Isn't the joining table used to match the values?

image

KH-Amir-TN:
> Why do we need a value column if we already have the meeting_option_id key in the joining table?
> 
> Isn't the joining table used to match the values?
> 
> Below, we can see that Room 1 (My Awesome Room) has the Meeting Option 1 (muteOnStart) and Meeting Option 2 (joinOnStart) without the use of the value column.
> 
> Am I wrong?
> 
> Is the value column just a duplication of the name column?
> 
> image
Hmm, I see where you're coming from.
This means that all rooms will share the same meeting options (settings) and there's no way to keep each room config scoped only to that room.
The meeting option can be thought of as a registry where all possible options are saved, in fact it will be populated only once by the app itself.
The actual room meeting options are those that links the room to the meeting option in the registry.
The meeting option `value` attribute will hold only the default value because I may not create `room_meeting_option` record for those who share the same value as the `meeting_option` in the registry.
Do you find calling the rename the `value -> default_value` attribute in the registry makes more sense?

app/models/room.rb
```diff
@@ -3,6 +3,7 @@
 class Room < ApplicationRecord
   belongs_to :user
   has_many :recordings, dependent: :destroy
+  has_many :room_meeting_options, dependent: :destroy
```
farhatahmad, line 6: I think we should add this `has_many :meeting_options, through: :room_meeting_options`

app/models/meeting_option.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class MeetingOption < ApplicationRecord
+  has_many :room_meeting_options, dependent: :restrict_with_exception
```
farhatahmad, line 4: I think we should add this  `has_many :rooms, through: :room_meeting_options`

db/migrate/20220418171035_create_meeting_options.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateMeetingOptions < ActiveRecord::Migration[7.0]
+  def change
+    create_table :meeting_options do |t|
+      t.string :name
```
farhatahmad, line 6: `t.string :name, index: { unique: true }`

app/models/meeting_option.rb
farhatahmad: What's your thinking behind `:restrict_with_exception`

KH-Amir-TN: Removing a meeting option when there's rooms that depends on it is crucial.
There was no meaning behind destroying or deleting any dependent record once we remove a meeting option but at least I wanted to prevent that from happening by restricting the operation and that also silences Rubocop xD.

app/models/room.rb
KH-Amir-TN: I think this not still needed?

farhatahmad (approved):

PR #3404, opened 2022-04-19
Backend: Add Delete User Form

scouillard:
## Description

Controller method and Specs for the Delete User form.


scouillard:
Depends on #3405.

app/controllers/api/v1/users_controller.rb
```diff
@@ -21,6 +21,16 @@ def create
         end
       end
 
+      def destroy
+        user = User.find(params[:id])
+        if user.destroy
+          # TODO: samuel - Need to destroy all adjacent Rooms and Recordings as well
```
farhatahmad, line 27: This is already done through the `dependent: :destroy`

spec/controllers/users_controller_specs.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  describe 'DELETE users#destroy' do
+    it 'deletes the user' do
+      user = create(:user)
+      delete :destroy, params: { id: user.id }
+      expect(response).to have_http_status(:ok)
+    end
+
+    it 'returns an error if the user id is invalid' do
+      delete :destroy, params: { id: 0 }
```
farhatahmad, line 18: 0 could be a valid id at some point. I would change it to nil, or something like `id: "invalid-id"`

farhatahmad:
You'll need to rebase and resolve the conflicts

spec/controllers/users_controller_spec.rb
```diff
@@ -28,4 +28,18 @@
       expect(user.reload.name).to eq(user.name)
     end
   end
+
+  describe 'DELETE users#destroy' do
+    it 'deletes the user' do
+      user = create(:user)
+      delete :destroy, params: { id: user.id }
+      expect(response).to have_http_status(:ok)
+      expect { user.reload }.to raise_error(ActiveRecord::RecordNotFound)
```
farhatahmad, line 37: This works but isn't quite the approach we want to take. Take a look at some of the other tests to see how we do it

scouillard: Agreed - I'm sorry, I should have checked before.

spec/controllers/users_controller_spec.rb
```diff
@@ -28,4 +28,17 @@
...
+  describe 'DELETE users#destroy' do
+    it 'deletes the user' do
+      user = create(:user)
+      expect(response).to have_http_status(:ok)
+      expect { delete :destroy, params: { id: user.id } }.to change(User, :count).by(-1)
+    end
+
+    it 'returns an error if the user id is invalid' do
+      delete :destroy, params: { id: 'invalid-id' }
+      expect(response).to have_http_status(:not_found)
```
farhatahmad, line 41: This still doesn't actually check that something was deleted - just checks that a :not_found was returned.

We need to actually check that a record in the database was deleted

scouillard: The test at line 33 does that.
Not sure test at line 39 is any good - please let me know.

farhatahmad: Sorry meant to say **not** deleted. 

This should do exactly as 36, but instead check that it wasn't changed

PR #3396, opened 2022-04-14
Frontend: Add User Update Name Form

scouillard:
## Description

Add Form to allow User to change it's own name.

## Testing Steps

Navigate to the user profile via /user or via the header dropdown.
Choose a new name and click on Update.

## Screenshots (if appropriate):

image

## TODO
Add the Snapshot tests.

PR #3405, opened 2022-04-19
Frontend: Add Delete User Form

scouillard:
## Description

User can now delete his account.
Created a "Delete Confirmation" Modal that can be re-used.

## Testing Steps

Unsure how to test as this PR is split into frontend and backend.

image

### DIFF
--- a/app/javascript/components/forms/DeleteUserForm.jsx
+++ b/app/javascript/components/forms/DeleteUserForm.jsx
@@ -1,0 +1,22 @@
+import React from 'react';
+import { useForm } from 'react-hook-form';
+import {
+  Button,
+} from 'react-bootstrap';
+import Form from './Form';
+import useDeleteUser from '../../hooks/mutations/users/useDeleteUser';
+import { useAuth } from '../../contexts/auth/AuthProvider';
+
+export default function DeleteUserForm() {
+  const currentUser = useAuth();
+  const methods = useForm();
+  const { onSubmit } = useDeleteUser(currentUser?.id);
+
+  return (
+    <Form methods={methods} onSubmit={onSubmit}>
+      <Button variant="danger" type="submit">
+        Delete
+      </Button>
+    </Form>
+  );
+}
--- a/app/javascript/components/forms/DeleteUserForm.jsx
+++ b/app/javascript/components/forms/DeleteUserForm.jsx
@@ -6,16 +6,19 @@ import {
 import Form from './Form';
 import useDeleteUser from '../../hooks/mutations/users/useDeleteUser';
 import { useAuth } from '../../contexts/auth/AuthProvider';
+import Spinner from '../shared/stylings/Spinner';
 
 export default function DeleteUserForm() {
   const currentUser = useAuth();
   const methods = useForm();
+  const { isSubmitting } = methods.formState;
   const { onSubmit } = useDeleteUser(currentUser?.id);
 
   return (
     <Form methods={methods} onSubmit={onSubmit}>
-      <Button variant="danger" type="submit">
+      <Button variant="danger" type="submit" disabled={isSubmitting}>
         Delete
+        { isSubmitting && <Spinner /> }
       </Button>
     </Form>
   );
--- a/app/javascript/components/forms/DeleteUserForm.jsx
+++ b/app/javascript/components/forms/DeleteUserForm.jsx
@@ -1,14 +1,15 @@
 import React from 'react';
 import { useForm } from 'react-hook-form';
 import {
-  Button,
+  Button, Stack,
 } from 'react-bootstrap';
+import PropTypes from 'prop-types';
 import Form from './Form';
 import useDeleteUser from '../../hooks/mutations/users/useDeleteUser';
 import { useAuth } from '../../contexts/auth/AuthProvider';
 import Spinner from '../shared/stylings/Spinner';
 
-export default function DeleteUserForm() {
+export default function DeleteUserForm({ handleClose }) {
   const currentUser = useAuth();
   const methods = useForm();
   const { isSubmitting } = methods.formState;
@@ -16,10 +17,19 @@ export default function DeleteUserForm() {
 
   return (
     <Form methods={methods} onSubmit={onSubmit}>
-      <Button variant="danger" type="submit" disabled={isSubmitting}>
-        Delete
-        { isSubmitting && <Spinner /> }
-      </Button>
+      <Stack direction="horizontal" gap={1}>
+        <Button variant="secondary" onClick={handleClose}>
+          Close
+        </Button>
+        <Button variant="danger" type="submit" disabled={isSubmitting}>
+          Delete
+          { isSubmitting && <Spinner /> }
+        </Button>
+      </Stack>
     </Form>
   );
 }
+
+DeleteUserForm.propTypes = {
+  handleClose: PropTypes.func.isRequired,
+};
--- a/app/javascript/components/forms/DeleteUserForm.jsx
+++ b/app/javascript/components/forms/DeleteUserForm.jsx
@@ -31,5 +31,9 @@ export default function DeleteUserForm({ handleClose }) {
 }
 
 DeleteUserForm.propTypes = {
-  handleClose: PropTypes.func.isRequired,
+  handleClose: PropTypes.func,
+};
+
+DeleteUserForm.defaultProps = {
+  handleClose: () => {},
 };
--- a/app/javascript/components/shared/stylings/DeleteModal.jsx
+++ b/app/javascript/components/shared/stylings/DeleteModal.jsx
@@ -1,0 +1,36 @@
+import React, { useState } from 'react';
+import { Modal, Button } from 'react-bootstrap';
+import PropTypes from 'prop-types';
+
+export default function DeleteModal({ title, body, children }) {
+  const [show, setShow] = useState(false);
+
+  const handleClose = () => setShow(false);
+  const handleShow = () => setShow(true);
+
+  return (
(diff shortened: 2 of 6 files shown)


scouillard:
Depends on #3395 and #3396.

app/javascript/components/shared/stylings/DeleteModal.jsx
```diff
@@ -0,0 +1,36 @@
...
+  const handleClose = () => setShow(false);
+  const handleShow = () => setShow(true);
+
+  return (
+    <>
+      <Button variant="danger" onClick={handleShow}>
+        Delete
+      </Button>
+
+      <Modal className="text-center" show={show} onHide={handleClose}>
```
KH-Amir-TN, line 17: I think it would be better if you create a generic `Modal` (Header, Title, Body,Footer, states, ...) component that can be used afterwards in the application and use it in the `DeleteModal` adding the elements necessary for the deletion context like the **Delete** button etc...
Also it may be better if the children prop is rendered in the body section and and not in the footer and make the footer itself a prop `<Modal title={title} footer={footer}> {body} </Modal>`.

app/javascript/components/forms/DeleteUserForm.jsx
```diff
@@ -0,0 +1,22 @@
...
+import {
+  Button,
+} from 'react-bootstrap';
+import Form from './Form';
+import useDeleteUser from '../../hooks/mutations/users/useDeleteUser';
+import { useAuth } from '../../contexts/auth/AuthProvider';
+
+export default function DeleteUserForm() {
+  const currentUser = useAuth();
+  const methods = useForm();
```
KH-Amir-TN, line 12: It's recommended always to provide default values to `react-hook-form`.
Rules section of https://react-hook-form.com/api/useform

app/javascript/components/forms/DeleteUserForm.jsx
```diff
@@ -0,0 +1,22 @@
...
+  const { onSubmit } = useDeleteUser(currentUser?.id);
+
+  return (
+    <Form methods={methods} onSubmit={onSubmit}>
+      <Button variant="danger" type="submit">
```
KH-Amir-TN, line 17: Since you are using the `mutateAsync` you can leverage the `isSubmitting` react hook form state to disable the delete button avoiding flooding the server and show maybe proper animations like a spinner or anything relevant to that context.

app/javascript/hooks/mutations/users/useDeleteUser.jsx
```diff
@@ -0,0 +1,23 @@
+import { useMutation, useQueryClient } from 'react-query';
+import axios from 'axios';
```
KH-Amir-TN, line 2: You can use the custom axios and the `ENDPOINTS` constant instance to further simplify this.

```
import axios, { ENDPOINTS } from '../../../helpers/Axios';
```

app/javascript/components/forms/DeleteUserForm.jsx
scouillard: It has no values for now, the form is only a button.
Will probably have to add a password confirmation field or a checkbox to confirm the deletion eventually.

app/javascript/hooks/mutations/users/useDeleteUser.jsx
scouillard: I thought we settled on not using the ENDPOINTS for now

app/javascript/components/shared/stylings/DeleteModal.jsx
scouillard: I agree for a generic Modal - I will work on this right now.
I don't understand what you mean by _making the footer a prop._
Maybe we should wait for mock ups before really digging into it.

app/javascript/components/forms/DeleteUserForm.jsx
KH-Amir-TN: My bad, I mixed between this form and the update profile one 🤣

app/javascript/hooks/mutations/users/useDeleteUser.jsx
KH-Amir-TN: We haven't discussed its use for now.
I'm still using it and I think Hadi used it in his last PR.
We should discuss it later on to keep consistency.

app/javascript/components/shared/stylings/DeleteModal.jsx
scouillard: I renamed the DeleteModal component to Modal.
The logic for putting the children (the Form) in the Modal Footer was that I (assumed incorrectly) that most of the Modal Forms would be a single button. 
But after some thinking, we should wait for more mock ups before spending time on customising the Modal.

app/javascript/components/shared/stylings/Modal.jsx
```diff
@@ -0,0 +1,36 @@
...
+    <>
+      <Button variant="danger" onClick={handleShow}>
+        Delete
+      </Button>
+
+      <BootstrapModal className="text-center" show={show} onHide={handleClose}>
+        <BootstrapModal.Header className="border-0" closeButton />
+        <BootstrapModal.Title>{title}</BootstrapModal.Title>
+        <BootstrapModal.Body>{body}</BootstrapModal.Body>
+        <BootstrapModal.Footer className="">
```
hadicheaito1, line 21: className can be removed here since it's empty

app/javascript/components/shared/stylings/Modal.jsx
```diff
@@ -0,0 +1,36 @@
...
+export default function Modal({ title, body, children }) {
+  const [show, setShow] = useState(false);
+
+  const handleClose = () => setShow(false);
+  const handleShow = () => setShow(true);
+
+  return (
+    <>
+      <Button variant="danger" onClick={handleShow}>
+        Delete
```
hadicheaito1, line 14: Should make this a prop too since this is now a general shared Modal

scouillard:
What do you guys think of the clonedChildren to pass method from Parent to Children?
We could use Context as well, but I think it is overkill and too complex for this usecase.
I think clonedChildren is common pratice.
I found it here:
https://stackoverflow.com/questions/32370994/how-to-pass-props-to-this-props-children?answertab=modifieddesc#tab-top

KH-Amir-TN:
> What do you guys think of clonedChildren to pass information from Parent to Children? We have something like (this below) and we would like to share a Parent method to the Children.
> 
> ```
> <Parent>
>   <Children>
> </Parent>
> ```
> 
> We could use Context as well, but I think it is overkill and too complex for this usecase. I think clonedChildren is common pratice.
> 
> I found it here: https://stackoverflow.com/questions/32370994/how-to-pass-props-to-this-props-children?answertab=modifieddesc#tab-top Demo here: https://stackblitz.com/edit/react-te35fc?file=src/App.js
> 
> @farhatahmad @hadicheaito1 @KH-Amir-TN

You can use the clone API it's much lighter weight then the children one.
`  const clonedChildren = React.cloneElement(children,{handleClose}) `

KH-Amir-TN (dismissed):

app/javascript/components/forms/DeleteUserForm.jsx
```diff
@@ -0,0 +1,35 @@
...
+          Delete
+          { isSubmitting && <Spinner /> }
+        </Button>
+      </Stack>
+    </Form>
+  );
+}
+
+DeleteUserForm.propTypes = {
+  handleClose: PropTypes.func.isRequired,
```
KH-Amir-TN, line 34: I think this should change to `handleColse: PropTypes.func` because you are creating the `DeleteUserForm` first without the `handleClose` prop and inject afterwards with the `cloneElement` **API**.
Check this warning the browser console:
```
Warning: Failed prop type: The prop `handleClose` is marked as required in `DeleteUserForm`, but its value is `undefined`.
```

KH-Amir-TN (changes requested):

app/javascript/components/shared/stylings/Modal.jsx
```diff
@@ -0,0 +1,36 @@
...
+      </BootstrapModal>
+    </>
+  );
+}
+
+Modal.propTypes = {
+  title: PropTypes.string.isRequired,
+  footer: PropTypes.element.isRequired,
+  modalButton: PropTypes.element.isRequired,
+  body: PropTypes.element.isRequired,
```
KH-Amir-TN, line 35: Even if it's not raising warning this still incorrect in a way since the `body` and `footer` props can hold `React elements,strings, numbers,...` and not only `React elements`.
Therefore, the adequate `propType` for that is `PropTypes.element -> PropTypes.node`.
https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes

KH-Amir-TN (changes requested):

app/javascript/components/forms/DeleteUserForm.jsx
scouillard: Agreed.
But now it is complaining that it has no default value.
Do you think this is acceptable in this case?
```
DeleteUserForm.defaultProps = {
  handleClose: '',
};
```

KH-Amir-TN: Hmm, setting it to an empty string is wrong it will raise the fallowing error in the browser console:
```
Warning: Failed prop type: Invalid prop `handleClose` of type `string` supplied to `DeleteUserForm`, expected `function`.
```
Try this instead:
```
DeleteUserForm.defaultProps = {
  handleClose: () => { },
};
```

farhatahmad (approved):

PR #3407, opened 2022-04-20
Search rooms by name

hadicheaito1:
## Description

- Can search Rooms by name
- Dynamically appear as you are typing, no need to click enter or anything
- No request is done, simply using the rooms var that is already there and filters through it based on search query
## Testing Steps

- Pull changes
- Create rooms
- Search by name

### DIFF
--- a/app/javascript/components/rooms/RoomsList.jsx
+++ b/app/javascript/components/rooms/RoomsList.jsx
@@ -1,7 +1,5 @@
-import React from 'react';
-import {
-  Row, Col, Button, Form, FormControl,
-} from 'react-bootstrap';
+import React, { useState } from 'react';
+import { Row, Col, Button } from 'react-bootstrap';
 import Spinner from '../shared/stylings/Spinner';
 import RoomCard from './RoomCard';
 import useRooms from '../../hooks/queries/rooms/useRooms';
@@ -12,23 +10,21 @@ export default function RoomsList() {
   const { isLoading, data: rooms } = useRooms();
   const newRoomData = { name: `Room ${Date.now()}` }; // TODO: amir - change this.
   const { handleCreateRoom, isLoading: createRoomIsLoading } = useCreateRoom(newRoomData);
-  if (isLoading) return <Spinner />; // TODO: amir - revisit this.
+  const [search, setSearch] = useState('');
+  if (isLoading) return <Spinner />;
 
   return (
     <div className="wide-background">
       <Row className="pt-4 mb-4">
         <Col>
-          {/* TODO: May need to change search form when implementing search functionality */}
-          <Form className="d-flex">
-            <FormControl
-              type="search"
-              placeholder="Search"
-              className=""
-              style={{ width: '19rem' }}
-              aria-label="Search"
-            />
-            <Button className="ms-2" variant="outline-secondary">Search</Button>
-          </Form>
+          <input
+            id="rooms-search"
+            className="rounded border border-dark"
+            placeholder=" Search Room"
+            type="search"
+            style={{ width: '19rem' }}
+            onChange={(event) => setSearch(event.target.value)}
+          />
         </Col>
         <Col>
           {/* TODO: Set this button to create new room page */}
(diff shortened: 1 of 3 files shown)


app/javascript/components/rooms/RoomsList.jsx
```diff
@@ -1,42 +1,49 @@
...
-              placeholder="Search"
-              className=""
-              style={{ width: '19rem' }}
-              aria-label="Search"
-            />
-            <Button className="ms-2" variant="outline-secondary">Search</Button>
-          </Form>
+          <input
+            id="rooms-search"
+            className="rounded border border-dark"
```
scouillard, line 19: I think you can remove border-dark? It's not on the mock ups.

scouillard: Also, IDK what is causing that, but you can add
```
  &:focus {
    outline: none;
  }
```
  
In #rooms-search CSS to remove the black outline.

app/javascript/components/rooms/RoomsList.jsx
```diff
@@ -1,42 +1,49 @@
...
-        {rooms.map((room) => (
-          <Col key={room.friendly_id}>
-            <RoomCard id={room.friendly_id} name={room.name}> </RoomCard>
-          </Col>
-        ))}
+        {
+          rooms.filter((room) => {
+            if (search === '') {
+              return room;
+            }
```
scouillard, line 36: What is this for? I think it does nothing?

app/javascript/components/rooms/RoomsList.jsx
```diff
@@ -1,42 +1,49 @@
...
-            <FormControl
-              type="search"
-              placeholder="Search"
-              className=""
-              style={{ width: '19rem' }}
-              aria-label="Search"
-            />
-            <Button className="ms-2" variant="outline-secondary">Search</Button>
-          </Form>
+          <input
```
scouillard, line 17: Might as well use the FormControl component instead of input for consistency. 
Should be useful when we add the search icon inside the input field.. which doesn't seem to be supported by React Bootstrap for some reason.

app/javascript/components/rooms/RoomsList.jsx
hadicheaito1: True :laughing:

app/javascript/components/rooms/RoomsList.jsx
```diff
@@ -1,29 +1,33 @@
...
   if (isLoading) return <Spinner />;
 
   return (
     <>
       <Row className="mt-4 mb-4">
         <Col>
-          {/* TODO: May need to change search form when implementing search functionality */}
-          <Form className="d-flex">
+          <Form>
             <FormControl
```
KH-Amir-TN, line 20: I think it's better to use a fully [uncontrolled ](https://reactjs.org/docs/uncontrolled-components.html) form rather then a fully [controlled](https://reactjs.org/docs/forms.html#controlled-components) one.
Also, we're using the [react-hook-form](https://react-hook-form.com/) so it's advised to use it even for such simple form for consistency, maintainability and performance.

app/javascript/components/rooms/RoomsList.jsx
```diff
@@ -32,11 +36,18 @@ export default function RoomsList() {
         </Col>
       </Row>
       <Row md={4} className="g-4">
-        {rooms.map((room) => (
-          <Col key={room.friendly_id}>
-            <RoomCard id={room.friendly_id} name={room.name}> </RoomCard>
-          </Col>
-        ))}
+        {
+          rooms.filter((room) => {
```
KH-Amir-TN, line 40: Hmm, are we going fully client side on this?
At least we can configure the `getRooms` query `interval` maybe to quasi-synchronize the rooms list in the UI with the actual data or consider maybe invalidating the query with each search (which may have some performance implication so at least debounce it).
We may keep it like this for this milestone, but we need to come back to it later.

farhatahmad (approved):

app/javascript/components/rooms/RoomsList.jsx
hadicheaito1: NOTE: We discussed this and decided to leave it as is

farhatahmad (approved):

PR #3409, opened 2022-04-21
Back-end: Room Recordings List

hadicheaito1:
## Description

- Added /rooms/:friendly_id/recordings GET request that returns the recordings that belong to the room with given friendly_id
- Added RSpec testing for rooms#recordings

### DIFF
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -4,7 +4,7 @@ module Api
   module V1
     class RoomsController < ApplicationController
       skip_before_action :verify_authenticity_token # TODO: amir - Revisit this.
-      before_action :find_room, only: %i[show start]
+      before_action :find_room, only: %i[show start recordings]
 
       # GET /api/v1/rooms.json
       # Returns: { data: Array[serializable objects(rooms)] , errors: Array[String] }
@@ -62,8 +62,7 @@ module Api
       # Returns: { data: Array[serializable objects] , errors: Array[String] }
       # Does: gets the recordings that belong to the specific room friendly_id
       def recordings
-        room = Room.find_by(friendly_id: params[:room_friendly_id]).id
-        recordings = room.recordings
+        recordings = @room.recordings
 
         render_json(data: recordings, status: :ok, include: :formats)
       end
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -62,9 +62,7 @@ module Api
       # Returns: { data: Array[serializable objects] , errors: Array[String] }
       # Does: gets the recordings that belong to the specific room friendly_id
       def recordings
-        recordings = @room.recordings
-
-        render_json(data: recordings, status: :ok, include: :formats)
+        render_json(data: @room.recordings, status: :ok, include: :formats)
       end
 
       private
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -62,8 +62,8 @@ module Api
       # Returns: { data: Array[serializable objects] , errors: Array[String] }
       # Does: gets the recordings that belong to the specific room friendly_id
       def recordings
-        room_id = Room.find_by(friendly_id: params[:room_friendly_id]).id
-        recordings = Recording.where(room_id:)
+        room = Room.find_by(friendly_id: params[:room_friendly_id]).id
+        recordings = room.recordings
 
         render_json(data: recordings, status: :ok, include: :formats)
       end
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -58,6 +58,16 @@ module Api
         render_json status: :created
       end
 
+      # GET /api/v1/rooms/:friendly_id/recordings.json
+      # Returns: { data: Array[serializable objects] , errors: Array[String] }
+      # Does: gets the recordings that belong to the specific room friendly_id
+      def recordings
+        room_id = Room.find_by(friendly_id: params[:room_friendly_id]).id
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/rooms_controller.rb
```diff
@@ -58,6 +58,16 @@ def create
         render_json status: :created
       end
 
+      # GET /api/v1/rooms/:friendly_id/recordings.json
+      # Returns: { data: Array[serializable objects] , errors: Array[String] }
+      # Does: gets the recordings that belong to the specific room friendly_id
+      def recordings
+        room_id = Room.find_by(friendly_id: params[:room_friendly_id]).id
+        recordings = Recording.where(room_id:)
```
farhatahmad, line 66: use `find_room` from above -> then you can just do `@room.recordings`

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -58,6 +58,16 @@ def create
...
+
+        render_json(data: recordings, status: :ok, include: :formats)
```
farhatahmad, line 68: Nice include :+1:

spec/controllers/rooms_controller_spec.rb
```diff
@@ -90,4 +90,25 @@
       expect(response).to have_http_status(:created)
     end
   end
+
+  describe '#recordings' do
+    it 'ids of recordings that belong only to room with given room_friendly_id' do
```
farhatahmad, line 95: This sentence is confusing :laughing: 

`it 'returns recordings belonging to the room'`

spec/controllers/rooms_controller_spec.rb
```diff
@@ -90,4 +90,25 @@
...
+      room1 = create(:room, user:, friendly_id: 'friendly_id_1')
+      room2 = create(:room, user:, friendly_id: 'friendly_id_2')
+      recordings = create_list(:recording, 5, room: room1)
+      create_list(:recording, 5, room: room2)
+      get :recordings, params: { room_friendly_id: room1.friendly_id }
+      recording_ids = JSON.parse(response.body)['data'].map { |recording| recording['id'] }
+      expect(response).to have_http_status(:ok)
+      expect(recording_ids).to eq(recordings.pluck(:id))
```
farhatahmad, line 103: I wouldn't do a direct `eq` here. Look at https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/contain-exactly-matcher

farhatahmad, line 103: Really nicely done though

spec/controllers/rooms_controller_spec.rb
```diff
@@ -90,4 +90,25 @@
...
+    end
+
+    it 'no matching ids of recordings that belong only to room with given room_friendly_id' do
```
farhatahmad, line 106: `it 'returns an empty array if the room has no recordings'`

spec/controllers/rooms_controller_spec.rb
scouillard: I don't think room2 is used?
Is it part of the test?
line 97 and 99

scouillard: room2 isn't used?
Is it part of the test?
line 97 and 99

scouillard: Would it be a valid test if we remove room2 (line 97 and 99) as well?

hadicheaito1: ```recordings_ids``` seems to have extra brackets for some reason which causes ```contain_exactly``` to fail. Not sure what to use in this case now. We also use ```eq``` above directly in other tests.
image

hadicheaito1: Yes it would still pass, but the point of them is to show that even though there are recordings for other rooms that exist in the db, the request only returns the recordings that belong to the room with the specific friendly_id given in the request.

app/controllers/api/v1/rooms_controller.rb
hadicheaito1: The param comes in as ```room_friendly_id```  (rails routes just named it like that automatically) and not ```friendly_id```, that's why I didn't use the ```find_room``` function since it wouldn't work.

I can clean it up to become like this though:
`room = Room.find_by(friendly_id: params[:room_friendly_id]).id`
` recordings = room.recordings`

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -58,6 +58,16 @@ def create
         render_json status: :created
       end
 
+      # GET /api/v1/rooms/:friendly_id/recordings.json
+      # Returns: { data: Array[serializable objects] , errors: Array[String] }
+      # Does: gets the recordings that belong to the specific room friendly_id
+      def recordings
+        room = Room.find_by(friendly_id: params[:room_friendly_id]).id
```
farhatahmad, line 65: You can use `find_room`

spec/controllers/rooms_controller_spec.rb
farhatahmad: `match_array`

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -58,6 +58,15 @@ def create
         render_json status: :created
       end
 
+      # GET /api/v1/rooms/:friendly_id/recordings.json
+      # Returns: { data: Array[serializable objects] , errors: Array[String] }
+      # Does: gets the recordings that belong to the specific room friendly_id
+      def recordings
+        recordings = @room.recordings
+
+        render_json(data: recordings, status: :ok, include: :formats)
```
farhatahmad, line 67: I would just do `render_json(data: room.recordings, status: :ok, include: :formats)` one shot

farhatahmad (approved):

PR #3410, opened 2022-04-21
Recording room list front end

hadicheaito1:
## Description

- Added RoomRecordings page that displays the recordings that only belong to that room being viewed (based on friendly_id)
- Changed name of RecordingsTable --> Recordings
- Created new useRoomRecordings query that does a GET request to back end
## Testing Steps

1. Pull changes
2. Create Room
3. Create Recording for that room (matching room_id of room created in step 2)
4. Go to /rooms page
5. Click on created room
6. Recording created for that Room should be displayed

image

### DIFF
--- a/app/javascript/components/rooms/FeatureTabs.jsx
+++ b/app/javascript/components/rooms/FeatureTabs.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Row, Tabs, Tab } from 'react-bootstrap';
-import RecordingsTable from './RecordingsTable';
+import RoomRecordings from '../recordings/RoomRecordings';
 import Presentation from './Presentation';
 import SharedAccess from './SharedAccess';
 import RoomSettings from './RoomSettings';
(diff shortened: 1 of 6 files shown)


app/javascript/components/rooms/FeatureTabs.jsx
```diff
@@ -10,7 +10,7 @@ export default function FeatureTabs() {
     <Row className="py-5">
```
scouillard, line 10: This row seems to be messing the layout for some reason - the FeatureTab is expanding over the container width.
Adding w-100 (width: 100%) to className will fix the overflow on the right (but not on the left somehow).

app/javascript/hooks/queries/recordings/useRoomRecordings.jsx
```diff
@@ -0,0 +1,11 @@
+import { useQuery } from 'react-query';
+import axios, { ENDPOINTS } from '../../../helpers/Axios';
+
+export default function useRoomRecordings(friendlyId) {
+  return useQuery('getRoomRecordings', () => axios.get(ENDPOINTS.room_recordings(friendlyId), {
+    headers: {
+      'Content-Type': 'application/json',
+      Accept: 'application/json',
+    },
```
scouillard, line 9: I think we don't need this since we are importing axiosInstance as default from the Axios helper?

farhatahmad (approved):

PR #3408, opened 2022-04-21
Backend: Populated the meeting options table.

KH-Amir-TN:
## Description

Populated the meeting options table with preset options for the general BBB parameters used by GL and GL rooms options.
## Testing Steps

> 1. Pull the code.
> 2. Install the dependencies `bundle install && npm|yarn install`.
> 3. Clean the previous assets build by running `rm app/assets/builds/*` (This won't remove .keep since it's hidden).
> 4. Clean the database and tmp files for a better isolation by running `rails tmp:clear && rails db:schema:cache:clear && rails db:drop && rails db:create && rails db:migrate:with_data`
> 5. Run the linter and specs `bundle exec rubocop --parallel && bundle exec rspec && npx eslint app/javascript/* --ext .jsx,.js`
> 6. Run `./bin/dev` to run the assets builders processes and the Puma server all at once.

PR #3411, opened 2022-04-21
Adding room meeting opts creation

KH-Amir-TN:
## Description

Adding room meeting options creation on room `#create`.
-> The `joinOnStart` feature is also included in this PR.
## Testing Steps

> 1. Pull the code.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,11 +5,11 @@ require:
 
 AllCops:
   Exclude:
-    - 'bin/**/*'
-    - 'db/schema.rb'
-    - 'db/data_schema.rb'
-    - 'vendor/**/*'
-    - 'vendor/bundle/**/*'
+    - "bin/**/*"
+    - "db/schema.rb"
+    - "db/data_schema.rb"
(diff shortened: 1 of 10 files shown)


KH-Amir-TN:
depends on #3408

KH-Amir-TN:
Rebased on 4e664e3d39f6b74011484615fbc271a75ddc5818

KH-Amir-TN:
Do you prefer extracting all tests into a dedicated PR?

.rubocop.yml
```diff
@@ -5,11 +5,11 @@ require:
```
KH-Amir-TN, line 5: The formatting was done by an extension.
If you feel that it's not useful, I'll revert it.

farhatahmad:
I think we're over complicating this from the get go, which will cause us problems down the line. I'd really like to take a simple, incremental approach to this to avoid over engineering it from the get go

Let's focus on one simple setting at a time and go from there, adjusting as needed.

Side note - I think we might actually drop `Automatically join the room` because it's not possible looking at the mockups

PR #3414, opened 2022-04-22
Frontend: Add Avatar Form/Feature

scouillard:
## Description

User can now upload an Avatar.
Avatar is displayed in the Header.
Add basic validation: 
- image needs to be a jpg, jpeg or png.
- max size has been set to 300000 after some trial and error.

## Testing Steps

Need to merge with #3417 (the backend) and preferably #3418 (frontend layout).
Or pull #3419 which merges all three PRs.

## Screenshots (if appropriate):

image

## TODO:
Add Remove Avatar button.

PR #3417, opened 2022-04-26
Backend: Add Avatar Form/Feature

scouillard:
## Description

Add ActiveStorage to the project.
Rework rendering of current_user in SessionsController - now we have more control on what current_user data that is returned.
If User has no avatar, a basic/default Avatar image is returned.

NOTE: We don't have an official mock up for the Avatar form/feature as of right now (26/04/22) - has to be approved by designer/team. 

## Testing Steps

Need to merge with #3414 (frontend) and preferably #3418 (frontend layout).
Or pull #3419 which merges all three PRs.

image


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::Base
-  before_action do
-    # Unless the request format is explicitly json Rails will mitigate the responsability to CSR to handle it.
-    render 'components/index' unless valid_api_request?
-  end
+  # before_action do
```
KH-Amir-TN, line 4: After development it would be better to uncomment this and make the required e2e tests in a complete environment.

scouillard:
Right now, there are real pdf and png files in specs/fixtures/files for the specs.
We are testing if the uploaded Avatar is in a proper image format.
Is there a way to create a fake file with FactoryBot?

app/models/user.rb
```diff
@@ -4,9 +4,35 @@ class User < ApplicationRecord
...
   validates :name, presence: true # TODO: amir - Change into full_name or seperate first and last name.
   # TODO: amir - Validate email format.
   validates :email, presence: true, uniqueness: { case_sensitive: false, scope: :provider }
   validates :provider, presence: true
   validates :password_confirmation, presence: true, on: :create
+
+  # TODO: samuel - ActiveStorage validations needs to be discussed and implemented.
+  validate :avatar_validation
+
+  after_commit :set_default_avatar, on: %i[create update]
```
farhatahmad, line 19: I would only do this on create

spec/controllers/users_controller_spec.rb
```diff
@@ -27,6 +27,18 @@
...
+      user = create(:user)
+      patch :update, params: { id: user.id, user: { avatar: fixture_file_upload(file_fixture('default-avatar.png'), 'image/png') } }
+      expect(user.reload.avatar.filename).to eq('default-avatar.png')
+    end
+
+    it 'returns an error if the avatar is a pdf' do
+      user = create(:user)
+      patch :update, params: { id: user.id, user: { avatar: fixture_file_upload(file_fixture('default-pdf.pdf'), 'pdf') } }
+      expect(user.reload.avatar.filename).not_to eq('default-pdf.pdf')
+    end
```
farhatahmad, line 41: You need a test for file size as well

app/models/user.rb
scouillard: I plan on adding a "Remove Avatar" button today.
The on update will (should) then set the default avatar as fallback of removing the avatar.

farhatahmad: I'm not sure I would do it like that - I'd probably manually set it to blank in the update

scouillard: What do you mean?
You would not add a "Remove Avatar" button?

farhatahmad: No, removing the avatar shouldn't depend on an `after_commit` to clear the avatar. I would just depend on the form data being submitted to be properly dealt with instead

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,10 @@
...
+  #   # Unless the request format is explicitly json Rails will mitigate the responsability to CSR to handle it.
+  #   render 'components/index' unless valid_api_request?
+  # end
```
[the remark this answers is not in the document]
farhatahmad: Dont forget about this

farhatahmad (dismissed):

spec/controllers/users_controller_spec.rb
```diff
@@ -27,6 +27,24 @@
...
+
+    it 'returns an error if the avatar size is too large' do
+      user = create(:user)
+      patch :update, params: { id: user.id, user: { avatar: fixture_file_upload(file_fixture('large-avatar.jpg'), 'jpg') } }
+      expect(user.reload.avatar.blob.byte_size).to be < 3_000_000
```
farhatahmad: This test doesn't actually work. You should check that no avatar is set

spec/controllers/users_controller_spec.rb
```diff
@@ -27,6 +27,32 @@
...
+      user.avatar.attach(io: fixture_file_upload('default-avatar.png'), filename: 'default-avatar.png', content_type: 'image/png')
+      expect(user.reload.avatar).to be_attached
+      delete :purge_avatar, params: { id: user.id }
+      expect(user.reload.avatar).not_to be_attached
+    end
+
+    it 'returns an error if the avatar is a pdf' do
+      user = create(:user)
+      patch :update, params: { id: user.id, user: { avatar: fixture_file_upload(file_fixture('default-pdf.pdf'), 'pdf') } }
+      expect(user.reload.avatar).not_to be_attached
```
hadicheaito1, line 48: Not sure if important, but if there is a way to check for an error occurring instead of the avatar not being attached it would be more specific.

Or just change ```'returns an error if the avatar is a pdf'``` to ```doesn't attach avatar is pdf``` or something similiar

spec/controllers/users_controller_spec.rb
```diff
@@ -27,6 +27,32 @@
...
+    end
+
+    it 'returns an error if the avatar size is too large' do
```
hadicheaito1, line 51: Same idea here as above (if you agree with the idea above)

hadicheaito1 (approved):

spec/controllers/users_controller_spec.rb
scouillard: Yeah, you're right.

PR #3437, opened 2022-05-05
Added waiting for room to start functionality

farhatahmad:
## Description

To try and keep this PR as simple as possible, this does not include:
- The check to see if a room has already started (it treats all rooms as if they haven't started)
- Visual waiting indicator

## Testing Steps

1. Go to any room and add `/join` to the end of the url
2. Open the room in another tab
3. Click `join` on the first window
4. Wait a couple seconds and click start on the second window
5. Ensure the first window is redirected at the same time

### DIFF
--- a/app/channels/rooms_channel.rb
+++ b/app/channels/rooms_channel.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class RoomsChannel < ApplicationCable::Channel
+  def subscribed
+    stream_from "#{params[:friendly_id]}_rooms_channel"
+  end
+
+  def unsubscribed; end
+end
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -39,6 +39,8 @@ module Api
           join_url = bbb_api.start_meeting room: @room, meeting_starter: current_user, options: options
           logger.info "meeting successfully started for room(friendly_id):#{@room.friendly_id} by #{meeting_starter}."
 
+          ActionCable.server.broadcast "#{@room.friendly_id}_rooms_channel", join_url.to_s
+
           render_json data: { join_url: }, status: :created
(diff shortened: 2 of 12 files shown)


app/channels/rooms_channel.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class RoomsChannel < ApplicationCable::Channel
+  def subscribed
+    stream_from "#{params[:friendly_id]}_rooms_channel"
```
farhatahmad, line 5: Each room gets it's own channel

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -39,6 +39,8 @@ def start
           join_url = bbb_api.start_meeting room: @room, meeting_starter: current_user, options: options
           logger.info "meeting successfully started for room(friendly_id):#{@room.friendly_id} by #{meeting_starter}."
 
+          ActionCable.server.broadcast "#{@room.friendly_id}_rooms_channel", join_url.to_s
```
farhatahmad, line 42: Broadcast the join url to everyone informing them that the room has started (needs to be changed in the next PR)

app/javascript/channels/rooms_channel.jsx
```diff
@@ -0,0 +1,24 @@
...
+
+    disconnected() {
+      // Called when the subscription has been terminated by the server
+      console.log('disconnected');
+    },
+
+    received(joinUrl) {
+      // Called when there's incoming data on the websocket for this channel
+      console.log('received');
+      window.location.replace(joinUrl);
```
farhatahmad, line 21: Redirect to the joinUrl once the broadcast is received

scouillard: Where does it gets joinUrl from?

farhatahmad: Yeah it's from the RoomsController

PR #3441, opened 2022-05-12
Back-end: Recordings Re-Sync

hadicheaito1:
## Description

- New recordingsReSync route that points to recordings#recordings
- Added get_recordings method in bbb_api
- Added recordings call that uses new RecordingSync service
- recording_resync function in recording_sync service gets the rooms that belong to the current user, deletes all their recordings, does a request to bbb server to get recording that match room meeting ids, creates recording, creates format(s) for that new recording created
- Added bbb#get_recordings test in bbb_api_spec

### DIFF
--- a/app/controllers/api/v1/recordings_controller.rb
+++ b/app/controllers/api/v1/recordings_controller.rb
@@ -15,8 +15,8 @@ module Api
       end
 
       def recordings
-        recording_sync = RecordingSync.new
-        recording_sync.recording_resync(user: current_user)
+        recording_sync = RecordingsSync.new
+        recording_sync.recordings_resync(user: current_user)
 
         render_json(status: :ok)
       end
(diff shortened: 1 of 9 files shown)


app/controllers/api/v1/recordings_controller.rb
```diff
@@ -11,6 +13,12 @@ def index
 
         render_json(data: recordings, status: :ok, include: :formats)
       end
+
+      def recordings
+        RecordingsSync.new(user: current_user).call
+
+        render_json(status: :ok)
+      end
```
farhatahmad, line 21: Missing tests for this

config/routes.rb
```diff
@@ -22,7 +22,11 @@
           get '/recordings', to: 'rooms#recordings'
         end
       end
-      resources :recordings, only: [:index]
+      resources :recordings, only: [:index] do
+        collection do
+          get '/recordingsReSync', to: 'recordings#recordings'
```
farhatahmad, line 27: I would make this just `/resync`

farhatahmad (dismissed):

farhatahmad (approved):

PR #3442, opened 2022-05-12
Recording sync front end

hadicheaito1:
## Description

New recording resync feature
- When button is pressed on the front end of the recordings page, recording_resync is triggered and then the back end is triggered
## Testing Steps

- Create branch
- Pull code
- Create two rooms
- Put one room with meeting_id random-5678484
- Put the other room with meetinf_id random-1291479
- Change .env file for GL to point to this server:

- Run your server
- Go to recordings page and click resync button
- Recordings should appear right after clicking the button

You can test this yourself as well by creating recordings on any bbb server using api mate, but to fully test it, you would need to enable other recording formats to test the creation of the recording when there are multiple formats given. https://docs.bigbluebutton.org/2.4/install.html#installing-additional-recording-processing-formats

image

### DIFF
--- a/app/controllers/api/v1/recordings_controller.rb
+++ b/app/controllers/api/v1/recordings_controller.rb
@@ -3,6 +3,8 @@
 module Api
   module V1
     class RecordingsController < ApplicationController
+      skip_before_action :verify_authenticity_token # TODO: - Revisit this.
+
       # GET /api/v1/recordings.json
       # Returns: { data: Array[serializable objects(recordings)] , errors: Array[String] }
       # Does: Returns all the Recordings from all the current user's rooms
@@ -10,6 +12,13 @@ module Api
         recordings = current_user.recordings&.includes(:formats)
 
         render_json(data: recordings, status: :ok, include: :formats)
+      end
+
+      def recordings
+        recording_sync = RecordingSync.new
+        recording_sync.recording_resync(user: current_user)
+
+        render_json(status: :ok)
       end
     end
   end
--- a/app/controllers/api/v1/recordings_controller.rb
+++ b/app/controllers/api/v1/recordings_controller.rb
@@ -15,8 +15,7 @@ module Api
       end
 
       def recordings
-        recording_sync = RecordingSync.new
-        recording_sync.recording_resync(user: current_user)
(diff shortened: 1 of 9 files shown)


app/services/recording_sync.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+class RecordingSync
+  def initialize; end
+
+  def recording_resync(user:)
```
farhatahmad, line 6: If this is a service that does one thing, this method should be called `call`

app/services/recording_sync.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+class RecordingSync
+  def initialize; end
```
farhatahmad, line 4: `user` should be passed here

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -11,6 +13,13 @@ def index
 
         render_json(data: recordings, status: :ok, include: :formats)
       end
+
+      def recordings
+        recording_sync = RecordingSync.new
+        recording_sync.recording_resync(user: current_user)
```
farhatahmad, line 19: Then you would call the service like `RecordingSync.new(user: current_user).call`

app/services/recording_sync.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def recording_resync(user:)
+    bbb_api = BigBlueButtonApi.new
```
farhatahmad, line 7: No need for a separate variable since it's only used once

app/services/recording_sync.rb
```diff
@@ -0,0 +1,60 @@
...
+    rooms = user.rooms
+
+    # Get the meeting_id of all the current user's rooms
+    meeting_ids = rooms.map(&:meeting_id)
+
+    # Clear current recordings
+    Recording.destroy_by(room_id: rooms.map(&:id))
+
+    # Get recordings
```
farhatahmad, line 16: There's too many comments here - drop all the comments that are straight forward and only leave comments for things that are either odd or complex

PR #3446, opened 2022-05-13
Backend share access cleanup

scouillard:

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -10,6 +10,18 @@ module Api
       # Returns: { data: Array[serializable objects] , errors: Array[String] }
       # Does: Creates and saves a new user record in the database with the provided parameters.
 
+      def index
+        users = User.all.map do |user|
+          {
+            id: user.id,
+            name: user.name,
+            email: user.email,
+            avatar: user_avatar(user)
+          }
+        end
+        render_json data: users, status: :ok
+      end
+
       def create
         # TODO: amir - ensure accessibility for unauthenticated requests only.
         user = User.new({ provider: 'greenlight' }.merge(user_params)) # TMP fix for presence validation of :provider
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -35,6 +35,11 @@ class ApplicationController < ActionController::Base
     @current_user ||= User.find_by(id: session[:user_id])
   end
 
+  def user_avatar(user)
+    default_avatar = ActionController::Base.helpers.image_path('default-avatar.png')
+    user.avatar.attached? ? url_for(user.avatar) : default_avatar
+  end
+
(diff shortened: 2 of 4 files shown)


app/controllers/api/v1/users_controller.rb
```diff
@@ -10,6 +10,18 @@ class UsersController < ApplicationController
...
+        users = User.all.map do |user|
+          {
+            id: user.id,
+            name: user.name,
+            email: user.email,
+            avatar: user_avatar(user)
+          }
+        end
+        render_json data: users, status: :ok
+      end
```
farhatahmad, line 23: Why would we ever need to return a list of all users?

app/controllers/application_controller.rb
```diff
@@ -35,6 +35,11 @@ def current_user
     @current_user ||= User.find_by(id: session[:user_id])
   end
 
+  def user_avatar(user)
+    default_avatar = ActionController::Base.helpers.image_path('default-avatar.png')
+    user.avatar.attached? ? url_for(user.avatar) : default_avatar
+  end
```
farhatahmad, line 41: I wouldn't put this here - I would move this to the `User` model instead

app/controllers/api/v1/users_controller.rb
scouillard: Administrators should be able to list/update all users

image

farhatahmad: That's a completely random issue that has nothing to do with what you're working on

app/controllers/application_controller.rb
scouillard: This method is used in User, Sessions and SharedAccess.
Should we create a User controller concern?
And maybe add current_user in there as well?

farhatahmad: No - it's an attribute for `user`, so it should be added to the `User` class.

That way whenever you call `user.avatar_with_default` or whatever you call the method, you get the right value returned

scouillard:
A Controller helper method is used in User model to generate a dynamic url to the AssetPipeline
`ActionController::Base.helpers.image_url('default-avatar.png')`

This implementation is dirty, but I haven't found another solution yet.

For what it's worth, it's the most (the only?) referenced solution on stackoverflow.

This logic would usually be in a view helper.

app/controllers/api/v1/users_controller.rb
```diff
@@ -10,6 +10,18 @@ class UsersController < ApplicationController
       # Returns: { data: Array[serializable objects] , errors: Array[String] }
       # Does: Creates and saves a new user record in the database with the provided parameters.
 
+      def index
```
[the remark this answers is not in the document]
scouillard: Agreed, I could have swore I had removed it.

farhatahmad (approved):

PR #3449, opened 2022-05-13
Backend: Add ShareAccess feature (3)

scouillard:
## Description

- Add ShareAccess model, controller, routing and specs.
- ShareAccess is a special joined table between User and Room.
- Add User<room> methods to verify if user has been shared the room or is the owner of the room.

### DIFF
--- a/app/controllers/api/v1/shared_accesses_controller.rb
+++ b/app/controllers/api/v1/shared_accesses_controller.rb
@@ -6,62 +6,54 @@ module Api
       skip_before_action :verify_authenticity_token
       before_action :find_room
 
-      # POST /api/v1/shared_accesses/room/friendly_id.json
+      include Avatarable
+
+      # POST /api/v1/shared_accesses.json
       def create
-        users = User.where(id: params[:users])
+        user_ids = params[:users][:shared_users]
 
-        users.each do |user|
-          SharedAccess.find_or_create_by!(user_id: user.id, room_id: @room.id) if user.room_shareable?(@room)
-        end
+        shared_accesses = user_ids.map { |user_id| { user_id:, room_id: @room.id } }
+
+        SharedAccess.create(shared_accesses)
 
         render_json status: :ok
       end
 
-      # DELETE /api/v1/shared_accesses/room/friendly_id.json
+      # DELETE /api/v1/shared_accesses/friendly_id.json
       def destroy
-        user = User.find_by(id: params[:user_id])
-
-        SharedAccess.find_by!(user_id: user.id, room_id: @room.id).delete
+        SharedAccess.delete_by(user_id: params[:user_id], room_id: @room.id)
 
         render_json status: :ok
       end
 
-      # GET /api/v1/shared_accesses/room/friendly_id/shared_users.json
-      def shared_users
-        shared_users = []
-
-        # User is added to the shared_user list if the room is shared to the user and it is not already included in shared_user
-        User.all.each do |user|
-          shared_users << user if user.room_shared?(@room) && shared_users.exclude?(user)
-        end
+      # GET /api/v1/shared_accesses/friendly_id.json
+      def show
+        shared_users = @room.shared_users.to_a
 
         shared_users.map! do |user|
           {
             id: user.id,
             name: user.name,
             email: user.email,
-            avatar: user.user_avatar
+            avatar: user_avatar(user)
           }
         end
 
         render_json data: shared_users, status: :ok
       end
 
-      # GET /api/v1/shared_accesses/room/friendly_id/shareable_users.json
+      # GET /api/v1/shared_accesses/friendly_id/shareable_users.json
       def shareable_users
-        shareable_users = []
-
-        # User is added to the shareable_user list unless it's the room owner or the room is already shared to the user
-        User.all.each do |user|
-          shareable_users << user if user.room_shareable?(@room)
-        end
+        # Can't share the room if it's already shared or it's the room owner
+        unshareable_users = [@room.shared_users.pluck(:id) << @room.user_id]
+        shareable_users = User.where.not(id: unshareable_users).to_a
 
         shareable_users.map! do |user|
           {
             id: user.id,
             name: user.name,
             email: user.email,
-            avatar: user.user_avatar
+            avatar: user_avatar(user)
           }
         end
 
--- a/app/controllers/api/v1/shared_accesses_controller.rb
+++ b/app/controllers/api/v1/shared_accesses_controller.rb
@@ -1,0 +1,78 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class SharedAccessesController < ApplicationController
+      skip_before_action :verify_authenticity_token
+      before_action :find_room
+
+      # POST /api/v1/shared_accesses/room/friendly_id.json
+      def create
+        users = User.where(id: params[:users])
+
+        users.each do |user|
+          SharedAccess.find_or_create_by!(user_id: user.id, room_id: @room.id) if user.room_shareable?(@room)
+        end
+
+        render_json status: :ok
+      end
+
+      # DELETE /api/v1/shared_accesses/room/friendly_id.json
+      def destroy
+        user = User.find_by(id: params[:user_id])
+
+        SharedAccess.find_by!(user_id: user.id, room_id: @room.id).delete
+
+        render_json status: :ok
+      end
+
+      # GET /api/v1/shared_accesses/room/friendly_id/shared_users.json
+      def shared_users
+        shared_users = []
+
+        # User is added to the shared_user list if the room is shared to the user and it is not already included in shared_user
+        User.all.each do |user|
+          shared_users << user if user.room_shared?(@room) && shared_users.exclude?(user)
+        end
+
+        shared_users.map! do |user|
+          {
+            id: user.id,
+            name: user.name,
+            email: user.email,
+            avatar: user.user_avatar
+          }
+        end
+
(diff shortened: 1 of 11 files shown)


config/routes.rb
```diff
@@ -26,6 +26,13 @@
         end
       end
       resources :room_settings, only: %i[update], param: :friendly_id
+      resources :shared_accesses, only: :destroy, path: '/shared_accesses/room', param: :friendly_id do
+        member do
+          post :create
```
scouillard, line 31: Not sure about this solution, need to find a way to send friendly_id with the POST/create request.
Could also have a hidden_field with the friendly_id in the Create form.

config/routes.rb
```diff
@@ -26,6 +26,13 @@
         end
       end
       resources :room_settings, only: %i[update], param: :friendly_id
+      resources :shared_accesses, only: :destroy, path: '/shared_accesses/room', param: :friendly_id do
```
farhatahmad, line 29: I would drop the `path: ...` here

config/routes.rb
```diff
@@ -26,6 +26,13 @@
...
+        member do
+          post :create
+          get '/shared_users', to: 'shared_accesses#shared_users'
```
farhatahmad, line 32: Can this not be `index` ?

db/schema.rb
```diff
@@ -42,9 +42,6 @@
     t.index ["blob_id", "variation_digest"], name: "index_active_storage_variant_records_uniqueness", unique: true
   end
 
-  create_table "data_migrations", primary_key: "version", id: :string, force: :cascade do |t|
-  end
-
```
farhatahmad, line 47: This shouldn't be here

db/schema.rb
```diff
@@ -97,6 +94,25 @@
...
+  end
+
+  create_table "user_rooms", force: :cascade do |t|
+    t.bigint "user_id"
+    t.bigint "room_id"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["room_id"], name: "index_user_rooms_on_room_id"
+    t.index ["user_id"], name: "index_user_rooms_on_user_id"
+  end
```
farhatahmad, line 114: ?

spec/controllers/shared_accesses_controller_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      expect(user.shared_rooms).to include(room)
+    end
+
+    it "doesn't share a room with a user that it not selected" do
+      room = create(:room)
+      user = create(:user)
+      random_user = create(:user)
+      post :create, params: { friendly_id: room.friendly_id, room_id: room.id, users: [random_user.id] }
+      expect(user.shared_rooms).not_to include(room)
+    end
```
farhatahmad, line 24: This test isn't really useful

app/controllers/api/v1/shared_accesses_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      skip_before_action :verify_authenticity_token
+      before_action :find_room
+
+      # POST /api/v1/shared_accesses/room/friendly_id.json
+      def create
+        users = User.where(id: params[:users])
+
+        users.each do |user|
+          SharedAccess.find_or_create_by!(user_id: user.id, room_id: @room.id) if user.room_shareable?(@room)
+        end
```
farhatahmad, line 15: Performance wise - this isn't very good. What I would do instead is filter the users first, and the create them all in one shot instead of looping through them manually

farhatahmad: And why the `find_or_create_by`?

app/controllers/api/v1/shared_accesses_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+
+        render_json status: :ok
+      end
+
+      # DELETE /api/v1/shared_accesses/room/friendly_id.json
+      def destroy
+        user = User.find_by(id: params[:user_id])
+
+        SharedAccess.find_by!(user_id: user.id, room_id: @room.id).delete
```
farhatahmad, line 24: https://apidock.com/rails/v6.0.0/ActiveRecord/Relation/delete_by

app/models/user.rb
```diff
@@ -18,6 +22,21 @@ class User < ApplicationRecord
   # TODO: samuel - ActiveStorage validations needs to be discussed and implemented.
   validate :avatar_validation
 
+  def room_owner?(room)
+    id == room.user_id
+  end
+
+  def room_shared?(room)
+    shared_rooms.pluck(:friendly_id).include?(room.friendly_id)
+  end
```
farhatahmad, line 31: I wouldn't put this in the User model - I'd probably put it in the Room model instead

app/controllers/api/v1/shared_accesses_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      end
+
+      # GET /api/v1/shared_accesses/room/friendly_id/shareable_users.json
+      def shareable_users
+        shareable_users = []
+
+        # User is added to the shareable_user list unless it's the room owner or the room is already shared to the user
+        User.all.each do |user|
+          shareable_users << user if user.room_shareable?(@room)
+        end
```
farhatahmad, line 57: Performance wise - this isn't that good either. We'd probably want to do a query here instead. Get all users where the user's id is not either the room owner's id, or an id of anyone with who it's already shared with

test/models/shared_access_test.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class SharedAccessTest < ActiveSupport::TestCase
+  # test "the truth" do
+  #   assert true
+  # end
+end
```
farhatahmad, line 9: You can delete this

test/fixtures/shared_accesses.yml
```diff
@@ -0,0 +1,9 @@
+# Read about fixtures at https://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
+
+one:
+  user: one
+  room: one
+
+two:
+  user: two
+  room: two
```
farhatahmad, line 9: You can delete this

config/routes.rb
scouillard: Well we need a endpoint for shared_users, and one for shareable_users.

scouillard: Or we could return both via index? and then filter via frontend?

config/routes.rb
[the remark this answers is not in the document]
scouillard: We need to scope this with friendly_id because it is needed as a params in the POST/create request.
Could also have a hidden_field with the friendly_id in the Create form.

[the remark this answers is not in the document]
farhatahmad: It will be scoped if you add it to `only: %i[create destory]`

config/routes.rb
farhatahmad: No they're 2 separate endpoints, but `shared_users` is the business definiton of what this is. Sorry not `index`, but instead `show`

farhatahmad: Since `show` means to return the shared accesses from a specific room

scouillard: You're right.

app/controllers/api/v1/shared_accesses_controller.rb
farhatahmad:
```
User.where.not(id: [room_owner_id, list of shared users]
```

config/routes.rb
scouillard: I find it annoying that the default path is `/shared_accesses/friendly_id ` when friendly_id is a param of room.

app/models/user.rb
scouillard: They are not needed anymore I will remove them

db/schema.rb
farhatahmad: Still shouldn't be here

spec/controllers/shared_accesses_controller_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    it 'unshares a room with a user' do
+      room = create(:room)
+      user = create(:user)
+      create(:shared_access, user_id: user.id, room_id: room.id)
+      delete :destroy, params: { friendly_id: room.friendly_id, room_id: room.id, user_id: user.id }
+      expect(user.shared_rooms).not_to include(room)
+    end
+  end
+
+  describe '#shared_users' do
```
farhatahmad, line 29: This should be `show`

db/schema.rb
scouillard: It seems to be removed

spec/controllers/shared_accesses_controller_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+RSpec.describe Api::V1::SharedAccessesController, type: :controller do
+  before do
+    request.headers['ACCEPT'] = 'application/json'
+  end
+
+  describe '#create' do
+    it 'shares a room with a user' do
+      room = create(:room)
+      user = create(:user)
+      post :create, params: { friendly_id: room.friendly_id, room_id: room.id, users: { shared_users: [user.id] } }
```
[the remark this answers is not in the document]
scouillard: Thank you, it's an artefact

db/schema.rb
farhatahmad: No as in this change shouldn't be here - why are you removing the `data_migrations` table?

scouillard: My bad - not sure what happened here.

PR #3454, opened 2022-05-17
Move user_avatar method from User model to Controller/Concerns

scouillard:
## Description

Move the user_avatar method from the User model to a Controller Concern.
Wherever the avatar url is needed, just add the following line `include Avatarable` in the controller and you will be able to use the `user_avatar(user)` method.

### DIFF
--- a/app/controllers/concerns/avatarable.rb
+++ b/app/controllers/concerns/avatarable.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Avatarable
+  extend ActiveSupport::Concern
+
+  def user_avatar(user)
+    return url_for(user.avatar) if user.avatar.attached?
+
+    ActionController::Base.helpers.image_url('default-avatar.png')
+  end
+end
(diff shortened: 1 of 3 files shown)


app/controllers/concerns/avatarable.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Avatarable
+  extend ActiveSupport::Concern
+
+  def user_avatar(user)
+    return url_for(user.avatar) if user.avatar.attached?
+
+    ActionController::Base.helpers.image_url('default-avatar.png')
```
farhatahmad, line 9: Can we drop `ActionController::Base.helpers.` now? and just do `image_url`?

scouillard: We can't.
Althought it seems view_context (from ActionView) should/could be used instead of ActionController.

`view_context.image_url('default-avatar.png')  `

scouillard: After some research, I think view_context makes more sense - and is cleaner. Do you agree?

farhatahmad: Yeah

PR #3447, opened 2022-05-13
Frontend: Add ShareAccess feature (4)

scouillard:
## Description

- Room owner can now share the room with other users.
- Users that have not been shared the room yet will appear in the "Select User" form.
- Users that have been shared the room will appear in the "Shared" list.

PR #3461, opened 2022-05-19
Delete Recording

hadicheaito1:
## Description

- Allows you to delete a recording that is on the 'Recordings' page
- Click on fa icon to delete

## Testing Steps

1. Pull code
2. Create room
3. Create recording for that room
4. Go to 'Recordings' page
5. Click on delete icon for that recording
6. Check in db using rails console to ensure that recording is deleted

image
image

### DIFF
--- a/app/controllers/api/v1/recordings_controller.rb
+++ b/app/controllers/api/v1/recordings_controller.rb
@@ -14,6 +14,12 @@ module Api
         render_json(data: recordings, status: :ok, include: :formats)
       end
 
+      def destroy
+        Recording.destroy(params[:id])
+
+        render_json(status: :ok)
+      end
+
       def resync
         RecordingsSync.new(user: current_user).call
 
--- a/app/javascript/components/recordings/Recordings.jsx
+++ b/app/javascript/components/recordings/Recordings.jsx
@@ -4,15 +4,15 @@ import {
 } from 'react-bootstrap';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faVideo, faTrashCan } from '@fortawesome/free-solid-svg-icons';
+import Modal from '../shared/Modal';
 import Spinner from '../shared/stylings/Spinner';
 import useRecordings from '../../hooks/queries/recordings/useRecordings';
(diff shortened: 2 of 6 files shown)


app/controllers/api/v1/recordings_controller.rb
```diff
@@ -14,6 +14,12 @@ def index
         render_json(data: recordings, status: :ok, include: :formats)
       end
 
+      def destroy
+        Recording.destroy(params[:id])
+
```
farhatahmad, line 19: You need to actually delete the recording here

app/javascript/components/recordings/Recordings.jsx
```diff
@@ -44,6 +46,9 @@ export default function Recordings() {
                       <div key={format.id}> {format.recording_type} </div>
                     ))}
                   </td>
+                  <td>
+                    <FontAwesomeIcon className="" icon={faTrashCan} size="lg" onClick={() => handleDeleteRecording({ recordId: recording.id })} />
```
farhatahmad, line 50: I think this should pop a modal up to be safe

scouillard: Heads up we have a Modal template in components/shared.

I did a "font-awesome" button variant as well if you can check it out in #3447 application.scss if you want (not merged yet).

image

app/controllers/api/v1/recordings_controller.rb
hadicheaito1: Wouldn't we need it to be destroy in order to remove the formats that belong it as well? Or is there another reason?

farhatahmad: As in, you need to make a request to BigBlueButton to destroy the recording

app/javascript/components/recordings/Recordings.jsx
hadicheaito1: Thanks for the tip Samuel helped a lot.

app/controllers/api/v1/recordings_controller.rb
hadicheaito1: Will do this in next PR

farhatahmad (approved):

PR #3464, opened 2022-05-19
Frontend: Add Require Moderator Approval to RoomOptions

scouillard:
## Description

Add the Room Option _Require moderator approval before joining_.

### DIFF
--- a/app/javascript/hooks/mutations/room_settings/useUpdateRoomSetting.jsx
+++ b/app/javascript/hooks/mutations/room_settings/useUpdateRoomSetting.jsx
@@ -14,16 +14,16 @@ export default function useUpdateRoomSetting(friendlyId) {
     return roomSettingData;
   };
 
-  const patchRoomSetting = (roomSettingData) => axios.patch(`/room_settings/${friendlyId}.json`, rewriteRoomSettingData(roomSettingData));
+  const patchRoomSetting = (roomSettingData) => axios.patch(`/room_settings/${friendlyId}.json`, roomSettingData);
 
   const mutation = useMutation(patchRoomSetting, {
-    // Re-fetch the current_user and redirect to homepage if Mutation is successful.
     onSuccess: () => {
       queryClient.invalidateQueries('useRoom');
     },
     onError: (error) => {
       console.log('mutate error', error);
     },
+    onMutate: (roomSettingData) => rewriteRoomSettingData(roomSettingData),
   });
 
   const handleUpdateRoomSetting = (roomSettingData) => {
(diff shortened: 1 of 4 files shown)


db/data/20220421162907_populate_meeting_options.rb
```diff
@@ -23,6 +23,7 @@ def up
       # GL only options:
       { name: 'glAnyoneCanStart', default_value: 'false' }, # true | false
       { name: 'glAnyoneJoinAsModerator', default_value: 'false' }, # true | false
+      { name: 'glRequireModeratorApproval', default_value: 'false' }, # true | false
```
farhatahmad, line 26: You can't change a migration that's already been run (since it won't run for anyone else). The setting for Moderator Approval alreadys exists - it's called Guest Policy (see v2)

hadicheaito1 (approved):

farhatahmad:
This case is a bit trickier because it shouldn't have a value of `true` or `false`. Looking at the BBB docs, we'll need to extend the `RoomSettingsRow` to handle this case

scouillard:
Right. I'm on it.

app/javascript/hooks/mutations/room_settings/useUpdateRoomSetting.jsx
```diff
@@ -2,7 +2,15 @@ import { useMutation, useQueryClient } from 'react-query';
...
 
 export default function useUpdateRoomSetting(friendlyId) {
-  const patchRoomSetting = (roomSettingData) => axios.patch(`/room_settings/${friendlyId}.json`, roomSettingData);
+  const patchRoomSetting = (roomSettingData) => {
+    // If guestPolicy setting is toggled, bool are rewritten to string values as per BBB API.
+    if (roomSettingData.settingName === 'guestPolicy') {
+      Object.defineProperty(roomSettingData, 'settingValue', {
+        value: roomSettingData.settingValue === true ? 'ASK_MODERATOR' : 'ALWAYS_ACCEPT',
+      });
+    }
```
farhatahmad, line 11: We should probably move this into a method within the mutation

app/javascript/hooks/mutations/room_settings/useUpdateRoomSetting.jsx
```diff
@@ -2,7 +2,15 @@ import { useMutation, useQueryClient } from 'react-query';
 import axios from '../../../helpers/Axios';
 
 export default function useUpdateRoomSetting(friendlyId) {
-  const patchRoomSetting = (roomSettingData) => axios.patch(`/room_settings/${friendlyId}.json`, roomSettingData);
+  const patchRoomSetting = (roomSettingData) => {
+    // If guestPolicy setting is toggled, bool are rewritten to string values as per BBB API.
+    if (roomSettingData.settingName === 'guestPolicy') {
+      Object.defineProperty(roomSettingData, 'settingValue', {
+        value: roomSettingData.settingValue === true ? 'ASK_MODERATOR' : 'ALWAYS_ACCEPT',
```
farhatahmad, line 9: What does `Object.defineProperty` do? Why's it needed here?

scouillard: esf will complain about "re-assigning" a var name if I try to overwrite a var.
This is in progress as I think I am over complicating things.

farhatahmad (approved):

PR #3465, opened 2022-05-19
Frontend: RoomSettings styling

scouillard:
## Description

Add styling to Room/RoomSettings.

image

IMO the RoomSettings label/switch would look better if it was centered in the column.

### DIFF
--- a/app/javascript/components/rooms/RoomSettingsRow.jsx
+++ b/app/javascript/components/rooms/RoomSettingsRow.jsx
@@ -11,22 +11,22 @@ export default function RoomSettingsRow(props) {
   const { handleUpdateRoomSetting } = useUpdateRoomSetting(friendlyId);
 
   return (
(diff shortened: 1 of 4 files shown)


app/javascript/components/rooms/RoomSettingsRow.jsx
```diff
@@ -9,22 +9,22 @@ export default function RoomSettingsRow(props) {
   const { handleUpdateRoomSetting } = useUpdateRoomSetting(friendlyId);
 
   return (
-    <span className="text-muted">
-      <label className="form-check-label me-5" htmlFor={settingId}>
+    <div className="text-muted py-3 d-flex" style={{ 'max-width': '400px', 'line-height': '28px' }}>
```
farhatahmad, line 12: Don't do inline styles :P

PR #3473, opened 2022-05-25
Add SearchBarQuery for backend query (1)

scouillard:
## Description

Add SearchBarQuery to complement SearchBar.

How to use SearchBarQuery:
It needs 3 components, and a hook/query:

- A Wrapper
- The SearchBarQuery
- The List of items that are searcheable

The wrapper wraps the other components.
The wrapper needs to be set up with a input/setInput useState() hook.
The wrapper needs to be set up with the hook/query to the database.

```
const [input, setInput] = useState('');
const { isLoading, data: recordings } = useRecordings(input);
```

The SearchBarQuery with setInput as props.
`      <SearchBarQuery setInput={setInput} />`

The List takes the response of the useRequest, and isLoading as props.
`        <RecordingsList recordings={recordings} isLoading={isLoading} />`

The query needs the input as a query key, and explicit params.
```
  return useQuery(['getRecordings', input], () => axios.get('/api/v1/recordings.json', {
    params: {
      q: input,
    },
```

## TODO:

- The "flickering issue" has been solved - added a recording/setRecording state (thanks for the solution @farhatahmad ). Some minor problems have arised, currently on it.

- Investigate the warning.. recordings PropType is instable.

- Need an extra fragment, or the code bricks

### DIFF
--- a/app/javascript/components/recordings/Recordings.jsx
+++ b/app/javascript/components/recordings/Recordings.jsx
@@ -8,8 +8,8 @@ import SearchBarQuery from '../shared/SearchBarQuery';
 import RecordingsList from './RecordingsList';
 
 export default function Recordings() {
-  const [input, setInput] = useState('');
-  const [recordings, setRecordings] = useState('');
+  const [input, setInput] = useState();
+  const [recordings, setRecordings] = useState();
   useRecordings(input, setRecordings);
 
   const { refetch: handleRecordingReSync } = useRecordingsReSync();
--- a/app/javascript/components/recordings/Recordings.jsx
+++ b/app/javascript/components/recordings/Recordings.jsx
@@ -1,64 +1,28 @@
-import React from 'react';
+import React, { useState } from 'react';
 import {
-  Table, Card, Row, Button,
+  Card, Button, Stack,
 } from 'react-bootstrap';
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { faVideo, faTrashCan } from '@fortawesome/free-solid-svg-icons';
-import Modal from '../shared/Modal';
-import Spinner from '../shared/stylings/Spinner';
 import useRecordings from '../../hooks/queries/recordings/useRecordings';
 import useRecordingsReSync from '../../hooks/queries/recordings/useRecordingsReSync';
-import DeleteRecordingForm from '../forms/DeleteRecordingForm';
+import SearchBarQuery from '../shared/SearchBarQuery';
+import RecordingsList from './RecordingsList';
 
 export default function Recordings() {
-  const { isLoading, data: recordings } = useRecordings();
-  const { refetch: handleRecordingReSync } = useRecordingsReSync();
+  const [input, setInput] = useState('');
+  const [recordings, setRecordings] = useState('');
+  useRecordings(input, setRecordings);
 
-  if (isLoading) return <Spinner />;
+  const { refetch: handleRecordingReSync } = useRecordingsReSync();
 
   return (
     <div className="pt-3 wide-background full-height-rooms">
-      <Row>
-        <Button className="my-2" onClick={handleRecordingReSync}>Re-sync Recordings</Button>
-        <Card className="border-0 shadow-sm p-0">
-          <Table hover className="text-secondary mb-0">
-            <thead>
-              <tr>
-                <th>Name</th>
-                <th>Length</th>
-                <th>Users</th>
-                <th>Visibility</th>
-                <th>Formats</th>
-              </tr>
-            </thead>
-            <tbody>
-              {recordings.map((recording) => (
-                <tr key={recording.id} className="recordings align-middle">
-                  <td className="text-dark">
-                    <div> <FontAwesomeIcon className="mx-2 mt-4" icon={faVideo} size="2xl" /> <strong> {recording.name} </strong> </div>
-                    <div className="small text-muted ms-5 ps-2"> {recording.created_at} </div>
-                  </td>
-                  <td> {recording.length}min </td>
-                  <td> {recording.users} </td>
-                  <td> {recording.visibility} </td>
-                  <td>
-                    {recording.formats.map((format) => (
-                      <div key={format.id}> {format.recording_type} </div>
-                    ))}
-                  </td>
-                  <td>
-                    <Modal
-                      modalButton={<FontAwesomeIcon className="" icon={faTrashCan} size="lg" />}
-                      title="Are you sure?"
-                      body={<DeleteRecordingForm recordId={recording.record_id} />}
-                    />
-                  </td>
-                </tr>
-              ))}
-            </tbody>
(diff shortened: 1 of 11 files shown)


app/javascript/components/recordings/Recordings.jsx
```diff
@@ -1,64 +1,37 @@
...
+          <thead>
+            <tr className="text-muted small">
+              <th>Name</th>
+              <th>Length</th>
+              <th>Users</th>
+              <th>Visibility</th>
+              <th>Formats</th>
+            </tr>
+          </thead>
+          <RecordingsList recordings={recordings} />
```
farhatahmad, line 32: Separating the `tbody` into a separate component seems a little weird to me

app/javascript/hooks/queries/recordings/useRecordings.jsx
```diff
@@ -1,8 +1,11 @@
 import { useQuery } from 'react-query';
-import axios, { ENDPOINTS } from '../../../helpers/Axios';
+import axios from 'axios';
 
-export default function useRecordings() {
-  return useQuery('getRecordings', () => axios.get(ENDPOINTS.recordings, {
+export default function useRecordings(input) {
+  return useQuery(['getRecordings', input], () => axios.get('/api/v1/recordings.json', {
+    params: {
+      q: input,
```
farhatahmad, line 7: Naming this param `q` is weird lol

app/models/recording.rb
```diff
@@ -9,4 +9,10 @@ class Recording < ApplicationRecord
   validates :visibility, presence: true
   validates :length, presence: true
   validates :users, presence: true
+
+  def self.search(input)
+    return where('lower(recordings.name) LIKE ?', "%#{input.downcase}%").includes(:formats) if input
```
farhatahmad, line 14: There's a postgres operator that can be used here instead called `ILIKE` https://kb.objectrocket.com/postgresql/how-to-use-ilike-in-postgres-1258

spec/controllers/recordings_controller_spec.rb
```diff
@@ -27,6 +27,15 @@
...
       expect(response_recording_ids).to be_empty
     end
+
+    it 'returns the proper recordings according to the query' do
+      recordings = create_list(:recording, 6)
+      create_list(:room, 5, user:, recordings:)
+
+      get :index, params: { q: recordings[0].name }
+      response_recording_id = JSON.parse(response.body)['data'][0]['id']
+      expect(response_recording_id).to eq(recordings[0].id)
```
farhatahmad, line 37: I think it would be better if this recording was a bit more explicit. Create 5 recordings and manually structure the names so that some have text in common

app/javascript/components/recordings/Recordings.jsx
```diff
@@ -1,64 +1,37 @@
...
+import SearchBarQuery from '../shared/SearchBarQuery';
+import RecordingsList from './RecordingsList';
 
 export default function Recordings() {
-  const { isLoading, data: recordings } = useRecordings();
+  const [input, setInput] = useState('');
+  const { data: recordings } = useRecordings(input);
   const { refetch: handleRecordingReSync } = useRecordingsReSync();
 
-  if (isLoading) return <Spinner />;
```
farhatahmad, line 17: Why are you dropping the `isLoading`?

app/javascript/components/recordings/Recordings.jsx
scouillard: Not a fan of using tables, personally I would use rows here.

I'd like to keep this as a component as it is re-used in RoomRecordings.

app/javascript/hooks/queries/recordings/useRecordings.jsx
scouillard: Would "query" be better?

app/javascript/components/recordings/Recordings.jsx
scouillard: Right now, the query is fired onChange and not onSubmit.
Not sure this is a good design as this will trigger a isLoading on every character.
The Spinner is really annoying.
Maybe we could solve this with placeholders.
We could talk about this in more details.

app/javascript/hooks/queries/recordings/useRecordings.jsx
farhatahmad: `search` would be best

app/javascript/components/recordings/Recordings.jsx
farhatahmad: Gotcha

app/javascript/hooks/queries/recordings/useRecordings.jsx
hadicheaito1: "search" might be better

app/javascript/components/recordings/Recordings.jsx
scouillard: Right.

app/javascript/hooks/queries/recordings/useRecordings.jsx
hadicheaito1: Question: What is making the query "re-fire" each time a letter is typed? Is it when the onChange changes the value of the query param, it automatically re-fires the query since the param changed?

scouillard: It's the input used as a query-key.
Query will fire every time the input, the key, changes.

scouillard: useQuery(['getRecordings', input], ...)

hadicheaito1: Oh okay cool I see it now :+1:

hadicheaito1 (dismissed):

spec/controllers/recordings_controller_spec.rb
```diff
@@ -27,6 +27,24 @@
...
+          course_recordings << recording
+        end
+      end
+
+      create(:room, user:, recordings:)
+
+      get :index, params: { search: course_name }
+      response_recording_ids = JSON.parse(response.body)['data'].map { |recording| recording['id'] }
+      expect(response_recording_ids).to eq(course_recordings.pluck(:id))
+    end
```
farhatahmad, line 47: I would take this a step further - manually create some recordings with some set names and then manually set the search param to a string and make sure the recordings returned are correct

app/models/recording.rb
```diff
@@ -9,4 +9,10 @@ class Recording < ApplicationRecord
   validates :visibility, presence: true
   validates :length, presence: true
   validates :users, presence: true
+
+  def self.search(input)
+    return where('recordings.name ILIKE ?', "%#{input}%").includes(:formats) if input
+
+    all.includes(:formats)
+  end
```
farhatahmad, line 17: You should add atleast 2 or 3 tests for this

PR #3479, opened 2022-05-27
Refactored meeting start and implement the room settings in BBB

farhatahmad:
## Testing Steps

1. Toggle the 3 BBB room settings
2. Start meeting / join meeting and make sure the settings are reflected in BBB


hadicheaito1 (approved):

scouillard:
Should we send an alert if the Admin updates the RoomSetting and the meeting is on-going?
Something like "The updated settings will take effect when the room is restarted" ?

farhatahmad:
We could look at adding something like that as a disclaimer in the room settings. I'll check with Tyler and see what he thinks

scouillard (approved):

PR #3482, opened 2022-05-30
Added external auth through openid_connect

farhatahmad:


scouillard:
If I login with my Gmail account via Submit button, and then logout, pressing submit again will re-log me back in.
I can't go back to the Gmail interface, and switch account for example.

farhatahmad:
> If I login with my Gmail account via Submit button, and then logout, pressing submit again will re-log me back in. I can't go back to the Gmail interface, and switch my gmail account for example.

@scouillard Yeah that's expected behaviour. Keycloak (and google) both cache the request. The only way around it is to go incognito

scouillard (approved):

PR #3490, opened 2022-06-01
Setup basic framework of roles

farhatahmad:
## Description

A user must have a role set before they are created

## Testing Steps

1. Run db:migrate:with_data
2. Create a new user
3. Make sure the user is assigned the user role

### DIFF
--- a/spec/factories/user_factory.rb
+++ b/spec/factories/user_factory.rb
@@ -7,7 +7,7 @@ FactoryBot.define do
     external_id { nil }
     provider { Faker::Company.name.downcase }
     password { Faker::Internet.password }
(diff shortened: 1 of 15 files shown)


spec/factories/user_factory.rb
```diff
@@ -7,7 +7,7 @@
     external_id { nil }
     provider { Faker::Company.name.downcase }
     password { Faker::Internet.password }
-    password_confirmation { password } # For password confirmation validation.
```
farhatahmad, line 10: Not needed anymore

hadicheaito1 (approved):

PR #3493, opened 2022-06-01
Back-end hcaptcha check

hadicheaito1:
## Description

- Only tries to verify token if hcaptcha is enabled
- Checks if hcaptcha is enabled by checking if the env variables are set
## Testing Steps

- Set both environment variables (Secret and site key)
- Try to sign in and sign up without clicking hcaptcha checkbox (it shouldn't work)
- Now remove one of the env variables and refresh server
- Try to sign in and sign up without clicking the hcaptcha checkbox (it should work)

### DIFF
--- a/app/controllers/api/v1/sessions_controller.rb
+++ b/app/controllers/api/v1/sessions_controller.rb
@@ -31,7 +31,13 @@ module Api
         user = User.find_by(email: session_params[:email])
 
         # TODO: Add proper error logging for non-verified token hcaptcha
-        if user.present? && verify_hcaptcha(response: params[:token]) && user.authenticate(session_params[:password])
+        if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
+          render json: { signed_in: false }, status: :bad_request
+          return
+        end
+
+        # TODO: Add proper error logging for non-verified token hcaptcha
+        if user.present? && user.authenticate(session_params[:password])
           sign_in user
           render json: { current_user:, signed_in: true }, status: :ok
         else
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1/sessions_controller.rb
```diff
@@ -31,7 +31,13 @@ def create
         user = User.find_by(email: session_params[:email])
 
         # TODO: Add proper error logging for non-verified token hcaptcha
-        if user.present? && verify_hcaptcha(response: params[:token]) && user.authenticate(session_params[:password])
+        if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
+          render json: { signed_in: false }, status: :bad_request
+          return
+        end
```
farhatahmad, line 37: You might be able to do this in one line 
```
return render json: { signed_in: false }, status: :bad_request  if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
```

app/controllers/application_controller.rb
```diff
@@ -5,4 +5,9 @@ class ApplicationController < ActionController::Base
   def current_user
     @current_user ||= User.find_by(id: session[:user_id])
   end
+
+  # Returns whether hcaptcha is enabled by checking if ENV variables are set
+  def hcaptcha_enabled?
+    (ENV['RECAPTCHA_SITE_KEY'].present? && ENV['RECAPTCHA_SECRET_KEY'].present?)
+  end
```
farhatahmad, line 12: This is fine for now, but I'm thinking we move this to a concern called `EnvSettings` or something later on

farhatahmad (approved):

PR #3514, opened 2022-06-07
Backend: Added recording update name.

KH-Amir-TN:
## Description

Backend logic of handling recordings update name requests.

### DIFF
--- a/app/controllers/api/v1/recordings_controller.rb
+++ b/app/controllers/api/v1/recordings_controller.rb
@@ -4,6 +4,7 @@ module Api
   module V1
     class RecordingsController < ApiController
       skip_before_action :verify_authenticity_token # TODO: - Revisit this.
+      before_action :find_recording, only: :update
 
       # GET /api/v1/recordings.json
       # Returns: { data: Array[serializable objects(recordings)] , errors: Array[String] }
@@ -37,6 +38,19 @@ module Api
         render_json(status: :ok)
       end
 
+      # PUT/PATCH /api/v1/recordings/:recording_id.json
+      # Returns: { data: Array[serializable objects(recordings)] , errors: Array[String] }
+      # Does: Update the recording name.
+      def update
+        new_name = recording_params[:name]
+        return render_json errors: [Rails.configuration.custom_error_msgs[:missing_params]], status: :bad_request if new_name.blank?
+
+        BigBlueButtonApi.new.update_recordings record_id: @recording.record_id, meta_hash: { meta_name: new_name }
+        Recording.update! @recording.id, name: new_name
+
+        render_json
+      end
+
       def resync
         RecordingsSync.new(user: current_user).call
 
@@ -56,6 +70,16 @@ module Api
         BigBlueButtonApi.new.publish_recordings(record_ids: record_id, publish:)
 
         render_json(status: :ok)
+      end
+
+      private
+
+      def recording_params
+        params.require(:recording).permit(:name)
+      end
+
+      def find_recording
+        @recording = Recording.find_by! record_id: params[:id]
       end
     end
   end
--- a/app/services/big_blue_button_api.rb
+++ b/app/services/big_blue_button_api.rb
@@ -40,6 +40,10 @@ class BigBlueButtonApi
     bbb_server.publish_recordings(record_ids, publish)
   end
 
+  def update_recordings(record_id:, meta_hash:)
+    bbb_server.update_recordings(record_id, {}, meta_hash)
+  end
+
   private
 
   def bbb_endpoint
--- a/spec/controllers/recordings_controller_spec.rb
+++ b/spec/controllers/recordings_controller_spec.rb
@@ -53,6 +53,40 @@ RSpec.describe Api::V1::RecordingsController, type: :controller do
     end
   end
 
+  describe '#update' do
+    let(:room) { create(:room, user:) }
+    let(:recording) { create(:recording, room:) }
+
+    before do
+      allow_any_instance_of(BigBlueButtonApi).to receive(:update_recordings).and_return(http_ok_response)
+    end
+
+    it 'updates the recordings name with valid params returning :ok status code' do
+      expect_any_instance_of(BigBlueButtonApi).to receive(:update_recordings).with(record_id: recording.record_id,
+                                                                                   meta_hash: { meta_name: 'My Awesome Recording!' })
+
+      expect { post :update, params: { recording: { name: 'My Awesome Recording!' }, id: recording.record_id } }.to(change { recording.reload.name })
+      expect(response).to have_http_status(:ok)
+    end
+
+    it 'does not update the recordings name for invalid params returning a :bad_request status code' do
+      expect_any_instance_of(BigBlueButtonApi).not_to receive(:update_recordings)
+
+      expect do
+        post :update,
+             params: { recording: { name: '' }, id: recording.record_id }
+      end.not_to(change { recording.reload.name })
+
+      expect(response).to have_http_status(:bad_request)
+    end
+
+    it 'does not update the recordings name for invalid recording id returning :not_found status code' do
+      expect_any_instance_of(BigBlueButtonApi).not_to receive(:update_recordings)
+      post :update, params: { recording: { name: '' }, id: '404' }
+      expect(response).to have_http_status(:not_found)
+    end
+  end
+
   # TODO: - Uncomment once delete_recordings is no longer in destroy
   # describe '#destroy' do
   #   it 'deletes recording from the database' do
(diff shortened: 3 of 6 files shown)


app/services/recordings_updater.rb
```diff
@@ -0,0 +1,14 @@
...
+    @recording = recording
+    @meta_hash = {
+      'meta_recording-name': new_name
+    }
+  end
+
+  def call
+    BigBlueButtonApi.new.update_recordings record_id: @recording.record_id, meta_hash: @meta_hash
+  end
+end
```
farhatahmad, line 14: A service here is overkill - you can just call the API from the controller

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -37,6 +38,17 @@ def destroy
         render_json(status: :ok)
       end
 
+      # PUT/PATCH /api/v1/recordings/:recording_id.json
+      # Returns: { data: Array[serializable objects(recordings)] , errors: Array[String] }
+      # Does: Update the recording name.
+      def update
+        new_recording_name = recording_update_params[:name]
```
farhatahmad, line 45: `new_name` should be fine here

spec/controllers/recordings_controller_spec.rb
```diff
@@ -53,6 +53,102 @@
...
+                       endTime: 'Wed, 08 Jun 2022 08:42:38 +0100'.to_datetime,
+                       participants: '1',
+                       metadata: { 'bbb-origin': 'greenlight', 'bbb-origin-version': '3', isBreakout: 'false',
+                                   'bn-presenter-name': 'Presenter' },
+                       playback: { format: { type: 'presentation',
+                                             url: 'https://blindsidenetworks.com/',
+                                             length: 1 } } }],
+        messageKey: '',
+        message: '' }
+    end
```
farhatahmad, line 89: Move these to private methods (see: `recordings_sync_service`)

spec/controllers/recordings_controller_spec.rb
```diff
@@ -53,6 +53,102 @@
...
+              state: 'published',
+              startTime: 'Wed, 08 Jun 2022 08:41:56 +0100'.to_datetime,
+              endTime: 'Wed, 08 Jun 2022 08:42:38 +0100'.to_datetime,
+              participants: '1',
+              metadata: { 'bbb-origin': 'greenlight', 'bbb-origin-version': '3', isBreakout: 'false',
+                          'bn-presenter-name': 'Presenter', 'recording-name': 'Older' },
+              playback: { format: { type: 'presentation',
+                                    url: 'https://blindsidenetworks.com/',
+                                    length: 1 } } }
+          ]
```
farhatahmad, line 113: This isn't needed - we can just mock the response from the BigBlueButtonApi (See `recordings_sync_service`)

spec/controllers/recordings_controller_spec.rb
```diff
@@ -53,6 +53,102 @@
...
+      stub_const('BigBlueButtonApi::RECORDINGS_LIST', pre_update_recordings_list)
+    end
+
+    it 'updates the recordings name with valid params returning :ok status code' do
+      expect { get :resync }.to change(Recording, :count).from(0).to(BigBlueButtonApi::RECORDINGS_LIST[:recordings].size)
+      recording_id = Recording.pluck(:record_id).sample
+      post :update, params: { recording: recording_update_params, id: recording_id }
+      expect(response).to have_http_status(:ok)
+      get :resync
+      recording = Recording.find_by record_id: recording_id
```
farhatahmad, line 131: This is definitely way too complex. I think you should review the other tests and approach them in a similar way.

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -37,6 +38,17 @@ def destroy
...
+        return render_json errors: [Rails.configuration.custom_error_msgs[:missing_params]], status: :bad_request if new_recording_name.blank?
+
+        RecordingsUpdater.new(recording: @recording, new_name: new_recording_name).call
```
farhatahmad, line 48: You also need to update the local `Recording` object

app/services/big_blue_button_api.rb
```diff
@@ -40,6 +40,11 @@ def publish_recordings(record_ids:, publish:)
     bbb_server.publish_recordings(record_ids, publish)
   end
 
+  def update_recordings(record_id:, meta_hash:)
+    filtered_meta = meta_hash.slice :'meta_recording-name'
```
farhatahmad, line 44: What's the purpose of this?

spec/controllers/recordings_controller_spec.rb
KH-Amir-TN: The problem that I had was the fallowing:
  The update recording logic is a bit different where the source of truth isn't Greenlight but rather the **BBB/SL** server until a` r-sync` request gets issued to synchronize **GL**.
  That means upon testing and when issuing an update, **GL** will send the update API call to **BBB/SL** but nothing will reflect on our side except that the operation was successful meaning that **GL** copy of recordings will not include the update leaving me unable to make the functional test.
  I found that I need a way to mock both of the **API** calls that r-sync and update logics uses in a way that the mocks share a state that the update will change so that the second `r-sync` reflect the changes in **GL** ending up updating the table so that I can complete the test otherwise I'll be mocking the end result before making any updates.
  This was the cleanest way to mock the underlying implementation of `BBB_API `not only its returned value so that I can trigger a change on the `get_recordings `result via `update_recordings`. 
  -> I feel that there may be something a lot simpler than this but couldn't simply find it.

app/controllers/api/v1/recordings_controller.rb
KH-Amir-TN: Hmm, I didn't approach it that way.
In this case there will not be any need for a full r-sync.
This will help solve all of the issues and further simplify the tests as you have mentioned.
Thanks, I'll rectify the design!

spec/controllers/recordings_controller_spec.rb
KH-Amir-TN: Totally agree, after fallowing the new design this became absurd.

app/services/big_blue_button_api.rb
KH-Amir-TN: Thought for some reason that I should filter out any meta data other than `recording-name``but it's useless because we'll be filtering the params.
I'll remove it.

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -37,6 +38,19 @@ def destroy
...
       end
 
+      # PUT/PATCH /api/v1/recordings/:recording_id.json
+      # Returns: { data: Array[serializable objects(recordings)] , errors: Array[String] }
+      # Does: Update the recording name.
+      def update
+        new_name = recording_params[:name]
+        return render_json errors: [Rails.configuration.custom_error_msgs[:missing_params]], status: :bad_request if new_name.blank?
+
+        BigBlueButtonApi.new.update_recordings record_id: @recording.record_id, meta_hash: { 'meta_recording-name': new_name }
```
farhatahmad, line 48: I think to avoid confusion, we should call this `meta_name` - the same thing as v2, so that when we import the recordings from old rooms, their names don't change

spec/controllers/recordings_controller_spec.rb
```diff
@@ -53,6 +53,46 @@
...
+      {
+        name: 'My Awesome Recording!'
+      }
+    end
+
+    let(:invalid_recording_update_params) do
+      {
+        name: ''
+      }
+    end
```
farhatahmad, line 69: I would remove these and just move the logic into each test case - that way it is clear whats happening

spec/controllers/recordings_controller_spec.rb
```diff
@@ -53,6 +53,46 @@
...
+
+    let(:room) { create(:room, user:) }
+    let(:recording) { create(:recording, room:) }
+
+    before do
+      allow_any_instance_of(BigBlueButtonApi).to receive(:update_recordings).and_return(http_ok_response)
+    end
+
+    it 'updates the recordings name with valid params returning :ok status code' do
+      expect { post :update, params: { recording: recording_update_params, id: recording.record_id } }.to(change { recording.reload.name })
```
farhatahmad, line 79: We should expect `BigBlueButtonApi` to receive a call to update recordings here

spec/controllers/recordings_controller_spec.rb
```diff
@@ -53,6 +53,46 @@
...
+      expect(response).to have_http_status(:ok)
+    end
+
+    it 'does not update the recordings name for invalid params returning a :bad_request status code' do
+      expect { post :update, params: { recording: invalid_recording_update_params, id: recording.record_id } }.not_to(change do
+                                                                                                                        recording.reload.name
+                                                                                                                      end)
```
farhatahmad, line 86: This indentation is weird.

```
expect { 
  post :update, 
  params: { recording: invalid_recording_update_params, id: recording.record_id } 
}.not_to change { recording.reload.name }
```

KH-Amir-TN: I'm using an extension to correct my work according to our Rubocop config.
What you have provided Rubocop doesn't seem to like.
Do you approve this instead:
```
     expect do
        post :update,
             params: { recording: invalid_recording_update_params, id: recording.record_id }
      end.not_to(change { recording.reload.name })
```

farhatahmad (approved):

PR #3516, opened 2022-06-08
(feat) Add access code to Rooms

scouillard:
## Description

Add AccessCode to Rooms.
If an access code is generated by the Room admin, the User will be prompted to enter the access code before being able to join a room.

## Testing Steps

- Log in your GLv3 profile.
- Go into a Room, and click on the Copy button to generate a "Join Room URL".
- Open an incognito browser window.
- In the incognito window, navigate to the "Join Room URL".
- Switch back to the non-incognito window. 
- Go to Room/Room Settings and play with the Generate/Delete access code feature.
- Generating an access code should now instantly prompt the incognito User to enter the access code.

image

## TODO
- Verify Code Generation design w/ Tyler

### DIFF
--- a/app/controllers/api/v1/meetings_controller.rb
+++ b/app/controllers/api/v1/meetings_controller.rb
@@ -56,7 +56,7 @@ module Api
 
       # Returns true if the room is accessible to all or if the access code is correct
       def authorize_access?
-        @room.access_code.blank? || @room.access_code == params[:access_code]
+        @room.viewer_access_code.blank? || @room.viewer_access_code == params[:viewer_access_code]
       end
     end
   end
--- a/app/controllers/api/v1/meetings_controller.rb
+++ b/app/controllers/api/v1/meetings_controller.rb
@@ -6,7 +6,7 @@ module Api
       skip_before_action :verify_authenticity_token # TODO: amir - Revisit this.
       before_action :find_room, only: %i[start join status]
 
-      # POST /api/v1/rooms/:friendly_id/start.json
+      # POST /api/v1/meetings/:friendly_id/start.json
       # Returns: { data: Array[serializable objects] , errors: Array[String] }
       # Does: Starts the Room meeting and joins in the meeting starter.
       def start
@@ -25,25 +25,38 @@ module Api
         }, status: :created
       end
 
-      # GET /api/v1/rooms/:friendly_id/join.json
+      # GET /api/v1/meetings/:friendly_id/join.json
       def join
-        render_json(data: BigBlueButtonApi.new.join_meeting(room: @room, name: params[:name], role: 'Viewer'), status: :ok)
+        if authorize_access?
+          render_json data: BigBlueButtonApi.new.join_meeting(room: @room, name: params[:name], role: 'Viewer'), status: :ok
+        else
+          render_json status: :unauthorized
+        end
       end
 
-      # GET /api/v1/rooms/:friendly_id/status.json
+      # GET /api/v1/meetings/:friendly_id/status.json
       def status
         data = {
           status: BigBlueButtonApi.new.meeting_running?(room: @room)
         }
         data[:joinUrl] = BigBlueButtonApi.new.join_meeting(room: @room, name: params[:name], role: 'Viewer') if data[:status]
 
-        render_json(data:, status: :ok)
+        if authorize_access?
+          render_json data:, status: :ok
+        else
+          render_json status: :unauthorized
+        end
       end
 
       private
 
       def find_room
         @room = Room.find_by!(friendly_id: params[:friendly_id])
+      end
+
+      # Returns true if the room is accessible to all or if the access code is correct
+      def authorize_access?
+        @room.access_code.blank? || @room.access_code == params[:access_code]
       end
     end
   end
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -4,7 +4,7 @@ module Api
   module V1
     class RoomsController < ApiController
       skip_before_action :verify_authenticity_token # TODO: amir - Revisit this.
-      before_action :find_room, only: %i[show update recordings recordings_processing purge_presentation]
+      before_action :find_room, only: %i[show update recordings recordings_processing purge_presentation access_code remove_access_code]
 
       include Avatarable
       include Presentable
@@ -39,14 +39,15 @@ module Api
         render_json data: rooms + shared_rooms, status: :ok
       end
 
+      # GET /api/v1/rooms/:friendly_id.json
       def show
-        # TODO: samuel - should probably create/use a UUID instead of object id
         room = {
           id: @room.id,
           name: @room.name,
           presentation_name: presentation_name(@room),
           thumbnail: presentation_thumbnail(@room),
-          created_at: @room.created_at.strftime('%A %B %e, %Y %l:%M%P')
+          created_at: @room.created_at.strftime('%A %B %e, %Y %l:%M%P'),
+          access_code: @room.access_code
         }
         if params[:include_owner] == 'true'
           room[:owner] = {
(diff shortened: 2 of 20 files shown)


Gemfile.lock
```diff
@@ -371,3 +371,9 @@ DEPENDENCIES
   tzinfo-data
   web-console
   webdrivers
+
+RUBY VERSION
+   ruby 3.1.0p0
+
+BUNDLED WITH
+   2.3.9
```
scouillard, line 379: Unsure what is causing this.
I've tried deleting the .lock, and copying the Gemfile from v3. 
It still does this.

app/javascript/components/rooms/RoomSettings.jsx
```diff
@@ -32,7 +43,26 @@ export default function RoomSettings() {
           <Row>
             <Col className="border-end border-2">
               <h6 className="text-primary">Room Name</h6>
+
+              <Button
+                variant="primary-light"
+                // Generates the access code: a string of six characters
+                onClick={() => setAccessCode(Math.random().toString(36).slice(2, 8))}
```
scouillard, line 49: Will change this to do a Mutation that generates a (new) Access Code via backend onClick.

farhatahmad:
We should be calling this `viewer_access_code` or `moderator_access_code` since we'll need a way to differentiate them in the future. Just calling it `access_code` means that you'll need to change it anyways in the next PR

app/javascript/components/rooms/RoomJoin.jsx
```diff
@@ -58,18 +63,39 @@ export default function RoomJoin() {
                 </Row>
               </div>
             ) : (
-              <div className="text-center">
-                <InputGroup className="mt-3">
+              <>
+                { room?.access_code?.length > 0
```
farhatahmad, line 67: Can we not just check `room.access_code`?

app/javascript/components/rooms/RoomSettings.jsx
```diff
@@ -32,6 +38,25 @@ export default function RoomSettings() {
           <Row>
             <Col className="border-end border-2">
               <h6 className="text-primary">Room Name</h6>
+              <Button
+                variant="primary-light"
+                onClick={() => {
+                  handleGenerateAccessCode();
+                }}
```
farhatahmad, line 45: You can just do `onClick={handleGenerateAccessCode}`

app/javascript/components/rooms/RoomSettings.jsx
```diff
@@ -32,6 +38,25 @@ export default function RoomSettings() {
...
+              >
+                Generate
+              </Button>
+              {
+                room?.access_code
+              }
+              <Button
+                variant="danger"
+                onClick={() => {
+                  handleDeleteAccessCode();
```
farhatahmad, line 55: Same as above

app/javascript/hooks/queries/rooms/useRoomJoin.jsx
```diff
@@ -2,20 +2,26 @@ import { useQuery } from 'react-query';
 import axios from 'axios';
 import subscribeToRoom from '../../../channels/rooms_channel';
 
-export default function useRoomJoin(friendlyId, name) {
+export default function useRoomJoin({
+  friendlyId, name, accessCode,
+}) {
```
farhatahmad, line 7: Why convert the arguments here to an object?

db/data/20220421162907_populate_meeting_options.rb
```diff
@@ -14,9 +14,7 @@ def up
       { name: 'guestPolicy', default_value: 'ALWAYS_ACCEPT' }, # ALWAYS_ACCEPT | ALWAYS_DENY | ASK_MODERATOR
       # GL only options:
       { name: 'glAnyoneCanStart', default_value: 'false' }, # true | false
-      { name: 'glAnyoneJoinAsModerator', default_value: 'false' }, # true | false
-      { name: 'glModeratorAccessCode', default_value: '' },
-      { name: 'glAttendeeAccessCode', default_value: '' }
```
farhatahmad, line 19: We need to leave these (as we discussed in our call)

app/javascript/components/rooms/RoomSettings.jsx
scouillard: It seems to need the arrow function to work.

scouillard: > We should be calling this `viewer_access_code` or `moderator_access_code` since we'll need a way to differentiate them in the future. Just calling it `access_code` means that you'll need to change it anyways in the next PR

I thought we could get away with `access_code` and `mod_access_code` but I agree

scouillard:
> We should be calling this `viewer_access_code` or `moderator_access_code` since we'll need a way to differentiate them in the future. Just calling it `access_code` means that you'll need to change it anyways in the next PR

I thought we could get away with `access_code` and `mod_access_code` but I agree

app/controllers/api/v1/meetings_controller.rb
```diff
@@ -19,19 +19,27 @@ def start
         }, status: :created
       end
 
-      # GET /api/v1/rooms/:friendly_id/join.json
+      # GET /api/v1/meetings/:friendly_id/join.json
       def join
-        render_json(data: BigBlueButtonApi.new.join_meeting(room: @room, name: params[:name], role: 'Viewer'), status: :ok)
+        if @room.access_code.blank? || @room.access_code == params[:access_code]
```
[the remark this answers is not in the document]
scouillard: Created room#authorize_access? method

app/javascript/components/rooms/RoomSettings.jsx
farhatahmad: We've done that before in other files (check `recordingResync`)

scouillard: It doesn't seem to be the same thing.
In recordingResync, the refetch method returned from the useQuery is renamed to handleRecordingResync.
In here, the handleGenerateAccessCode is the mutation itself.
Still investigating this.

spec/controllers/meetings_controller_spec.rb
```diff
@@ -83,6 +83,20 @@
       expect_any_instance_of(BigBlueButtonApi).to receive(:join_meeting).with(room:, name: user.name, role: 'Viewer')
       get :join, params: { friendly_id: room.friendly_id, name: user.name }
     end
+
+    it 'returns ok if the access code is right' do
```
hadicheaito1, line 87: I think the description would be a bit better if it were something like `allows join when access code is correct` and `denies join when access code is incorrect`.

Same applies for the rest. I just think it's easier for someone who is reading the tests to understand exactly what it is testing vs just saying "returns ok if ...".

Just my opinion, lmk if you agree or not

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -39,13 +39,14 @@ def index
...
+      # GET /api/v1/rooms/:friendly_id.json
       def show
-        # TODO: samuel - should probably create/use a UUID instead of object id
         room = {
           id: @room.id,
           name: @room.name,
           presentation: room_presentation(@room),
-          created_at: @room.created_at.strftime('%A %B %e, %Y %l:%M%P')
+          created_at: @room.created_at.strftime('%A %B %e, %Y %l:%M%P'),
+          viewer_access_code: @room.viewer_access_code
```
farhatahmad, line 49: The front end shouldn't have access to the `access_code` because that would make it possible for the user to see it by inspecting the network tab. We should just return true or false if either of the access codes is present

farhatahmad: Hmmm that's a small problem, since this query is used in the room join as well

farhatahmad: 2 potential solutions:
- Create a new endpoint for access codes
- Bundle along the access code with the RoomSettings query

scouillard: I will add a new endpoint to GET access codes. Thank you.

app/controllers/api/v1/meetings_controller.rb
```diff
@@ -25,26 +25,39 @@ def start
...
       end
 
       private
 
       def find_room
         @room = Room.find_by!(friendly_id: params[:friendly_id])
       end
+
+      # Returns true if the room is accessible to all or if the access code is correct
+      def authorize_access?
```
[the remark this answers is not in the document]
scouillard: Please do not put much thoughts into this as this is being rebuilt for the moderator_access_code addition.

hadicheaito1 (dismissed):

farhatahmad:
@scouillard Some more conflicts - I'll hold off on merging anything else

farhatahmad (approved):

PR #3521, opened 2022-06-09
Frontend: Added update language to users on client side.

KH-Amir-TN:
## Description

--
This PR completes **3.** & **4.**
--
~2. Create/Extend an action on the Users controller to enable the API to set the user language.~
~3. Populate the language select element with the available languages for users on the profile section in the client side.~
~4. Enable profile update requests to use GL API to update the user details including its selected language.~
5. Add required action callbacks that:
   5.1. Upon each authenticated request extracts the user chosen language from its record if it's available in GL.
   5.2. If the user have not chose any language then GL should infer its preferred language from the browser HTTP `HTTP_ACCEPT_LANGUAGE` headers.
  5.3 If the language still not available then GL will default to the configured  `default_locale`.
6. Enable `default_locale` and maybe `available_locales` simple configuration  via environmental variables   or in clear config file.
7. Add dictionaries for GL API and localize responses accordingly.
8. Add a **I18n** lib  for GL frontend app, create dictionaries and localize the UI with GL API as the source of truth for the user chosen language.
 

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/controllers/api/v1/sessions_controller.rb
+++ b/app/controllers/api/v1/sessions_controller.rb
@@ -18,7 +18,11 @@ module Api
               email: current_user.email,
               provider: current_user.provider,
               avatar: user_avatar(current_user),
+              lang: current_user.lang,
               signed_in: true
+            },
+            server_state: {
+              available_locales: Rails.configuration.available_locales_hash
             }
           }
         else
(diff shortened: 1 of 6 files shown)


KH-Amir-TN:
DEPENDS ON #3519

PR #3522, opened 2022-06-10
Localazation 2: Extended user_controller action to infer language on creation.

KH-Amir-TN:
## Description

This PR completes **2.**
--
~~2. Create/Extend an action on the Users controller to enable the API to set the user language:
   2.2. GL should infer the language from the browser HTTP `ACCEPT_LANGUAGE` header and set it in its record upon creation.
  2.3 If the language cannot be inferred then GL will default to the configured  `default_locale`.~~

3. Enable `default_locale`  to be initialized via environmental variables.
4. Upon each authenticated request extracts the user chosen language from its record.
5. Populate the language select element on the profile section in the client side with static list of languages.
6. Enable profile update requests to use GL API to update the user details including its selected language.
7. Add a **I18n** lib  for GL frontend app, create dictionaries and localize the UI according to the user chosen language.
8. The language select menu should include the list of available languages in the client app.

### User story [Creating an account and Updating the language]:
---
1. User creates an account.
2. GL will infer the language code from the HTTP request with a fallback to the `default_locale` that will be initialized from an ENV variable `DEFAULT_LOCALE` defaulting to **'en'**.
3. User gets their account created if it passes validation with the language set to what GL has chosen.
4. User will have the UI localized to what GL inferred if the language is available on the client side I18n config with a fallback to **'en'**  _(NO I18n on API side, only client side I18n will be used in GLv3)._.
6. User goes to the profile section
7. User finds a select menu full of the available languages on the client app according to its config.
8. User will have their inferred language selected.
9. User can select another language and hit the update button.
10. The client app will send the update request to persist the user chosen language in GL db while updating the session context for the user changing the UI localization to the chosen language.

## Testing Steps

> 1. Pull the code.


KH-Amir-TN:
DEPENDS ON #3518

app/controllers/external_controller.rb
```diff
@@ -4,12 +4,12 @@ class ExternalController < ApplicationController
   def create
     credentials = request.env['omniauth.auth']
     user_info = credentials['info']
-
     user = User.find_or_create_by(email: user_info['email']) do |u|
       u.external_id = credentials['uid']
       u.name = user_info['name']
       u.provider = 'greenlight'
       u.role = Role.find_by(name: 'User') # TODO: - Ahmad: Move to service
+      u.language = user_info['locale'].try(:scan, /^[a-z]{2}/)&.first || I18n.default_locale
```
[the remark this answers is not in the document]
farhatahmad: Yes preferably

app/controllers/api/v1/users_controller.rb
```diff
@@ -4,7 +4,7 @@ module Api
   module V1
     class UsersController < ApiController
       skip_before_action :verify_authenticity_token # TODO: amir - Revisit this.
-
+      before_action :default_language, only: :create
```
farhatahmad, line 7: This shouldn't be done in the before action - it should be done directly in the create action

spec/requests/users_controller_spec.rb
```diff
@@ -31,6 +33,26 @@
         post api_v1_users_path, params: valid_user_params, headers: headers
         expect(User.find_by(email: valid_user_params[:user][:email]).role).to eq(role)
       end
+
+      context 'User language' do
+        before do
+          allow(I18n).to receive(:default_locale).and_return(:gl)
```
farhatahmad, line 39: I would move this into the second test since it is only needed for a specific test case

spec/requests/users_controller_spec.rb
```diff
@@ -31,6 +33,26 @@
...
+        end
+
+        it 'Persists the user language in the user record' do
+          expect { post api_v1_users_path, params: valid_user_params, headers: }.to change(User, :count).from(0).to(1)
+          expect(response).to have_http_status(:created)
+          expect(User.take.language).to eq('en')
```
farhatahmad, line 45: Just for extra safety, I would do `expect(User.find_by(email: valid_user_params[:user][:email]).language).to eq('en')`

spec/requests/users_controller_spec.rb
KH-Amir-TN: But the other tests will then run with unmacked `I18n` instance that may default to the same language in the `user_params`?

KH-Amir-TN: What about changing the language in the `user_params` to a none language Alpha2 code like 'language'?

farhatahmad: How would `I18n.default_locale` be set to anything?

spec/requests/users_controller_spec.rb
```diff
@@ -31,6 +33,24 @@
...
+        end
+
+        it 'defaults user language to default_locale if the language isn\'t specified' do
+          allow(I18n).to receive(:default_locale).and_return(:gl)
+          expect(I18n).to receive(:default_locale)
+
+          valid_user_params[:user][:language] = nil
+          expect { post api_v1_users_path, params: valid_user_params, headers: }.to change(User, :count).from(0).to(1)
+          expect(response).to have_http_status(:created)
+          expect(User.take.language).to eq('gl')
```
farhatahmad, line 51: Same as above with the `User.find_by`

spec/requests/users_controller_spec.rb
```diff
@@ -31,6 +33,24 @@
         post api_v1_users_path, params: valid_user_params, headers: headers
         expect(User.find_by(email: valid_user_params[:user][:email]).role).to eq(role)
       end
+
+      context 'User language' do
+        it 'Persists the user language in the user record' do
+          expect { post api_v1_users_path, params: valid_user_params, headers: }.to change(User, :count).from(0).to(1)
+          expect(response).to have_http_status(:created)
+          expect(User.find_by(email: valid_user_params[:user][:email]).language).to eq('en')
```
farhatahmad, line 41: We can cut down on the code duplication a bit here

```
post api_v1_users_path, params: valid_user_params, headers:
expect(User.find_by(email: valid_user_params[:user][:email]).language).to eq('en')
```

spec/requests/users_controller_spec.rb
farhatahmad: and with the duplication
```
allow(I18n).to receive(:default_locale).and_return("test-language")
post api_v1_users_path, params: valid_user_params, headers:
expect(User.find_by(email: valid_user_params[:user][:email]).language).to eq('test-language')
```

spec/requests/users_controller_spec.rb
KH-Amir-TN: Rails will set it by default to `:en` if the `DEFAULT_LOCALE` isn't specified.
If the ENV VAR is specified then the `I18n.default_locale` will match it according to the `locales` initializer introduces in #3523.

farhatahmad: In that case, I would change the param coming in to be something that's not the standard default value `fr` or `es` should be fine

KH-Amir-TN: Good idea, I'll change it

spec/requests/users_controller_spec.rb
KH-Amir-TN: But that means that we won't be ensuring an adequate response code and that the user in the database may not be created by the `post api_v1_users_path` or is it normal to be omitted?

spec/requests/users_controller_spec.rb
KH-Amir-TN: But, that means that we won't be ensuring an adequate response code and that the user in the database may not be created by the `post api_v1_users_path` or is it normal to be omit this?

spec/requests/users_controller_spec.rb
KH-Amir-TN: Here's what I thought when developed the tests:
>I wanted to ensure that setting the language won't prevent the creation of a user (that's why I was testing the creation) and the response should be as expected for the client app (that's why I was testing the response status code).

spec/requests/users_controller_spec.rb
farhatahmad: The other tests already cover the user creation case, so in this case, it makes more sense just to simplify it down to ensuring the language is set correctly

farhatahmad (approved):

PR #3526, opened 2022-06-10
Pre-upload presentation backend + rpsec testing

hadicheaito1:
## Description

- Passes in url to back-end bbb server on start meeting
- Adjusted existing tests to work with new changes (adding url param)
- Added back-end tests for rooms update to test presentation
## Testing Steps

- Pull changes
- Create room
- Hardcode url in 'meetings_controller' to any url such as `https://riders.drivemag.com/wp-content/uploads/default/0002/02/kawasaki-2327.jpeg`
- Start meeting

NOTE: This is the only way to test it locally since pre-upload presentation doesn't work unless the domain you are on is accessible outside of your computer. Normally, you'd be able to upload anything and then start the meeting from within GL.

### DIFF
--- a/spec/controllers/meetings_controller_spec.rb
+++ b/spec/controllers/meetings_controller_spec.rb
@@ -11,7 +11,7 @@ RSpec.describe Api::V1::MeetingsController, type: :controller do
   end
 
   describe '#start' do
-    let(:room) { create(:room, user:, presentation: fixture_file_upload(file_fixture('default-avatar.png'), 'image/png')) }
+    let(:room) { create(:room, user:) }
 
     before do
       allow_any_instance_of(BigBlueButtonApi)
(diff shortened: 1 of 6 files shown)


spec/controllers/meetings_controller_spec.rb
```diff
@@ -22,8 +22,9 @@
     it 'makes a call to the MeetingStarter service with the right values' do
       logout = 'http://example.com'
       request.env['HTTP_REFERER'] = logout
+      url = nil
 
-      expect(MeetingStarter).to receive(:new).with(room:, logout_url: logout)
+      expect(MeetingStarter).to receive(:new).with(room:, logout_url: logout, url:)
 
       post :start, params: { friendly_id: room.friendly_id }
```
farhatahmad, line 29: You're missing a test case here. You need to test the case where you actually attach a real file to a room and make sure it gets passed through

PR #3528, opened 2022-06-13
Preupload presentation Front End click to upload

hadicheaito1:
## Description

- Front end change, now can click on text to upload file
- Added accepted formats

image

### DIFF
--- a/app/assets/stylesheets/presentation.scss
+++ b/app/assets/stylesheets/presentation.scss
@@ -1,0 +1,3 @@
+.presentation-upload {
+  cursor: pointer;
+}
--- a/app/javascript/components/rooms/Presentation.jsx
+++ b/app/javascript/components/rooms/Presentation.jsx
@@ -23,7 +23,7 @@ export default function Presentation() {
               <label htmlFor="file" className="presentation-upload">Click to Upload
                 <input
                   id="file"
-                  style={{ display: 'none' }}
+                  className="d-none"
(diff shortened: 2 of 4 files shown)


app/assets/stylesheets/presentation.scss
```diff
@@ -0,0 +1,3 @@
+.presentation-upload {
+  cursor: pointer;
```
farhatahmad, line 2: This is fine for now but we really should add a class called `pointer` to avoid having to do this

app/javascript/components/rooms/Presentation.jsx
```diff
@@ -20,16 +19,21 @@ export default function Presentation() {
             <div className="user-icon-circle rounded-circle d-block mx-auto mb-3">
               <FontAwesomeIcon icon={faCloudArrowUp} className="fa-4x text-primary d-block mx-auto pt-3" />
             </div>
-            <Card.Title className="text-primary"> Click to upload or drag and drop </Card.Title>
+            <Card.Title className="text-primary">
+              <label htmlFor="file" className="presentation-upload">Click to Upload
+                <input
+                  id="file"
+                  style={{ display: 'none' }}
```
farhatahmad, line 26: We shouldn't be doing inline styles like this. Check for a bootstrap class that can do this for you

farhatahmad (approved):

PR #3531, opened 2022-06-14
Change Password 0: API

KH-Amir-TN:
## Description

Enable users to change their password.

This PR completes **0.**
--
~~0. Make an API for the change password that receives an object with the old and new passwords and ensures the fulfillment of the scenarios in 4. of the User story below.~~
1. Add the change password UI to the profile section (CSR, template).
2. Add the change password form ensuring the respect of 3. constraints.
3. Add a changePassword mutation that initiates a valid change password request to the API with
respect to the 5. criteria.

### User story [Changing a password]:
---
1. An authenticated user navigates to their profile section.
2. User selects the change password option.
3. User enters their Old password and New password while confirming the last one:
   Constraints: Password update request will not initiate until:
	1- All fields are filled.
	2- The new password and its confirmation matches.
	
4. A change password request will initiate to the API endpoint.
   Scenarios:
	1. If the old password authenticates the user then the new password hash will
	be saved.
	2. If the old password doesn't authenticate the user then the user password
	will remain the same and an adequate error will be sent back the client.
   Note: If GLv3 include a password policy then the new password should conform to it.
5. The client UI will show feedbacks according to the API response.
   Note: For this increment we'll only log to the console appropriate messages for success
	and errors until notifications and feedbacks in GLv3 gets implemented properly.

## Testing Steps

> 1. Pull the code.
> 2. Install the dependencies `bundle install && npm|yarn install`.
> 3. Clean the previous assets build by running `rm app/assets/builds/*` (This won't remove .keep since it's hidden).
> 4. Clean the database and tmp files for a better isolation by running `rails tmp:clear && rails db:schema:cache:clear && rails db:drop && rails db:create && rails db:migrate:with_data`
> 5. Run the linter and specs `bundle exec rubocop --parallel && bundle exec rspec && npx eslint app/javascript/* --ext .jsx,.js`
> 6. Run `./bin/dev` to run the assets builders processes and the Puma server all at once.

PR #3535, opened 2022-06-14
Reset Password 1: Forget Password API

KH-Amir-TN:
## Description

Enable users to reset their password.

This PR completes **1.** 
--
~~0. Add `reset_token_digest` string attribute and `reset_token_expires_at` to users model with no restrictions.~~
~~1. Make an API for the forget password that receives the user email:
	- Finds the user by email.
	- generate a secure random **UNIQUE** complex token.
	- saves the token digest in the user `reset_token_digest` and set the `reset_token_expires_at`.
	>NOTE: The `reset_token_expires_at `= Time.now + PERIOD(seconds) where PERIOD > 0
	, PERIOD by default will be 15mins in seconds.
	- Sends the user the reset email with the token  _[Upcoming milestones]_.~~

>Note: For security reasons only the digest will be saved in the database and the token will be created and sent on the run.

2. Make an API for the reset password that receives the token , the new password and the user email:
	- Takes the token and validate its authenticity and integrity.
	- Validate the token expiration.
	- Update the user password.
	- Remove the user reset token attributes _[If the user update password succeeds]_.
3. Create the forget password UI and integrate it with the forget password API.
4. Create the reset password UI and integrate it with the reset password API.

### User story [Reset password]:
---
1. User navigates to sign-in page.
2. User clicks the forget password.
3. User enters the forget password UI to be prompted for their email:

>    Scenarios:
> 	1. if the user email is correct then a reset email will be sent to the user while
> 	indicating to check their inbox or spam folder.
> 	2. if the user email is not found then no email will be sent to the user but the user
> 	**WILL STILL** be indicated as if the email was sent _[Obscurity]_.

4. User receives the link with the **token digest**.
5. User _[if accesses the link before expiration]_ will be prompted for a new password and its 
confirmation.
  >Constraints: No reset password API request will initiate until:
	1. All fields are filled.
	2. Password confirmation matches the new password.
	_Note: If GLv3 include a password policy then the new password should conform to it._
6. User will get redirected to login page with the proper feedbacks and/or notifications.
[OR]
6. User will get redirected to login page if something went wrong and authenticate 
with the new password on success.

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -23,11 +23,31 @@ class User < ApplicationRecord
 
   # TODO: samuel - ActiveStorage validations needs to be discussed and implemented.
   validate :avatar_validation
+  validates :reset_digest, uniqueness: true, on: :update
 
   def self.search(input)
     return where('name ILIKE ?', "%#{input}%") if input
 
     all
+  end
+
+  # Generates a token digest.
+  def self.generate_digest(token)
+    Digest::SHA2.hexdigest(token)
+  end
+
+  # Create a unique random token
+  def generate_unique_token
+    token = SecureRandom.alphanumeric(40)
+    digest = User.generate_digest(token)
+
+    update! reset_digest: digest, reset_sent_at: Time.current
+
+    token
+  rescue ActiveRecord::RecordInvalid
+    raise unless errors.attribute_names.include? :reset_digest
+
+    retry
   end
 
   private
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -20,6 +20,7 @@ RSpec.describe User, type: :model do
     # it { is_expected.to validate_presence_of(:password_confirmation) }
 
     it { is_expected.to validate_uniqueness_of(:email).scoped_to(:provider).case_insensitive }
+    it { is_expected.to validate_uniqueness_of(:reset_digest).on(:update) }
 
     context 'password confirmation' do
       it 'invalidate the record for mismatched password confirmation' do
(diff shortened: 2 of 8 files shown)


KH-Amir-TN:
# DEPENDS ON #3531

app/models/user.rb
```diff
@@ -30,6 +30,25 @@ def self.search(input)
     all
   end
 
+  # Generates a token digest.
+  def self.generate_digest(token)
+    Digest::SHA2.hexdigest(token)
+  end
```
farhatahmad, line 36: For now this is fine, though we may want to move this to a service in the future

spec/models/user_spec.rb
```diff
@@ -42,4 +42,30 @@
...
+      end
+    end
+  end
+
+  context 'static methods' do
+    describe '#generate_digest' do
+      it 'calls Digest::SHA2#hexdigest to generate a digest' do
+        allow(Digest::SHA2).to receive(:hexdigest).and_return(true)
+        expect(Digest::SHA2).to receive(:hexdigest).with('test')
+        described_class.generate_digest('test')
```
farhatahmad, line 67: I would do this a different way 

```
expect(described_class.generate_digest('test')).to eq(Digest::SHA2.hexdigest('test'))

```

app/models/user.rb
KH-Amir-TN: Yup, we may generate more secure tokens and crypto sign there digest

app/controllers/api/v1/users_controller.rb
```diff
@@ -88,6 +107,10 @@ def user_params
       def change_password_params
         params.require(:user).permit(:old_password, :new_password)
       end
+
+      def forget_password_params
+        params.require(:user).permit(:email)
+      end
```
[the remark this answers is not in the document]
KH-Amir-TN: Agreed,

farhatahmad (approved):

PR #3536, opened 2022-06-14
Preupload presentation Front-end (Preview existing presentation)

hadicheaito1:
## Description

- Front end changes to display current uploaded presentation attached to room
- Displays preview of representable files (images, videos, and pdf) and a file icon for non representable files 
## Testing Steps

- Pull code changes
- Create room
- Upload presentation to room
- Make sure room is displayed correctly

image

image


config/environments/development.rb
```diff
@@ -9,7 +9,8 @@
   # it changes. This slows down response time but is perfect for development
   # since you don't have to restart the web server when you make code changes.
   config.cache_classes = false
-
+  # TODO: Hadi - Need to discuss this...
+  config.active_storage.variant_processor = :mini_magick
```
hadicheaito1, line 13: This is needed in order for the `representation' to work in rails.
We could discuss this further

Gemfile.lock
```diff
@@ -373,3 +380,9 @@ DEPENDENCIES
   tzinfo-data
   web-console
   webdrivers
+
+RUBY VERSION
+   ruby 3.1.0p0
+
+BUNDLED WITH
+   2.3.10
```
farhatahmad, line 388: Should be removed

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -44,10 +44,10 @@ def show
         room = {
           id: @room.id,
           name: @room.name,
-          presentation: room_presentation(@room),
+          presentation_name: presentation_name(@room),
+          presentation_representation: presentation_representation(@room),
```
farhatahmad, line 48: I would call this `thumbnail` or something a little more understandable

app/controllers/concerns/presentable.rb
```diff
@@ -3,7 +3,13 @@
 module Presentable
   extend ActiveSupport::Concern
 
-  def room_presentation(room)
-    return url_for(room.presentation) if room.presentation.attached?
+  def presentation_name(room)
+    return room.presentation.filename if room.presentation.attached?
+  end
+
+  def presentation_representation(room)
```
farhatahmad, line 10: Same as above with thumbnail

app/controllers/concerns/presentable.rb
```diff
@@ -3,7 +3,13 @@
...
+    return unless room.presentation.attached? && room.presentation.representable?
```
farhatahmad, line 11: I prefer if we don't do `unless` with `||` or `&&`.

`return if !room.presentation.attached? && !room.presentation.representable?`

app/controllers/concerns/presentable.rb
```diff
@@ -3,7 +3,13 @@
...
+
+    rails_representation_url(room.presentation.representation(resize: ['500x500']).processed)
```
farhatahmad, line 13: Why are we resizing to 500x500 here? Shouldn't we just resize directly to the size we want?

config/environments/development.rb
farhatahmad: This is fine

Gemfile.lock
scouillard: We need to run bundle for minimagick, and this gets printed.

app/controllers/concerns/presentable.rb
hadicheaito1: It doesn't actually change the size of the picture right away, it just changes the resolution (I know it's weird).

I will change it to match the size that we are changing it to though in the `presentation.scss`

hadicheaito1:
> We get a broken thumbnail when uploading a PDF. Maybe we should explicitly render the FontAwesome icon unless it's a .pptx or an .png, .jpg? (for doc, pdf, etc) There is a FontAwesome pdf https://fontawesome.com/search?q=pdf&s=solid%2Cbrands
> 
> image

This is weird, I tried the same pdf and it works. We will chat later to see what the issue is

farhatahmad (approved):

PR #3537, opened 2022-06-15
Reset Password 2: Reset Password & verify token APIs

KH-Amir-TN:
## Description

Enable users to reset their password.

This PR completes **2.** 
--
~~0. Add `reset_token_digest` string attribute and `reset_token_expires_at` to users model with no restrictions.~~
~~1. Make an API for the forget password that receives the user email:
	- Finds the user by email.
	- generate a secure random **UNIQUE** complex token.
	- saves the token digest in the user `reset_token_digest` and set the `reset_token_expires_at`.
	>NOTE: The `reset_token_expires_at `= Time.now + PERIOD(seconds) where PERIOD > 0
	, PERIOD by default will be 15mins in seconds.
	- Sends the user the reset email with the token  _[Upcoming milestones]_.~~

~~>Note: For security reasons only the digest will be saved in the database and the token will be created and sent on the run.~~

~~2. Make an API for the reset password that receives the token , the new password and the user email:
	- Takes the token and validate its authenticity and integrity.
	- Validate the token expiration.
	- Update the user password.
	- Remove the user reset token attributes _[If the user update password succeeds]_.~~
	
3. Create the forget password UI and integrate it with the forget password API.
4. Create the reset password UI and integrate it with the reset password API.

### User story [Reset password]:
---
1. User navigates to sign-in page.
2. User clicks the forget password.
3. User enters the forget password UI to be prompted for their email:

>    Scenarios:
> 	1. if the user email is correct then a reset email will be sent to the user while
> 	indicating to check their inbox or spam folder.
> 	2. if the user email is not found then no email will be sent to the user but the user
> 	**WILL STILL** be indicated as if the email was sent _[Obscurity]_.

4. User receives the link with the **token digest**.
5. User _[if accesses the link before expiration]_ will be prompted for a new password and its 
confirmation.
  >Constraints: No reset password API request will initiate until:
	1. All fields are filled.
	2. Password confirmation matches the new password.
	_Note: If GLv3 include a password policy then the new password should conform to it._
6. User will get redirected to login page with the proper feedbacks and/or notifications.
[OR]
6. User will get redirected to login page if something went wrong and authenticate 
with the new password on success.

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/controllers/api/v1/reset_password_controller.rb
+++ b/app/controllers/api/v1/reset_password_controller.rb
@@ -4,6 +4,7 @@ module Api
   module V1
     class ResetPasswordController < ApiController
       skip_before_action :verify_authenticity_token
+      before_action :verify_reset_request, only: %i[reset verify]
 
       # POST /api/v1/reset_password.json
       # Expects: { user: {:email} }
@@ -22,6 +23,47 @@ module Api
         token = user.generate_reset_token!
 
         render_json data: { token: } # TODO: enable reset email sending.
+      end
+
+      # POST /api/v1/reset_password/reset.json
+      # Expects: { user: {:token, :new_password} }
+      # Returns: { data: Array[serializable objects] , errors: Array[String] }
+      # Does: Validates the token and reset the user password.
+
+      def reset
+        new_password = params[:user][:new_password]
+
+        return render_json status: :bad_request if new_password.blank?
+
+        # Invalidating token and changing the password.
+        # TODO: optimise this.
+        return render_json status: :internal_server_error unless @user.invalidate_reset_token
+
+        @user.update! password: new_password
+
+        render_json
+      end
+
+      # POST /api/v1/reset_password/verify.json
+      # Expects: { user: {:token} }
+      # Returns: { data: Array[serializable objects] , errors: Array[String] }
+      # Does: Validates the token.
+
+      def verify
+        render_json
+      end
+
+      private
+
+      def verify_reset_request
+        return render_json status: :bad_request if params[:user].blank?
+
+        token = params[:user][:token]
+
+        return render_json status: :bad_request if token.blank?
+
+        @user = User.verify_reset_token(token)
+        render_json status: :forbidden unless @user
       end
     end
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -2,6 +2,9 @@
 
 class User < ApplicationRecord
   MAX_AVATAR_SIZE = 3_000_000
+  # Reset token max validity period.
+  # It's advised to not increase this to more than 1 hour.
+  RESET_TOKEN_VALIDITY_PERIOD = 1.hour
 
   has_secure_password validations: false
 
@@ -32,9 +35,35 @@ class User < ApplicationRecord
     all
   end
 
+  # Verifies the token existence, fetches its user and validates its expiration
+  # and invalidates the user token if expired.
+
+  def self.verify_reset_token(token)
+    digest = generate_digest(token)
+
+    user = find_by reset_digest: digest
+    return false unless user
+
+    return false unless user.reset_sent_at
+
+    expired = reset_token_expired?(user.reset_sent_at)
+
+    if expired
+      user.invalidate_reset_token
+      return false
+    end
+
+    user
+  end
+
   # Generates a token digest.
   def self.generate_digest(token)
     Digest::SHA2.hexdigest(token)
+  end
+
+  # Checkes the expiration of a token.
+  def self.reset_token_expired?(sent_at)
+    Time.current > (sent_at.in(RESET_TOKEN_VALIDITY_PERIOD))
   end
 
   # Create a unique random reset token
@@ -64,6 +93,10 @@ class User < ApplicationRecord
     raise unless errors.attribute_names.include? :activation_digest
 
     retry
+  end
+
+  def invalidate_reset_token
+    update reset_sent_at: nil, reset_digest: nil # Remove expired/valid tokens.
   end
 
   private
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -53,7 +53,12 @@ Rails.application.routes.draw do
         end
       end
       resources :env, only: :index
-      resources :reset_password, only: :create
+      resources :reset_password, only: :create do
+        collection do
+          post '/reset', to: 'reset_password#reset'
+          post '/verify', to: 'reset_password#verify'
+        end
+      end
       resources :verify_account, only: :create
       resources :roles
     end
--- a/spec/controllers/reset_password_controller_spec.rb
+++ b/spec/controllers/reset_password_controller_spec.rb
@@ -42,4 +42,71 @@ RSpec.describe Api::V1::ResetPasswordController, type: :controller do
       expect(response).to have_http_status(:ok)
     end
   end
+
+  describe 'POST reset_password#reset' do
+    let(:valid_params) do
+      { new_password: 'Glv3IsAwesome!', token: 'ZekpWTPGFsuaP1WngE6LVCc69Zs7YSKoOJFLkfKu' }
+    end
+
+    it 'updates the found user by digest for valid params' do
+      user = create(:user, password: 'Test12345678+')
+      allow(User).to receive(:verify_reset_token).with(valid_params[:token]).and_return(user)
+
+      post :reset, params: { user: valid_params }
+      expect(response).to have_http_status(:ok)
+      expect(user.reload.authenticate(valid_params[:new_password])).to be_truthy
+      expect(user.reset_digest).to be_blank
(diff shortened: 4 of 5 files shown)


KH-Amir-TN:
# DEPENDS ON #3535

app/models/user.rb
```diff
@@ -50,6 +78,13 @@ def generate_unique_token
     retry
   end
 
+  def invalidate_token!
+    digest = reset_digest
+    update! reset_sent_at: nil, reset_digest: nil # Remove expired/valid tokens.
+    logger.info "User(id):#{id} reset password token with digest:#{digest} got removed."
```
farhatahmad, line 84: We definitely shouldn't be logging the digest, as it's considered senstiive data.

I would just `update! reset_sent_at: nil, reset_digest: nil` without any logging

config/application.rb
```diff
@@ -28,5 +28,9 @@ class Application < Rails::Application
       record_not_found: 'Record Not Found',
       server_error: 'Something Went Wrong'
     }
+
+    # Reset token max validity period.
+    # It's advised to not increase this to more than 15 minutes.
+    config.reset_token_validity_period = 15.minutes
```
farhatahmad, line 34: I would move this as a constant to the `User` model instead

app/controllers/api/v1/reset_password_controller.rb
```diff
@@ -25,6 +26,44 @@ def create
...
+      # Returns: { data: Array[serializable objects] , errors: Array[String] }
+      # Does: Validates the token and reset the user password.
+
+      def reset
+        new_password = params[:user][:new_password]
+
+        return render_json status: :bad_request if new_password.blank?
+
+        # Invalidating token and changing the password.
+        @user.invalidate_token!.update! password: new_password
```
farhatahmad, line 41: I wouldn't chain these just to be safe .

```
@user.update! password: new_password
@user.invalidate_token!

```

app/models/user.rb
```diff
@@ -31,11 +31,39 @@ def self.search(input)
...
   end
 
+  # Verifies the token existence, fetches its user and validates its expiration
+  # and invalidates the user token if expired.
+
+  def self.verify_token(token)
+    digest = generate_digest(token)
+
+    user = find_by reset_digest: digest
+    logger.info "Unable to find a user with digest: #{digest}." unless user
```
farhatahmad, line 41: Same as above

config/application.rb
farhatahmad: Also - 15mins is quite short. I've normally seen it set to around 1 hour

app/models/user.rb
KH-Amir-TN: You're right, I wasn't thinking about it.
Will change it.

KH-Amir-TN:
# DEPENDS ON #3555

app/models/user.rb
```diff
@@ -66,6 +97,13 @@ def generate_activation_token!
     retry
   end
 
+  def invalidate_reset_token!
+    sent_at = reset_sent_at
+    update! reset_sent_at: nil, reset_digest: nil # Remove expired/valid tokens.
+    logger.info "User(id):#{id} reset their password reset token(sent_at):#{sent_at}."
+    self
```
farhatahmad, line 104: Why are we returning `self` here?

app/models/user.rb
```diff
@@ -32,11 +35,39 @@ def self.search(input)
...
+
+    expired = reset_token_expired?(user.reset_sent_at)
+
+    logger.info "Token(sent_at):#{user.reset_sent_at} expired for user(id):#{user.id}." if expired
+
+    user.invalidate_reset_token! if expired
+
+    return if expired
+
+    user
```
farhatahmad, line 58: This should just return `true` or `false` - not raise any errors

app/models/user.rb
```diff
@@ -32,11 +35,36 @@ def self.search(input)
...
+
+    return unless user.reset_sent_at
+
+    expired = reset_token_expired?(user.reset_sent_at)
+
+    user.invalidate_reset_token! if expired
+
+    return if expired
+
+    user
```
farhatahmad, line 55: This returns `nil` and `user` not `true` and `false`

app/models/user.rb
```diff
@@ -32,11 +35,36 @@ def self.search(input)
     all
   end
 
+  # Verifies the token existence, fetches its user and validates its expiration
+  # and invalidates the user token if expired.
+
+  def self.verify_reset_token!(token)
```
farhatahmad, line 41: You can drop the `!`

app/models/user.rb
```diff
@@ -66,6 +94,10 @@ def generate_activation_token!
     retry
   end
 
+  def invalidate_reset_token!
+    update! reset_sent_at: nil, reset_digest: nil # Remove expired/valid tokens.
+  end
+
```
farhatahmad, line 100: Unless this needs to be reused, we can just move this logic into the method above

app/models/user.rb
KH-Amir-TN: You prefer it to be an explicit `true|false`?

app/models/user.rb
KH-Amir-TN: I'm using that as a convention just like with models'query methods.
It indicates that the method uses at least one query method that raises an exception.

I think since we want the validation operation to be silent using query methods that raises exceptions is prohibited.
Correct?

app/models/user.rb
KH-Amir-TN: It's  reused in `reset_password_controller#reset` action.

app/controllers/api/v1/reset_password_controller.rb
```diff
@@ -23,6 +24,46 @@ def create
...
+      def verify_reset_request
+        return render_json status: :bad_request unless params[:user]
+
+        token = params[:user][:token]
+
+        return render_json status: :bad_request if token.blank?
+
+        @user = User.verify_reset_token!(token)
+        render_json status: :forbidden unless @user
+      end
```
scouillard, line 66: I think that using unless and if's in the same block, back to back, makes it hard to follow (maybe it's just me).

For example, I would do `if params[:user].empty? `on line 58

Also, shouldn't we use validations for that kind of thing? If it's missing a param or if it's blank?
Or have another solution? It clutters the controller IMO.

KH-Amir-TN: Do you like using unless instead of if statements?

As for the missing it's intentional for security reasons to obscure the responses to avoid given information about sensitive operations such as the reset password.
For example if we return an error status code for an invalid token we're giving attackers a potential attack vector that can find a token in use in our system.
Another example would be on generating the token for a user that is found by email, returning an error like `:not_found` will put an endpoint to guess our users emails from the outside.

KH-Amir-TN: I think I misunderstood you, you meant we should be using the `ActionController::Parameters` API for params validation?

KH-Amir-TN: Do you like using `unless` instead of if statements?
As for the validations are you suggesting the use of the `ActionController::Parameters` API for params validation?

app/models/user.rb
KH-Amir-TN: This method will:
1. Generate a digest for the given the token.
2. Find a user with the corresponding digest.
3. Validate the expiration of the token.
For valid tokens the method should return the found user that's why we need it to return its record for valid tokens.
Making the method return an explicit `true` value will cause to do an extra step on the action:
1. Re-generate a digest for the given token.
2. Re-find the user by digest.
Which feels unnecessary since in the validation process it requires generating the digest and finding the user record which is guaranteed to be truthy.

app/controllers/api/v1/reset_password_controller.rb
```diff
@@ -23,6 +24,47 @@ def create
...
+      # Does: Validates the token and reset the user password.
+
+      def reset
+        new_password = params[:user][:new_password]
+
+        return render_json status: :bad_request if new_password.blank?
+
+        # Invalidating token and changing the password.
+        # TODO: optimise this.
+        return render_json status: :internal_server_error unless @user.invalidate_reset_token
```
KH-Amir-TN, line 40: Should we make the password reset happen even when unable to invalidate the reset token?
> Note: Doing that will give the user (any one with the token) under the validity period the ability to infinitely reset the password.

app/models/user.rb
```diff
@@ -66,6 +95,10 @@ def generate_activation_token!
     retry
   end
 
+  def invalidate_reset_token
+    update reset_sent_at: nil, reset_digest: nil # Remove expired/valid tokens.
```
KH-Amir-TN, line 99: We can bypass `ActiveRecord` validations for such operations.

farhatahmad (approved):

PR #3541, opened 2022-06-15
Preupload presentation Delete

hadicheaito1:
## Description

Deletes attached presentation
## Testing Steps

- Pull code
- Create room
- Add file to it
- Click on delete icon on presentation
- Presentation should be deleted

image

### DIFF
--- a/app/assets/stylesheets/presentation.scss
+++ b/app/assets/stylesheets/presentation.scss
@@ -6,3 +6,7 @@
   width: 225px;
   height: 125px;
 }
+
+.delete-presentation {
+  cursor: pointer;
+}
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -83,11 +83,8 @@ module Api
 
       def purge_presentation
         @room.presentation.purge
-        if @room.presentation.purge
-          render_json status: :ok
-        else
-          render_json errors: @room.errors.to_a, status: :bad_request
-        end
+
+        render_json status: :ok
       end
 
       # GET /api/v1/rooms/:friendly_id/recordings.json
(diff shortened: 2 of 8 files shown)


hadicheaito1:
> I would
> 
> * size down the presentation row
> * use this trash icon instead `import { faTrashAlt } from '@fortawesome/free-regular-svg-icons'`
> 
> So it looks more like the mock-ups

Personally I find the row being bigger looks nicer than the mock-ups. The mock-up presentation page looks too empty when it's thin and makes it look as if there can be more presentations added due to the big empty space. @farhatahmad what do you think?

The icon I will change :+1:

hadicheaito1:
> I'm not sure about having a Modal confirmation for deleting a presentation - it's annoying and not critical IMO.
> 
> I would also use this button instead (taken from SharedAccess Shared Users list) so it acts like a button when you hover it.
> 
> ```
>                     <Button
>                       variant="font-awesome"
>                       className="float-end pe-2"
>                       onClick={() => handleDeleteSharedAccess({ user_id: user.id })}
>                     >
>                       <FontAwesomeIcon icon={faTrashAlt} />
>                     </Button>
> ```
> 
> I made this template (variant="font-awesome") for those cases when we want to use FontAwesome icons as buttons.

Yeah I know what you mean, but I also find the modal good too because without it, making a mistake of deleting a presentation would be too easy and not recoverable. @farhatahmad opinion?

farhatahmad (dismissed):

farhatahmad:
@hadicheaito1 @scouillard We need modals for every single delete operation - there are many cases where people will accidentally click the button, so we need to make sure there is a confirmation for it. 

As for the row size, choose whatever you feel is nicer. No need to stick 1000% to the mockups

farhatahmad (approved):

PR #3553, opened 2022-06-20
Change Password: Restricted the operation to AuthN req only

KH-Amir-TN:
## Description

Enable users to change their password.

This PR rectifies **3.** to rectify the change password API design.
--
~~0. Make an API for the change password that receives an object with the old and new passwords and ensures the fulfillment of the scenarios in 4. of the User story below.~~
~~1. Add the change password UI to the profile section (CSR, template).~~
~~2. Add the change password form ensuring the respect of 3. constraints.~~
~~3. Add a changePassword mutation that initiates a valid change password request to the API with
respect to the 5. criteria.~~

### User story [Changing a password]:
---
1. An authenticated user navigates to their profile section.
2. User selects the change password option.
3. User enters their Old password and New password while confirming the last one:
   Constraints: Password update request will not initiate until:
	1- All fields are filled.
	2- The new password and its confirmation matches.
	
4. A change password request will initiate to the API endpoint.
   Scenarios:
	1. If the old password authenticates the user then the new password hash will
	be saved.
	2. If the old password doesn't authenticate the user then the user password
	will remain the same and an adequate error will be sent back the client.
   Note: If GLv3 include a password policy then the new password should conform to it.
5. The client UI will show feedbacks according to the API response.
   Note: For this increment we'll only log to the console appropriate messages for success
	and errors until notifications and feedbacks in GLv3 gets implemented properly.

## Testing Steps

> 1. Pull the code.

image
image

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -56,23 +56,24 @@ module Api
         render_json status: :ok
       end
 
-      # POST /api/v1/users/:id/change_password.json
+      # POST /api/v1/users/change_password.json
       # Expects: { user: { :old, :new } }
       # Returns: { data: Array[serializable objects] , errors: Array[String] }
       # Does: Validates and change the user password.
 
       def change_password
-        # TODO:  add a before_action callback to find_user.
-        user = User.find(params[:id])
+        return render_json status: :unauthorized unless current_user # TODO: generalise this.
+
+        return render_json status: :forbidden if current_user.external_id?
 
         old_password = change_password_params[:old_password]
         new_password = change_password_params[:new_password]
 
-        return render_json status: :bad_request if new_password.blank?
+        return render_json status: :bad_request if new_password.blank? || old_password.blank?
 
-        return render_json status: :unauthorized unless user.authenticate old_password
+        return render_json status: :bad_request unless current_user.authenticate old_password
 
-        user.update! password: new_password
+        current_user.update! password: new_password
         render_json
       end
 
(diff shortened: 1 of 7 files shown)


KH-Amir-TN:
The change password as an operation should be limited only to the authenticated user account `current_user`.
No user other then the authenticated one can change it even the administrators.
Administrators could enforce password reset but **NOT** password change so having an open endpoint scoped to the given user id param felt unnatural.

app/controllers/api/v1/users_controller.rb
```diff
@@ -56,23 +56,24 @@ def purge_avatar
...
-      # POST /api/v1/users/:id/change_password.json
+      # POST /api/v1/users/change_password.json
       # Expects: { user: { :old, :new } }
       # Returns: { data: Array[serializable objects] , errors: Array[String] }
       # Does: Validates and change the user password.
 
       def change_password
-        # TODO:  add a before_action callback to find_user.
-        user = User.find(params[:id])
+        return render_json status: :unauthorized unless current_user # TODO: generalise this.
```
farhatahmad, line 65: Admins should be able to change a user's password, so we shouldn't switch this to current_user

KH-Amir-TN: An admin can change a user password?
The `change_password` expects the current password and a new one.
If the admin can change any user password they need to provide their current password which is odd.
Or, Admins can reset any user password without any need to provide their current credentials.
You mean admins can reset user's password?
If that's the case isn't that a separate endpoint other then this one?

farhatahmad: Yeah sorry I mixed them up. An admin can send a reset password email

farhatahmad (approved):

farhatahmad (approved):

PR #3554, opened 2022-06-20
Account Verification 1: Account Activation Create API

KH-Amir-TN:
## Description

Enable users to activate their accounts.

This PR completes **1.** 
--
~~0. Add `active ` _boolean attribute, default to false_; `activation_digest` _unique string_; `activation_sent_at` _datetime_ to the `users` model.~~
~~1. Add `verify_account_controller#create` that:~~
   ~~- Creates a **UNIQUE** activation token.~~
   ~~- saves its digest and sent time to the _selected user_.~~
   ~~- Send the token to the selected user email _[Saved for upcoming sprints]_.~~
   
2. Extend/Update `users_controller#create `to trigger the `verify_account_controller#create` logic for the created user account.
3. Extend/Update `users_controller#update` to deactivate the user account if their email changes.
4. Add `verify_account_controller#activate` that:
   - Validates the given token.
   - activate its associated account _[if valid]_.
   - Invalidate the used/expired token.
5. Add resend verification UI page and hook the resend to `verify_account_controller#create` API.
6. Hook the UI app to `verify_account_controller#activate` API.

### User story [Account activation]:
---
1. User creates an account.
2. User receives an email with a link to activate their account.

_[If the link isn't used]:_
3. User sign-in.
3. User will have a UI that displayes a descriptive message indicating that they have to validate their accounts.
4. User click the resend button to receive a new activation link.
NOTE: User will be able to only update their profile, sign-in, sign-out, resend activation link until they activate their account. 

_[If the link is clicked while it's valid]:_
1. User will have an adequate feedback from the app.
2. User will be redirected to sign-in _[if not authenticated]_ or to rooms space _[if authenticated]_.

_[If the link is invalid]:_
1. User will have an adequate feedback from the app.
2. User will be redirected to homepage.

## Testing Steps

> 1. Pull the code.
> 2. Install the dependencies `bundle install && npm|yarn install`.
> 3. Clean the previous assets build by running `rm app/assets/builds/*` (This won't remove .keep since it's hidden).
> 4. Clean the database and tmp files for a better isolation by running `rails tmp:clear && rails db:schema:cache:clear && rails db:drop && rails db:create && rails db:migrate:with_data`
> 5. Run the linter and specs `bundle exec rubocop --parallel && bundle exec rspec && npx eslint app/javascript/* --ext .jsx,.js`
> 6. Run `./bin/dev` to run the assets builders processes and the Puma server all at once.

PR #3556, opened 2022-06-20
Verify account 2..4: Activation API

KH-Amir-TN:
## Description

Enable users to activate their accounts.

This PR completes **2. & 3. & 4.** 
--
~~0. Add `active ` _boolean attribute, default to false_; `activation_digest` _unique string_; `activation_sent_at` _datetime_ to the `users` model.~~
~~1. Add `verify_account_controller#create` that:~~
   ~~- Creates a **UNIQUE** activation token.~~
   ~~- saves its digest and sent time to the _selected user_.~~
   ~~- Send the token to the selected user email _[Saved for upcoming sprints]_.~~
   
~~2. Extend/Update `users_controller#create `to trigger the `verify_account_controller#create` logic for the created user account.~~
~~3. Extend/Update `users_controller#update` to deactivate the user account if their email changes.~~
~~4. Add `verify_account_controller#activate` that:~~
   ~~- Validates the given token.~~
   ~~- activate its associated account _[if valid]_.~~
   ~~- Invalidate the used/expired token.~~
5. Add resend verification UI page and hook the resend to `verify_account_controller#create` API.
6. Hook the UI app to `verify_account_controller#activate` API.

### User story [Account activation]:
---
1. User creates an account.
2. User receives an email with a link to activate their account.

_[If the link isn't used]:_
3. User sign-in.
3. User will have a UI that displayes a descriptive message indicating that they have to validate their accounts.
4. User click the resend button to receive a new activation link.
NOTE: User will be able to only update their profile, sign-in, sign-out, resend activation link until they activate their account. 

_[If the link is clicked while it's valid]:_
1. User will have an adequate feedback from the app.
2. User will be redirected to sign-in _[if not authenticated]_ or to rooms space _[if authenticated]_.

_[If the link is invalid]:_
1. User will have an adequate feedback from the app.
2. User will be redirected to homepage.

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -23,7 +23,8 @@ module Api
 
         if user.save
           session[:user_id] = user.id
-          render_json status: :created
+          token = user.generate_activation_token!
+          render_json data: { token: }, status: :created # TODO: enable activation email sending.
         else
           # TODO: amir - Improve logging.
           render_json errors: user.errors.to_a, status: :bad_request
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -5,6 +5,9 @@ class User < ApplicationRecord
   # Reset token max validity period.
   # It's advised to not increase this to more than 1 hour.
   RESET_TOKEN_VALIDITY_PERIOD = 1.hour
+  # Account activation token max validity period.
+  # It's advised to not increase this to more than 1 hour.
+  ACTIVATION_TOKEN_VALIDITY_PERIOD = 1.hour
 
   has_secure_password validations: false
 
@@ -97,6 +100,44 @@ class User < ApplicationRecord
 
   def invalidate_reset_token
     update reset_sent_at: nil, reset_digest: nil # Remove expired/valid tokens.
+  end
+
+  # Verifies the token existence, fetches its user and validates its expiration
+  # and invalidates the user activationtoken if expired.
+
+  def self.verify_activation_token(token)
+    digest = generate_digest(token)
+
+    user = find_by activation_digest: digest
+    return false unless user
+
+    return false unless user.activation_sent_at
+
+    expired = activation_token_expired?(user.activation_sent_at)
+
+    if expired
+      user.invalidate_activation_token
+      return false
+    end
+
+    user
+  end
+
+  # Checkes the expiration of a token.
+  def self.activation_token_expired?(sent_at)
+    Time.current > (sent_at.in(ACTIVATION_TOKEN_VALIDITY_PERIOD))
+  end
+
+  def invalidate_activation_token
+    update activation_sent_at: nil, activation_digest: nil # Remove expired/valid tokens.
+  end
+
+  def activate!
+    update! active: true
+  end
+
+  def deactivate!
+    update! active: false
   end
 
   private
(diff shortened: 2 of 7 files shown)


KH-Amir-TN:
# DEPENDS ON #3554

KH-Amir-TN:
# REBASED ON 6fb89357f600222a3e627e019f88da8220684373

app/controllers/api/v1/users_controller.rb
```diff
@@ -33,7 +34,10 @@ def create
       # TODO: Add a before_action callback to update,destory and purge_avatar calling a find_user.
       def update
         user = User.find(params[:id])
+        old_email = user.email
+        # TODO: optimise this.
         if user.update(user_params)
+          user.deactivate! if old_email != user.email
```
farhatahmad, line 40: Let's drop this from this PR since we need to discuss whether we need this or not

app/models/user.rb
```diff
@@ -66,6 +69,53 @@ def generate_activation_token!
...
+    expired = activation_token_expired?(user.activation_sent_at)
+
+    logger.info "Activation Token(sent_at):#{user.activation_sent_at} expired for user(id):#{user.id}." if expired
+
+    user.invalidate_activation_token! if expired
+
+    return if expired
+
+    user
+  end
```
farhatahmad, line 95: We should change this to return `true` or `false` - not raise errors

app/models/user.rb
```diff
@@ -66,6 +69,46 @@ def generate_activation_token!
...
+
+    return unless user.activation_sent_at
+
+    expired = activation_token_expired?(user.activation_sent_at)
+
+    user.invalidate_activation_token! if expired
+
+    return if expired
+
+    user
```
farhatahmad, line 89: Same as the reset PR

app/models/user.rb
```diff
@@ -66,6 +69,46 @@ def generate_activation_token!
...
+
+  def activate!
+    update! active: true
+    self
+  end
+
+  def deactivate!
+    update! active: false
+    self
+  end
```
farhatahmad, line 110: I wouldn't return self for any of these

app/controllers/api/v1/users_controller.rb
```diff
@@ -23,7 +23,8 @@ def create
 
         if user.save
           session[:user_id] = user.id
-          render_json status: :created
+          token = user.generate_activation_token!
+          render_json data: { token: }, status: :created # TODO: enable activation email sending.
```
farhatahmad, line 27: Why are we returning the token from the user create?

KH-Amir-TN: When we're supposed to send an email with the token I'm sending the token in response.
Shouldn't we send activation token on create?

KH-Amir-TN: Until we setup the email how could we test the reset password and account activation if we're not returning the token?
The only time GL will have the token is on generating or receiving it.
Do you like if we log it to the console instead?

farhatahmad (dismissed):

app/controllers/api/v1/users_controller.rb
KH-Amir-TN: Yup, I'm aware of it that's why I left the inline todo comments `# TODO: enable activation email sending.`

farhatahmad (approved):

PR #3565, opened 2022-06-22
Sorting recordings 0: Extended recordings index to enable sorting.

KH-Amir-TN:
## Description

Enable users to sort recordings by name, length and visibility

This PR completes **0.** 
--
~~0. Extend the recordings API to sort the recordings list as indicated by the client app.~~
1. Update the Recordings table UI to show proper icons and handle click events.

At this point:
> Clicking any of the columns sort by icon should:
	- Show proper feedback i.e. activate the icon.
2. Connect the client app to the API and further extend the UI and mutations:

At this point:
> Clicking any of the columns sort by icon should:
	- Show proper feedback i.e. activate the icon.
	- Invalidate the recordings query cache.
	- Request a new list sorted by the selected column.

> Clicking attempts and outcomes:
   - First click order the list by the selected column ASC.
   - Second click order the list by the selected column DESC.
   - Third click resets the list (No sorting).

>Note: Users can sort by multiple columns at once.

### User story [Sorting recordings]:
---
0. User login.
1. User open All recordings or a room recordings page.
2. User will have a list of the recordings.
3. User can clicks on the columns (date, recording name, room name, length, number of users, visibility, formats.) to sort the list by.
4. User will observe adequate feedbacks and the expected result.

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/controllers/api/v1/recordings_controller.rb
+++ b/app/controllers/api/v1/recordings_controller.rb
@@ -5,14 +5,16 @@ module Api
     class RecordingsController < ApiController
       skip_before_action :verify_authenticity_token # TODO: - Revisit this.
       before_action :find_recording, only: :update
+      before_action :config_sorting, only: :index, if: -> { params.key?(:sort) }
 
       # GET /api/v1/recordings.json
       # Returns: { data: Array[serializable objects(recordings)] , errors: Array[String] }
       # Does: Returns all the Recordings from all the current user's rooms
 
       def index
-        recordings = current_user.recordings&.search(params[:search])
+        recordings = current_user.recordings&.order(@sort_config)&.search(params[:search])
 
+        # TODO: Optimise this.
         recordings.map! do |recording|
           {
             id: recording.id,
@@ -76,6 +78,18 @@ module Api
 
       def recording_params
         params.require(:recording).permit(:name)
+      end
+
+      def config_sorting
+        allowed_columns = %w[name length visibility]
+        allowed_directions = %w[ASC DESC]
+
+        sort_column = params[:sort][:column]
+        sort_direction = params[:sort][:direction]
+
+        return render_json status: :bad_request unless allowed_columns.include?(sort_column) && allowed_directions.include?(sort_direction)
+
+        @sort_config = { sort_column => sort_direction }
       end
 
       def find_recording
--- a/spec/controllers/recordings_controller_spec.rb
+++ b/spec/controllers/recordings_controller_spec.rb
@@ -15,7 +15,6 @@ RSpec.describe Api::V1::RecordingsController, type: :controller do
       recordings = create_list(:recording, 6)
       create_list(:room, 5, user:, recordings:)
       get :index
-
       expect(response).to have_http_status(:ok)
       response_recording_ids = JSON.parse(response.body)['data'].map { |recording| recording['id'] }
       expect(response_recording_ids).to eq(recordings.pluck(:id))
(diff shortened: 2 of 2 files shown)


KH-Amir-TN:
# REBASED ON 117d87671825d8d4dea484c04b239e1314088fb2

spec/controllers/recordings_controller_spec.rb
```diff
@@ -51,6 +51,20 @@
...
+      expect_any_instance_of(ActiveRecord::Relation).to receive(:order).with({ name: 'ASC', length: 'DESC' })
+      get :index, params: { sort: { name: 'ASC', length: 'DESC' } }
+    end
+
+    it 'filters sorting options' do
+      create(:recording)
+      allow_any_instance_of(ActiveRecord::Relation).to receive(:order).and_return(Recording.take)
+      expect_any_instance_of(ActiveRecord::Relation).to receive(:order).with({ visibility: 'ASC' })
+      get :index, params: { sort: { visibility: 'SQL_INJECTION', invalid_attr: 'invalid' } }
+    end
```
farhatahmad, line 67: These tests don't actually test anything - we shouldn't be stubbing the `.order`, What we should do instead is create some recordings with manual values, pass in the sort params, and then manually confirm that they were sorted as we expected

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -78,6 +79,15 @@ def recording_params
...
       end
 
+      def recordings_sort_params
+        allowed_values = %w[ASC DESC]
+        columns = params.permit(sort: %i[name length visibility])[:sort].to_h
+        columns.transform_values do |v|
+          res = v if allowed_values.include?(v)
+          res || 'ASC'
+        end
+      end
```
farhatahmad, line 89: This is too complex and hard to read. I think it may be better if the API supports `sort: { column: 'name', direction: 'ASC' }` instead

spec/controllers/recordings_controller_spec.rb
KH-Amir-TN: Does not that means that we're testing the `order` method functionality?
We're only providing it with the ordering configuration.
It felt similar to `bbb_api`

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -78,6 +80,21 @@ def recording_params
         params.require(:recording).permit(:name)
       end
 
+      def recordings_sort_params
+        params.permit(sort: %i[column direction])
+      end
```
farhatahmad, line 85: There isn't a need for this

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -78,6 +80,21 @@ def recording_params
...
+
+      def config_sorting
+        allowed_columns = %w[name length visibility]
+
+        sort_column = recordings_sort_params[:sort][:column]
+        sort_direction = recordings_sort_params[:sort][:direction]
+
+        return render_json status: :bad_request unless allowed_columns.include?(sort_column) && %w[ASC DESC].include?(sort_direction)
```
farhatahmad, line 93: Just fot consistency, I would add `allowed_directions = %w[ASC DESC]` on line 89

spec/controllers/recordings_controller_spec.rb
```diff
@@ -51,6 +50,36 @@
...
       expect(response_recording_ids).to match_array(recordings.pluck(:id))
     end
+
+    context 'ordering' do
+      before do
+        names = %w[B A C]
+        recordings = build_list(:recording, 3).each do |recording|
+          recording.name = names.shift
+          recording.save!
+        end
```
farhatahmad, line 60: This is complex for no reason. Just build the 3 recordings manually

spec/controllers/recordings_controller_spec.rb
```diff
@@ -51,6 +50,36 @@
...
+      end
+
+      it 'orders the recordings list by column and direction ASC' do
+        get :index, params: { sort: { column: 'name', direction: 'ASC' } }
+        expect(response).to have_http_status(:ok)
+        # Order is important match_array isn't adequate for this test.
+        expect(JSON.parse(response.body)['data'].pluck('name')).to eq(%w[A B C])
+      end
+
+      it 'returns :bad_request for bad params if sort config is porivded' do
```
farhatahmad, line 78: `porivded` ?

KH-Amir-TN: Hh, sorry provided.

farhatahmad (approved):

PR #3569, opened 2022-06-22
Admin Panel setup

hadicheaito1:
## Description

- Skeleton code for front end and back end of Admin Panel

image

## NOTE
Could not get regular font awesome icons for the ones needed, instead had to use solid ones

### DIFF
--- a/app/assets/stylesheets/admin_panel.scss
+++ b/app/assets/stylesheets/admin_panel.scss
@@ -1,0 +1,31 @@
+#admin-panel {
+
+  #profile-sidebar {
+    height: 100%;
+    border-right: 2px solid whitesmoke;
+    padding: 1rem;
+  }
+
+  .nav-pills .nav-link.active {
+    color: $primary;
+    background-color: $primary-light;
+    font-weight: 600;
+  }
+
+  .tab-content {
+    display: flex;
+  }
+
+  .tab-content > .tab-pane {
+    display: block;
+    visibility: hidden;
+    margin-right: -100%;
+    width: 100%;
+  }
+
+  .tab-content > .active {
+    visibility: visible;
+  }
+
+}
+  
--- a/app/assets/stylesheets/admin_panel.scss
+++ b/app/assets/stylesheets/admin_panel.scss
@@ -1,6 +1,6 @@
 #admin-panel {
 
-  #profile-sidebar {
+  #admin-sidebar {
     height: 100%;
     border-right: 2px solid whitesmoke;
     padding: 1rem;
--- a/app/controllers/api/v1/admin/users_controller.rb
+++ b/app/controllers/api/v1/admin/users_controller.rb
@@ -3,7 +3,7 @@
 module Api
   module V1
     module Users
-      class UserController < ApiController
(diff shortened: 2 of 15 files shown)


app/controllers/api/v1/admin/users_controller.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
```
KH-Amir-TN, line 1: What would this controller be in charge of?

db/schema.rb
```diff
@@ -134,7 +134,6 @@
     t.datetime "activation_sent_at", precision: nil
     t.index ["activation_digest"], name: "index_users_on_activation_digest", unique: true
     t.index ["email", "provider"], name: "index_users_on_email_and_provider", unique: true
-    t.index ["reset_digest"], name: "index_users_on_reset_digest", unique: true
```
KH-Amir-TN, line 137: This shouldn't be removed

db/schema.rb
```diff
@@ -98,9 +98,9 @@
     t.datetime "last_session"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.integer "recordings_processing", default: 0
     t.string "viewer_access_code"
     t.string "moderator_access_code"
-    t.integer "recordings_processing", default: 0
```
KH-Amir-TN, line 103: It's better if we leave the schema for migrations to shape or we'll be having a lot of changes whenever a migration change occurs.

KH-Amir-TN (changes requested):

app/controllers/api/v1/admin/users_controller.rb
hadicheaito1: I'm not sure, I spoke to @farhatahmad and he said to make one. Could you explain Ahmad?

app/assets/stylesheets/admin_panel.scss
```diff
@@ -0,0 +1,31 @@
+#admin-panel {
+
+  #admin-sidebar {
+    height: 100%;
+    border-right: 2px solid whitesmoke;
+    padding: 1rem;
```
farhatahmad, line 6: this can be done in bootstrap

app/controllers/api/v1/admin/users_controller.rb
KH-Amir-TN: What about the `users_controller#index`?
This way we're creating a differentiating between all roles and an Admin.
Admins now have their own API when actually Admin is a Role just like `User` and `Guest` but with absolute permissions.
If we're still designing GL3 to be permission based and since it's now represented as an Restful API I think it makes more sense to have our controllers subject based like Users, Recordings, ... and make the proper restrictions so that accessing certain endpoints or accessing certain data is controlled.

KH-Amir-TN: If a certain area of the API encapsulates a lot of complexity then for simplicity sake we can extract it into its own controller, Is that what you intended @farhatahmad ?

farhatahmad (approved):

PR #3570, opened 2022-06-22
Add Protected to Recording visibility

scouillard:


app/controllers/api/v1/recordings_controller.rb
```diff
@@ -57,19 +57,28 @@ def resync
...
+          published = false
+          protected = false
+        when 'Protected'
+          published = true
+          protected = true
         end
+
         Recording.find_by(record_id:).update(visibility:)
-        BigBlueButtonApi.new.publish_recordings(record_ids: record_id, publish:)
+        BigBlueButtonApi.new.publish_recordings(record_ids: record_id, publish: published)
```
KH-Amir-TN, line 78: We can further optimise this since some expensive updates aren't always needed.
Example: From Published to Unpublished only the published attribute changes so no need to further change the protected.
Same for Published and Protected.

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -57,19 +57,28 @@ def resync
...
         record_id = params[:record_id]
-        case publish
-        when 'true'
-          visibility = 'Published'
-        when 'false'
-          visibility = 'Unpublished'
+        visibility = params[:visibility]
+
+        case visibility
+        when 'Published'
```
KH-Amir-TN, line 66: I think this can be given to the `#publish_recordings` which will have the complexity of this business logic to determine the right configuration and maybe call `#update_recordings` or other services when needed.

app/javascript/hooks/mutations/recordings/useUpdateRecordingVisibility.jsx
```diff
@@ -0,0 +1,15 @@
+import { useMutation, useQueryClient } from 'react-query';
+import axios from 'axios';
```
KH-Amir-TN, line 2: Using the custom configured `axiosInstance` can be required in the upcoming sprints.

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -57,19 +57,28 @@ def resync
...
-        publish = params[:publish]
+      # POST /api/v1/recordings/update_visibility.json
+      def update_visibility
         record_id = params[:record_id]
-        case publish
-        when 'true'
-          visibility = 'Published'
-        when 'false'
-          visibility = 'Unpublished'
+        visibility = params[:visibility]
```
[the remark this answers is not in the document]
scouillard: Agreed.

app/controllers/api/v1/recordings_controller.rb
scouillard: I'll put some thoughts into trying to optimize this.

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -57,19 +57,28 @@ def resync
...
+          protected = false
+        when 'Protected'
+          published = true
+          protected = true
         end
+
         Recording.find_by(record_id:).update(visibility:)
-        BigBlueButtonApi.new.publish_recordings(record_ids: record_id, publish:)
+        BigBlueButtonApi.new.publish_recordings(record_ids: record_id, publish: published)
+        BigBlueButtonApi.new.update_recordings(record_ids: record_id, meta_hash: { meta_protected: protected })
```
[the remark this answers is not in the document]
scouillard: Is there a method we can use to edit the params and the metaparams of a recording?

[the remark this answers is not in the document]
KH-Amir-TN: All of the `bbb_ruby_api` gem API methods are empowered by a general purpose method the `send_api_request`.
I've used it in my internship to create new API methods for custom requests, we can use it to create our own custom API requests.
Don't know if @farhatahmad would like this because will be tightly coupled to the implementation of a dependency.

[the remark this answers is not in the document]
KH-Amir-TN: But, that's if the BBB API itself allows us to publish and set the recording `meta_data` at once because I think they are two different APIs `publish` vs `update`.

app/controllers/api/v1/recordings_controller.rb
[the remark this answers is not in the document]
scouillard: This will have to be done when refactoring as I would have to make big changes to the RecordingController to allow the scoping.

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -57,19 +59,27 @@ def resync
...
-        when 'false'
-          visibility = 'Unpublished'
-        end
+      # POST /api/v1/recordings/update_visibility.json
+      def update_visibility
+        record_id = params[:id]
+        visibility = params[:visibility]
+
+        # Updates the recording locally
         Recording.find_by(record_id:).update(visibility:)
```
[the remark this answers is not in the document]
KH-Amir-TN: Shouldn't we update the record on the DB after we update the recording data on the BBB?
This is safer for cases where the requests to BBB fails i.e. a network issue we won't update the recording state unless it's updated on the BBB server.
Plus on the `send_XXX_request` I think we count on checking the old and new visibility values that will always be the same since we're calling them after updating the recrod.
That means that both methods will always be turthy and therefore the optmisation isn't happening.

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -57,19 +59,27 @@ def resync
...
-        record_id = params[:record_id]
-        case publish
-        when 'true'
-          visibility = 'Published'
-        when 'false'
-          visibility = 'Unpublished'
-        end
+      # POST /api/v1/recordings/update_visibility.json
+      def update_visibility
+        record_id = params[:id]
```
KH-Amir-TN, line 64: I think we don't need this because we're using the `find_recording` callback.

app/controllers/api/v1/recordings_controller.rb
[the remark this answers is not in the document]
KH-Amir-TN: It may be a good idea even for consistency if we further catch invalid/missing params.
`return render_json status: :bad_request unless params[:visibility]`
until we do the refactoring.

app/services/big_blue_button_api.rb
```diff
@@ -51,8 +51,8 @@ def publish_recordings(record_ids:, publish:)
     bbb_server.publish_recordings(record_ids, publish)
   end
 
-  def update_recordings(record_id:, meta_hash:)
-    bbb_server.update_recordings(record_id, {}, meta_hash)
+  def update_recordings(record_ids:, meta_hash:)
```
farhatahmad, line 54: What's the point of this switch?

app/services/recording_creator.rb
```diff
@@ -23,6 +23,17 @@ def call
 
   private
 
+  # Returns the visibility of the recording (published, unpublished or protected)
+  def get_recording_visibility(recording:)
+    return 'Protected' if recording[:metadata][:protected] == 'true'
```
farhatahmad, line 28: Uhhhh I don't think `protected` is under `metadata` I think its directly under recording

farhatahmad:
Was this tested using a BigBlueButton Server that supports protected recordings?

app/services/recording_creator.rb
scouillard: It seems to be metadata in GLv2.

scouillard: Who could confirm this? It's not in the docs.

app/services/big_blue_button_api.rb
scouillard: To conform to the BBB-api-gem - it specifies record_ids and not record_id.

app/models/recording.rb
```diff
@@ -15,4 +15,34 @@ def self.search(input)
...
+
+    true
+  end
+
+  # Returns true if the recording visibility is changing from Published or Unpublished to Protected and vice versa
+  def send_update_request(new_visibility)
+    return false if (visibility == 'Published' && new_visibility == 'Unpublished') ||
+                    (visibility == 'Unpublished' && new_visibility == 'Published')
+
+    true
```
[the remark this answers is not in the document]
scouillard: Would you put the logic in the Controller?
Or you would do things a completely different way?

app/services/recording_creator.rb
scouillard: BBB does not support protected yet.
It is in talk for BBB 2.6 or later (later is more likely).

scouillard:
BBB does not support protected yet.
It is in talk for BBB 2.6 or later (later is more likely).

app/controllers/api/v1/recordings_controller.rb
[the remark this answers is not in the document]
KH-Amir-TN: What validations?
The `ActiveRecord::Validations`?

app/services/big_blue_button_api.rb
farhatahmad: Our service works separately from the gem. At the moment, there isn't any need to update multiple recordings at once, so I would leave the service as is

app/services/recording_creator.rb
```diff
@@ -23,6 +23,17 @@ def call
...
 
+  # Returns the visibility of the recording (published, unpublished or protected)
+  def get_recording_visibility(recording:)
+    return 'Protected' if recording[:protected] == 'true'
+
+    if recording[:published]
+      'Published'
+    else
+      'Unpublished'
+    end
```
farhatahmad, line 34: I would keep using the same pattern

```
return 'Protected' if recording[:protected] == 'true'
return 'Published' if recording[:published]
'Unpublished'

```

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -46,17 +46,31 @@ def resync
...
+        bbb_api.publish_recordings(record_ids: @recording.record_id, publish: true) if old_visibility == 'Unpublished'
+
+        # send unpublish
+        bbb_api.publish_recordings(record_ids: @recording.record_id, publish: false) if new_visibility == 'Unpublished'
+
+        # send protect
+        bbb_api.update_recordings(record_ids: @recording.record_id, meta_hash: { protect: true }) if new_visibility == 'Protected'
+
+        # send unprotect
+        bbb_api.update_recordings(record_ids: @recording.record_id, meta_hash: { protect: false }) if old_visibility == 'Protected'
```
farhatahmad, line 71: I would remove the comments here

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -46,17 +46,31 @@ def resync
...
-        end
-        Recording.find_by(record_id:).update(visibility:)
-        BigBlueButtonApi.new.publish_recordings(record_ids: record_id, publish:)
+      # POST /api/v1/recordings/update_visibility.json
+      # unpublished (publish: false, protected: false) -> protected (publish: true, protected: true)
+      # unpublished (publish: false, protected: false) -> published (publish: true, protected: false)
+      # protected (publish: true, protected: true) -> unpublished (publish: false, protected: false)
+      # protected (publish: true, protected: true) -> published (publish: true, protected: false)
+      # published (publish: true, protected: false) -> protected (publish: true, protected: true)
+      # published (publish: true, protected: false) -> unpublished (publish: false, protected: false)
```
farhatahmad, line 55:
```
      # unpublished -> protected (publish: true, protected: true)
      # unpublished -> published (publish: true, protected: false)
      # protected -> unpublished (publish: false, protected: false)
      # protected  -> published (publish: true, protected: false)
      # published -> protected (publish: true, protected: true)
      # published  -> unpublished (publish: false, protected: false)
```

spec/controllers/recordings_controller_spec.rb
```diff
@@ -134,13 +135,107 @@
     end
   end
 
-  describe '#publish_recording' do
-    it 'Updates Recording with new visibility value' do
+  describe '#update_visibility' do
+    # published (publish: true, protected: false) -> unpublished (publish: false, protected: false)
+    it 'calls the publish request with the right values when changing recording visibility from Published to Unpublished' do
```
farhatahmad, line 140: I would remove the comments and call the test something simpler `it 'moves a recording from Published to Unpublished'`

scouillard: There is 6 tests that checks if the recording is changed on the server (if the requests are being made) and 6 tests that checks if the recording visibility is changed locally. Should I bundle them together? I thought it would be easier by splitting the db from the requests.

spec/controllers/recordings_controller_spec.rb
```diff
@@ -134,13 +135,98 @@
...
+    end
+
+    # protected -> unpublished
+    it 'changes the recording visibility when changing from Protected to Unpublished locally' do
+      recording = create(:recording, visibility: 'Protected')
+      allow_any_instance_of(BigBlueButtonApi).to receive(:publish_recordings)
+      allow_any_instance_of(BigBlueButtonApi).to receive(:update_recordings)
+      post :update_visibility, params: { visibility: 'Unpublished', id: recording.record_id }
+      expect(recording.reload.visibility).to eq('Unpublished')
     end
```
farhatahmad, line 230: This is a little overkill since these tests all do the same. I would do just 1 test to make sure the local value is changed

app/javascript/hooks/mutations/recordings/useUpdateRecordingVisibility.jsx
```diff
@@ -0,0 +1,19 @@
...
+  const mutation = useMutation(updateRecordingVisibility, {
+    onSuccess: () => {
+      toast.success('Recording visibility updated');
+      queryClient.invalidateQueries('getRecordings');
+    },
+  });
+
+  const handleUpdateRecordingVisibility = (visibilityData) => mutation.mutateAsync(visibilityData)
+    .catch(/* Prevents the promise exception from bubbling */() => {});
+  return { handleUpdateRecordingVisibility, ...mutation };
```
farhatahmad, line 18: This needs to be updated to match the new format

app/services/big_blue_button_api.rb
farhatahmad: Yeah we should - but we can do that in a separate PR

PR #3571, opened 2022-06-22
Initial toast work

farhatahmad:


farhatahmad:
I chose to use `react-hot-toast` instead of the bootstrap ones since they're much nicer and easier to use

scouillard (approved):

PR #3572, opened 2022-06-23
Create Role 0: Schema

KH-Amir-TN:
## Description

Enable users with the right set of privileges to create roles.

This PR completes **0.** 
--
1. Add `RolesController#create` to create a new role while providing its name.
> When permissions is made Roles just like `RoomMeetingOptions` should take the default
configuration on `:create` and can be updated later on.
2. Extend the UI and create a CreateRole modal that prompts  the user for a role name when clicking a button in the Roles page of the organization settings.
3. Extend/Add required mutations for integration with the `RolesController#create` API and
show proper feedbacks.
 
---
1. A User with the right set of priviledges authenticates.
2. User goes to the Organasation settings.
[OR]
[OR]

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -4,4 +4,12 @@ class Role < ApplicationRecord
   has_many :users, dependent: :restrict_with_exception
 
   validates :name, presence: true
+
+  before_validation :set_random_color, on: :create
+
+  private
+
+  def set_random_color
+    self.color = "##{SecureRandom.hex(3)}"
+  end
 end
--- a/db/migrate/20220531180853_create_roles.rb
+++ b/db/migrate/20220531180853_create_roles.rb
@@ -4,6 +4,7 @@ class CreateRoles < ActiveRecord::Migration[7.0]
   def change
     create_table :roles do |t|
       t.string :name, null: false
+      t.string :color, default: '', null: false
       t.timestamps
     end
   end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -76,6 +76,7 @@ ActiveRecord::Schema[7.0].define(version: 2022_06_16_175700) do
 
   create_table "roles", force: :cascade do |t|
     t.string "name", null: false
+    t.string "color", default: "", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
   end
--- a/spec/models/role_spec.rb
+++ b/spec/models/role_spec.rb
@@ -7,5 +7,22 @@ RSpec.describe Role, type: :model do
     subject { create(:role) }
 
     it { is_expected.to have_many(:users).dependent(:restrict_with_exception) }
+
+    context 'color attribute' do
+      context 'on :create' do
+        it 'defaults the color when not provided on :create' do
+          allow(SecureRandom).to receive(:hex).and_return('ffffff')
+          role = create(:role)
+          expect(role.reload.color).to eq('#ffffff')
+        end
+
+        it 'overwrites any color value when provided on :create' do
+          allow(SecureRandom).to receive(:hex).and_return('ffffff')
+          role = build(:role, color: '#HEXTET')
+          expect(role).to be_valid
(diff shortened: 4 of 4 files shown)


scouillard (approved):

farhatahmad:
I'm thinking of partially changing this requirement. Role color has little impact and allowing the admin to change the colors doesn't really seem that useful. 

I think we should set some default colors for the 3 base roles and then just auto generate a new color for any new role that's created

app/validators/hex_color_validator.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class HexColorValidator < ActiveModel::EachValidator
+  HEX_COLOR_PATTERN = /\A#([A-F0-9]{6}|[A-F0-9]{3})\Z/i
+
+  def validate_each(record, attribute, value)
+    record.errors.add(attribute, message: 'Invalid CSS hex color code.') unless HEX_COLOR_PATTERN.match?(value) # TODO: Add I18n.
+  end
+end
```
farhatahmad, line 9: I think we can drop this

app/models/role.rb
```diff
@@ -4,4 +4,13 @@ class Role < ApplicationRecord
...
 
   validates :name, presence: true
+  validates :color, hex_color: true
+
+  before_validation :set_default_color, on: :create
+
+  private
+
+  def set_default_color
+    self.color = Rails.configuration.styles[:primary] unless color?
```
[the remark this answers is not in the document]
KH-Amir-TN: Likes the idea,

app/validators/hex_color_validator.rb
KH-Amir-TN: Yup, if we're generating the colors

app/models/role.rb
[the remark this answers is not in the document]
KH-Amir-TN: For the update I think we should give the admins the right to generate a new random color from the client app just like with GitHub labels?

app/models/role.rb
```diff
@@ -4,4 +4,12 @@ class Role < ApplicationRecord
   has_many :users, dependent: :restrict_with_exception
 
   validates :name, presence: true
+
+  before_validation :set_random_color, on: :create
+
+  private
+
+  def set_random_color
+    self.color = "##{SecureRandom.hex(3)}"
```
farhatahmad, line 13: To avoid collisions, isn't 6 char hex better?

db/migrate/20220531180853_create_roles.rb
```diff
@@ -4,6 +4,7 @@ class CreateRoles < ActiveRecord::Migration[7.0]
   def change
     create_table :roles do |t|
       t.string :name, null: false
+      t.string :color, default: '', null: false
```
farhatahmad, line 7: default: '' allows empty records in the db which isn't what we want

app/models/role.rb
KH-Amir-TN: What do you mean?
`SecureRandom.hex(3)` generates 6 hex chars random.
It's a bit deceiving but the parameter isn't the length but the byteLength.

db/migrate/20220531180853_create_roles.rb
KH-Amir-TN: The problem was with rubocop, I haven't thought about it.
Should I disable the convention for that line?

KH-Amir-TN: Have you faced something like this when setting `null: false`?

app/models/role.rb
farhatahmad: :facepalm:

db/migrate/20220531180853_create_roles.rb
farhatahmad: Hmm that should be fine then

farhatahmad (approved):

PR #3574, opened 2022-06-23
Roles controller skeleton

KH-Amir-TN:
## Description

This PR prepares the `RolesControllerSkeleton` to base Role creation, edit and delete PRs for a fast-forward branching.

PR #3575, opened 2022-06-23
Roles edit 1: Added Roles update API.

KH-Amir-TN:
## Description

Enable users with the right set of privileges to update roles.

This PR completes **1.** 
--
~~1. Add  `RolesController#update` to edit a new role while validating its color to be a HEX 
color.~~
> When permissions are made roles updates should change their permissions.
2. Extend the UI on Roles page in the organization settings to select a Role and show
an edit page like with room settings with a form to update name, color and turn on/off each
 role permission.
3. Extend/Add required mutations and feedbacks to integrate with the `UpdateRoleForm` 
with `RolesController#update` API.

 
---
2. User goes to the Organization settings.
[OR]
[OR]

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/controllers/api/v1/admin/roles_controller.rb
+++ b/app/controllers/api/v1/admin/roles_controller.rb
@@ -4,6 +4,8 @@ module Api
   module V1
     module Admin
       class RolesController < ApiController
+        before_action :find_role, only: :update
+
         # POST /api/v1/admin/roles.json
         # Expects: {}
         # Returns: { data: Array[serializable objects] , errors: Array[String] }
@@ -29,10 +31,25 @@ module Api
           render_json status: :created
         end
 
+        # POST /api/v1/:id/roles.json
+        # Expects: { role: {:name, :color} }
+        # Returns: { data: Array[serializable objects] , errors: Array[String] }
+        # Does: Updates a role.
+
+        def update
+          return render_error errors: @role.errors.to_a, status: :bad_request unless @role.update role_params
+
+          render_json status: :ok
+        end
+
         private
 
         def role_params
           params.require(:role).permit(:name)
+        end
+
+        def find_role
+          @role = Role.find params[:id]
         end
       end
     end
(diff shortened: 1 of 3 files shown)


KH-Amir-TN:
# DEPENDS ON #3574

app/controllers/api/v1/admin/roles_controller.rb
```diff
@@ -17,6 +18,40 @@ def index
...
+
+        # POST /api/v1/roles.json
+        # Expects: { role: {:name, :color} }
+        # Returns: { data: Array[serializable objects] , errors: Array[String] }
+        # Does: Creates a role.
+
+        def create
+          role = Role.new role_params
+
+          return render_json errors: role.errors.to_a, status: :bad_request unless role.save
```
hadicheaito1, line 30: Although this works. I prefer the simpler (easier to read) way that we have in other controllers such as in rooms_controller

app/controllers/api/v1/admin/roles_controller.rb
```diff
@@ -17,6 +18,40 @@ def index
...
+          render_json status: :created
+        end
+
+        # POST /api/v1/:id/roles.json
+        # Expects: { role: {:name, :color} }
+        # Returns: { data: Array[serializable objects] , errors: Array[String] }
+        # Does: Updates a role.
+
+        def update
+          return render_json errors: @role.errors.to_a, status: :bad_request unless @role.update role_params
```
hadicheaito1, line 41: Same as above comment

app/controllers/api/v1/admin/roles_controller.rb
KH-Amir-TN: Which is?

KH-Amir-TN: Is it the  `render_error`?
If that's the case then this PR was made way before including it so I forgot to include it, will do Inshallah.

hadicheaito1: It isn't a big deal and you don't need to change it. I am just trying to keep simple CRUD procedures following the same format is possible

PR #3577, opened 2022-06-24
Admin Manage Users - Active tab

hadicheaito1:
## Description

- Admin Panel Manage Users tab now shows all users (needs to be changed to only show active users)
## Testing Steps

- Pull changes
- Go to Admin Panel
- Should show a list of all users for now

image

## TODO
Create rspec test in next PR


app/controllers/api/v1/admin/users_controller.rb
```diff
@@ -4,6 +4,26 @@ module Api
...
+          # TODO: Change to get active users only
+          users = User.all.to_a
+
+          users.map! do |user|
+            {
+              id: user.id,
+              name: user.name,
+              email: user.email,
+              provider: user.provider,
+              role: Role.find(user.role_id).name,
```
scouillard, line 19: You can use user.role.name

scouillard: Could we call this active_users instead of active_index? And then pending_users, banned_users, deleted_users

app/javascript/components/admin/AdminUserRow.jsx
```diff
@@ -0,0 +1,38 @@
...
+import Avatar from '../users/Avatar';
+
+export default function AdminUserRow({ user }) {
+  return (
+    <tr key={user.id} className="align-middle text-muted">
+      <td className="text-dark">
+        <Stack direction="horizontal">
+          <div>
+            <Avatar className="avatar" avatar={user.avatar} radius={40} />
+          </div>
```
scouillard, line 13: You can add <div className="me-2"> to avatar parent here instead of having an extra CSS class.

app/assets/stylesheets/application.bootstrap.scss
```diff
@@ -91,3 +91,8 @@ input.search-bar {
 .hi-xl {
   width: 72px;
 }
+
+// Eventually remove once spacing in className works for avatar in AdminUserRow
+.avatar {
+  margin-right: 5px;
+}
```
scouillard, line 98: You can add `<div className="me-2">` to avatar parent div instead of having an extra CSS class.

app/javascript/hooks/queries/admins/useActiveUsers.jsx
```diff
@@ -0,0 +1,11 @@
+import { useQuery } from 'react-query';
+import axios from 'axios';
```
KH-Amir-TN, line 2: The use of the custom `axiosInstance` will be necessary in the upcoming sprints.

app/javascript/components/admin/ManageUsersTable.jsx
```diff
@@ -0,0 +1,44 @@
...
+  );
+}
+
+ManageUsersTable.defaultProps = {
+  users: [],
+};
+
+ManageUsersTable.propTypes = {
+  users: PropTypes.arrayOf(PropTypes.shape({
+    length: PropTypes.number.isRequired,
```
KH-Amir-TN, line 41: I think this should be:
```
ManageUsersTable.propTypes = {
  users: PropTypes.arrayOf(PropTypes.shape({
    id: PropTypes.number.isRequired,
  })),
};
```

config/routes.rb
```diff
@@ -58,6 +58,18 @@
...
+      namespace :admin do
+        resources :admins, only: %i[]  do
+          collection do
+            get '/active_users', to: 'users#active_index'
+          end
+        end
+      end
+    end
+  end
+
```
farhatahmad, line 72: You shouldn't re-create the whole `api - v1` namespace. Just move it under the existing one

app/javascript/components/admin/ActiveUsers.jsx
```diff
@@ -0,0 +1,11 @@
...
+import useActiveUsers from '../../hooks/queries/admins/useActiveUsers';
+import ManageUsersTable from './ManageUsersTable';
+
+export default function ActiveUsers() {
+  const { data: users } = useActiveUsers();
+
+  return (
+    <ManageUsersTable users={users} />
+  );
+}
```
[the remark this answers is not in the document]
hadicheaito1: I made it this way looking into the future.

The reason I did this is so that when we implement the other tabs such as 'Pending', we can re-use the <ManageUsersTable> and just pass in the pending 'Users' into it. It cleans up the code and gets rid of repetition

config/routes.rb
hadicheaito1: > You'll prob want to put `/active_users` under the users controller

I'm confused, it's already in the users admin controller ?

app/javascript/components/admin/ActiveUsers.jsx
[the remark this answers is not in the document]
scouillard: Right. Nicely done!

farhatahmad (approved):

PR #3578, opened 2022-06-24
Admin users list rspec

hadicheaito1:
## Description

Just the simple test that checks if the id's of the users created matches the id's of the users from the response.

TODO: Eventually this test will change since it is not checking for active users

### DIFF
--- a/spec/controllers/admins/users_controller_spec.rb
+++ b/spec/controllers/admins/users_controller_spec.rb
@@ -11,7 +11,7 @@ RSpec.describe Api::V1::Admin::UsersController, type: :controller do
     it 'ids of rooms in response are matching room ids that belong to current_user' do
       # TODO: Change this test to create active users and not just any users
       users = create_list(:user, 5)
-      get :active_index
+      get :active_users
       expect(response).to have_http_status(:ok)
       response_user_ids = JSON.parse(response.body)['data'].map { |user| user['id'] }
       expect(response_user_ids).to eq(users.pluck(:id))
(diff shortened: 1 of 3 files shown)


hadicheaito1:
Waiting for #3577

spec/controllers/admins/users_controller_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  describe '#active_users' do
+    it 'ids of rooms in response are matching room ids that belong to current_user' do
+      # TODO: Change this test to create active users and not just any users
+      users = create_list(:user, 5)
+      get :active_index
+      expect(response).to have_http_status(:ok)
+      response_user_ids = JSON.parse(response.body)['data'].map { |user| user['id'] }
+      expect(response_user_ids).to eq(users.pluck(:id))
```
KH-Amir-TN, line 17: This is a flaky test, the order of ids may differ braking the tests run.
I would do this instead:
```
expect(JSON.parse(response.body)['data'].pluck('id')).to match(users.pluck(:id))
```

KH-Amir-TN (changes requested):

spec/controllers/admins/users_controller_spec.rb
hadicheaito1: Order doesn't seem to be considered when using `eq`, we have many tests that follow this same format and never fail

farhatahmad: Those tests were fixed by Samuel since they were occasionally failing. We should make the change to `match_array`

farhatahmad (approved):

PR #3583, opened 2022-06-27
Roles View 1: Added `admin::roles_controller#index` API.

KH-Amir-TN:
## Description

Enable users with the right set of privileges to view roles.

This PR completes **0.** 
--
### Increments [View Roles]:
~~0. Add  `RolesController#index` to view all roles.~~
1. Extend the UI and mutations to display the list of roles in the Roles table.

---
 
### User story [View Roles]:
2. User goes to the Admin Panel.
4. User view the list of all roles paginated.

**_[If user types a search text]:_**
1. The roles list will get filtered out and paginated according to the search input.

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/controllers/api/v1/admin/roles_controller.rb
+++ b/app/controllers/api/v1/admin/roles_controller.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module Admin
+      class RolesController < ApiController
+        skip_before_action :verify_authenticity_token
+
+        # POST /api/v1/admin/roles.json
+        # Expects: {}
+        # Returns: { data: Array[serializable objects] , errors: Array[String] }
+        # Does: Fetches and returns the list of roles.
+
+        def index
+          sort_config = config_sorting(allowed_columns: %w[name])
+
+          roles = Role.select(:id, :name, :color)&.order(sort_config)&.search(params[:search])
+          render_json data: roles
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/admin/roles_controller.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module Admin
+      class RolesController < ApiController
+        skip_before_action :verify_authenticity_token
+        before_action -> { config_sorting(allowed_columns: %w[name]) }, only: :index, if: -> { params.key?(:sort) }
```
farhatahmad, line 8: I'm not a huge fan of this style of programming - I find it extremely difficult to follow. I'd rather we manually call it in the action and return values rather than setting instance variables through before_actions

KH-Amir-TN: The problem is with the difference between a regular method and a callback.
Regular methods if called the `render|redirect_to` will not stop the request response cycle which will lead us to do some extra logic in order to ensure the expected completion of the response like passing a reference or returning certain outputs.
Do you approve this style instead:
```
before_action do
  # Code
end
```

farhatahmad: I would just call the method directly and ignore anything that isn't allowed. There's no need for that method to raise an error

KH-Amir-TN:
# DEPENDS ON #3592

KH-Amir-TN:
# REBASED ON 634ce9961d7e37cc1ea5f88e276ca04cbe53dd8c

PR #3586, opened 2022-06-28
Axios: Added search params middleware.

KH-Amir-TN:
## Description

This PR introduces the axios middleware responsible for extracting search params (query params) from the URL, filtering and then merging them with any request-params given to the axios Instance before forwarding the HTTP request.
Now, having a query parameter like `search`, `sort[column]` and `sort[direction]` will be allowed in axios requests automatically with no intervention.
In the upcoming sprints this will further include `page` query param.
This enables GL3 to use one of the native global and shared states which is the URL for sorting, searching and pagination.
This adds the benefits of:
1. Inversion Of Control: A component like `SortBy` can be used throughout the application and enable developers to **inject** sorting capabilities for any other component for any field with no hard dependency and leave the control logic external.
2. Separation Of concern; A single component like `SortBy` will only be concerned to listen to certain events and translate them into the URL state accordingly with no other complexity.
3.  Reusability and maintenance: The two points above concludes the reusability of generic components in the application with little to no extra code addition and therefore having a better maintenance in the long run.
4. Bookmarking and history: Using the URL state for subjects like search, sorting and pagination is natural.
All visited URLs will be pushed to the browser history and can be bookmarked.
Further making the URL state use bidirectional (Altering the URL state alters the UI and vice versa) will enable users to bookmark and save certain results based on the URL.

### DIFF
--- a/app/javascript/helpers/Axios.jsx
+++ b/app/javascript/helpers/Axios.jsx
@@ -1,5 +1,35 @@
 import axios from 'axios';
 
+// Merges custom params with search params.
+function mergeConfigParams(params, configParams = { }) {
+  Object.entries(configParams).forEach(
+    (param) => {
+      const key = param.at(0);
+      const val = param.at(1);
+      if (val) {
+        params.set(key, val);
+      }
+    },
+  );
+
+  return params;
+}
+
+// Filters out unwatned serach params.
+function filterParams(params) {
+  const allowedParams = ['search', 'sort[column]', 'sort[direction]'];
+  const isAllowed = (val) => allowedParams.includes(val);
+
+  const keys = Array.from(params.keys());
+  keys.forEach((key) => {
+    if (!isAllowed(key)) {
+      params.delete(key);
+    }
+  });
+
+  return params;
+}
+
 export const ENDPOINTS = {
   signup: '/users.json',
   signin: '/sessions.json',
@@ -10,6 +40,7 @@ export const ENDPOINTS = {
   updateRecording: (recordId) => `/recordings/${recordId}.json`,
   changePassword: '/users/change_password.json',
   forget_password: '/reset_password.json',
+  recordings: '/recordings.json',
 };
 
 const axiosInstance = axios.create(
@@ -30,5 +61,12 @@ const axiosInstance = axios.create(
     timeout: 30_000, // default is `0` (no timeout)
   },
 );
+
+// Intercepts requests, filters out and forward search params to the API.
+axiosInstance.interceptors.request.use((_config) => {
+  const params = new URLSearchParams(window.location.search);
+  mergeConfigParams(filterParams(params), _config.params);
+  return { ..._config, ...{ params } };
+}, (error) => Promise.reject(error));
 
 export default axiosInstance;


app/javascript/helpers/Axios.jsx
```diff
@@ -1,5 +1,35 @@
...
+      const key = param.at(0);
+      const val = param.at(1);
+      if (val) {
+        params.set(key, val);
+      }
+    },
+  );
+
+  return params;
+}
```
KH-Amir-TN, line 16: `mergeConfigParams(params: URLSearchParams, configParams: Object): URLSearchParams`
This function will be responsible for:
1. Taking in the search `params` extracted from the `window.location.search` and the `configParams` object given by the axios configuration used in a request `config.params` [MORE DETAILS](https://axios-http.com/docs/req_config).
2. Iterating over all axios configured params and add (merge) them to (with) the search `params` _if_ they have a truthy value since the config params allows `undefined|null|''` values to be passed for the JSON parser to remove.
-> The result will be a new list of params containing both of what extracted from the URL and what given to axios when setting the request therefore this function merges config params with search params.

app/javascript/helpers/Axios.jsx
```diff
@@ -1,5 +1,35 @@
...
+
+  const keys = Array.from(params.keys());
+  keys.forEach((key) => {
+    if (!isAllowed(key)) {
+      params.delete(key);
+    }
+  });
+
+  return params;
+}
```
KH-Amir-TN, line 31: `filterParams(params: URLSearchParams): URLSearchParams`
This function will be responsible for:
1. Taking in a list of serach `params`.
2. Remove any param that is not allowed by iterating over all param names and check their legitimately with a predefined list of allowed param names `allowedParams`.
-> The end result will be search params that are allowed only to be passed in the requests.
This prevents registering and  sending random search params i.e. for `?foo=foo&bar=bar&search=GL` will be treated as `?search=GL` according to this list of allowed params.
This function then filter out serach params.

app/javascript/helpers/Axios.jsx
```diff
@@ -31,4 +62,11 @@ const axiosInstance = axios.create(
   },
 );
 
+// Intercepts requests, filters out and forward search params to the API.
+axiosInstance.interceptors.request.use((_config) => {
+  const params = new URLSearchParams(window.location.search);
+  mergeConfigParams(filterParams(params), _config.params);
+  return { ..._config, ...{ params } };
+}, (error) => Promise.reject(error));
+
```
KH-Amir-TN, line 71: This axios request interceptor (middleware) will run automatically for each request and be responsible for:
1. Extracting the current list of search params through the `URLSearchParams` API [MORE DETAILS](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams).
2. Filter out the current list of search params `filterParams(params)`.
3. Merge the filtered params with the config params `mergeConfigParams(filterParams(params), _config.params);`
4. Return a new axios configuration that has the new merged params   `return { ..._config, ...{ params } };` according to the Interceptors API specs [MORE DETAILS](https://axios-http.com/docs/interceptors).
-> This interceptor will then extract, filter search params and merge them with config params and send them on each request to the API.
The API will then can extract those params when needed to do searching, sorting or pagination.

KH-Amir-TN:
# REBASED ON 73ffb7f969d25c5daa2f2a92d7dce109bb7b923e

PR #3589, opened 2022-06-28
Added a basic room serializer

farhatahmad:

### DIFF
--- a/app/controllers/api/v1/api_controller.rb
+++ b/app/controllers/api/v1/api_controller.rb
@@ -33,7 +33,7 @@ module Api
       end
 
       def render_data(data: {}, status: :ok, include: nil)
-        render json: data, status:, include:, root: "data"
+        render json: data, status:, include:, root: 'data'
       end
 
       def render_error(errors: [], status: :ok)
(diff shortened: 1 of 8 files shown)


app/controllers/api/v1/api_controller.rb
```diff
@@ -32,6 +32,16 @@ def log_exception(exception)
...
+      def render_data(data: {}, status: :ok, include: nil)
+        render json: data, status:, include:, root: "data"
+      end
+
+      def render_error(errors: [], status: :ok)
+        render json: {
+          errors:
+        }, status:
+      end
+
```
farhatahmad: Because of how serializing works, we'll need to split these into 2 separate methods. Eventually get rid of `render_json`

app/controllers/api/v1/api_controller.rb
```diff
@@ -32,6 +32,16 @@ def log_exception(exception)
...
+      def render_data(data: {}, status: :ok, include: nil)
+        render json: data, status:, include:, root: 'data'
+      end
+
+      def render_error(errors: [], status: :ok)
+        render json: {
+          errors:
+        }, status:
+      end
+
```
farhatahmad, line 44: Because of how serialization works we'll need to split this method into 2 - eventually we'll get rid of `render_json` entirely

KH-Amir-TN (approved):

PR #3590, opened 2022-06-29
Admin Panel: Minor restructuring to enable nested routing for admin panel tabs.

KH-Amir-TN:
## Description

This PR tends to make a minor restructuring to how admin panel tabs works in GL3.
It tends to extend Admin panel tabs to be synchronized and organized under specific routes in a bidirectional manner (visiting the manage roles route will render the Admin panel with the roles tab selected and manage roles content shown and clicking a tab on the side bar will route the user to the adequate page).
This PR tend to provide:
1. A new structure for the admin panel components.
2. A new approach to work with admin panel tab contents.
Which will enable using the advantages offered by react router for the admin panel which will further simplify development and improve flexibility.


KH-Amir-TN:
This PR compared differs from #3591 in using nested routes and leveraging the IOC (inversion of control) pattern offered by react router to change certain parts of a page only for specific child paths by injecting adequate components in the `<Outlet />` placeholder which will further reduce duplication of the `AdminPanel` skeleton code by centralizing it.

PR #3591, opened 2022-06-29
Admin Panel: Minor restructuring to enable routing for admin panel tabs.

KH-Amir-TN:
## Description

This PR tends to make a minor restructuring to how admin panel tabs works in GL3.
It tends to extend Admin panel tabs to be synchronized and organized under specific routes in a bidirectional manner (visiting the manage roles route will render the Admin panel with the roles tab selected and manage roles content shown and clicking a tab on the side bar will route the user to the adequate page).
This PR tend to provide:
1. A new structure for the admin panel components.
2. A new approach to work with admin panel tab contents.
Which will enable using the advantages offered by react router for the admin panel which will further simplify development and improve flexibility.


KH-Amir-TN:
This PR differs from #3590 in avoiding the use of nested routes and therefore the `<Outlet />` component to reduce the some of the complexity introduced by the use of nested routing (IOC (inversion of control) pattern can make fallowing the flow hard because of its dynamic nature.) but this will introduce some duplication for the `AdminPanel` skeleton code.

farhatahmad (approved):

PR #3597, opened 2022-06-30
Roles edit 2: Added EditRole UI.

KH-Amir-TN:
## Description

Enable users with the right set of privileges to update roles.

This PR completes **2.** 
--
~~1. Add  `RolesController#update` to edit a new role while validating its color to be a HEX 
color.~~
> When permissions are made roles updates should change their permissions.

~~2. Extend the UI on Roles page in the organization settings to select a Role and show
an edit page like with room settings with a form to update name and turn on/off each
 role permission.~~
3. Extend/Add required mutations and feedbacks to integrate with the `UpdateRoleForm` 
with `RolesController#update` API.

 
---
2. User goes to the Organization settings.
[OR]
[OR]

## Testing Steps

> 1. Pull the code.

image

### DIFF
--- a/app/javascript/components/admin/EditRole.jsx
+++ b/app/javascript/components/admin/EditRole.jsx
@@ -1,0 +1,63 @@
+import { ArrowCircleLeftIcon } from '@heroicons/react/outline';
+import React from 'react';
+import {
+  Col, Container, Row, Tab, Card, Breadcrumb,
+} from 'react-bootstrap';
+import { Navigate, useNavigate, useParams } from 'react-router-dom';
+import useRole from '../../hooks/queries/admin/roles/useRole';
+import EditRoleForm from '../forms/admin/EditRoleForm';
+import Spinner from '../shared/stylings/Spinner';
+import AdminNavSideBar from './shared/AdminNavSideBar';
+
+export default function EditRole() {
+  const { roleId } = useParams();
+  const navigate = useNavigate();
+  const { data: role, isError } = useRole(roleId);
+
+  if (isError) {
+    return <Navigate to="/adminpanel/roles" replace />;
+  }
+
+  return (
+    <div id="admin-panel" className="wide-background">
+      <h2 className="my-5"> Administrator Panel </h2>
+      <Card className="border-0 shadow-sm">
+        <Tab.Container activeKey="roles">
+          <Row>
+            <Col sm={3}>
+              <div id="admin-sidebar">
+                <AdminNavSideBar />
+              </div>
+            </Col>
+            <Col sm={9}>
+              <Tab.Content className="p-3 ps-0">
+                <Container>
+                  <Row className="mt-1"><h3>Roles</h3></Row>
+                  <Row className="mt-0 mb-1">
+                    <Col order="first">
+                      <Breadcrumb className="float-start">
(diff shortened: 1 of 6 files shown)


KH-Amir-TN:
# DEPENDS ON #3575

KH-Amir-TN:
# TODO: Split this into several PRs.

app/javascript/hooks/queries/admin/roles/useRole.jsx
```diff
@@ -0,0 +1,10 @@
+import { useQuery } from 'react-query';
+import axios from '../../../../helpers/Axios';
+
+export default function useRole(roleId) {
+  return useQuery(
+    ['getRole', roleId],
+    () => axios.get(`/admin/roles/${roleId}.json`).then((resp) => resp.data.data),
+    { retry: 0 },
```
farhatahmad, line 8: Why `retry: 0`?

app/javascript/components/forms/admin/EditRoleForm.jsx
```diff
@@ -0,0 +1,46 @@
...
+import Form from '../Form';
+import FormControl from '../FormControl';
+import Spinner from '../../shared/stylings/Spinner';
+
+export default function EditRoleForm({ role }) {
+  const { defaultValues } = editRoleFormConfig;
+  defaultValues.name = role.name;
+
+  const methods = useForm(editRoleFormConfig);
+  const { isSubmitting } = methods.formState;
```
farhatahmad, line 15: Shouldn'y we be using the mutation here?

app/javascript/components/admin/EditRole.jsx
```diff
@@ -0,0 +1,63 @@
...
+import Spinner from '../shared/stylings/Spinner';
+import AdminNavSideBar from './shared/AdminNavSideBar';
+
+export default function EditRole() {
+  const { roleId } = useParams();
+  const navigate = useNavigate();
+  const { data: role, isError } = useRole(roleId);
+
+  if (isError) {
+    return <Navigate to="/adminpanel/roles" replace />;
```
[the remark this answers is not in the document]
KH-Amir-TN: It will require more boilerplate code and a use of a `useEffect` binded to the `isError` state.
The `Navigate` component solves all of that.

app/javascript/hooks/queries/admin/roles/useRole.jsx
KH-Amir-TN: React queries by default retries the query on error 3 times.
There's no hope or need to wait for 3 retrials to reroute the user back to `roles` page.
The first error is adequate otherwise they user will have to wait for the retrials delay.

app/javascript/components/forms/admin/EditRoleForm.jsx
KH-Amir-TN: Yup, forgot to change it.

app/javascript/hooks/queries/admin/roles/useRole.jsx
farhatahmad: This isn't something we can just randomly introduce in 1 PR - these kinds of things need to be applied across all mutations. 

Also, there could be an instant when the DB goes down for a milisecond causing the first request to fail. A second request could go through successfully

KH-Amir-TN: Alright, let's address it in all queries in a future PR.

PR #3598, opened 2022-06-30
Manage Users UI Fixes + Basic Admin Edit & Delete with Dropdown

hadicheaito1:
## Description

- Front end changes to make consistent with other tables in admin panel
- 3 dots is a drop-down that allows the user to go to delete or edit user

image
image
image

### DIFF
--- a/app/assets/stylesheets/admin_panel.scss
+++ b/app/assets/stylesheets/admin_panel.scss
@@ -39,4 +39,8 @@
       border-color: whitesmoke !important;
     }
   }
+
+  .dropdown-toggle::after {
+    display: none !important; 
+  }
 }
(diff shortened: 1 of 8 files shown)


scouillard:
image

We lost the toggle arrow on the Navbar

app/assets/stylesheets/admin_panel.scss
```diff
@@ -36,3 +36,7 @@
     }
   }
 }
+// Removes arrow from drop down
+.dropdown-toggle::after {
+  display: none !important;
```
scouillard, line 41: This needs to be scoped

scouillard:
Didn't we say that clicking on the three dots would link to a dedicated "Edit User" page - and not be a dropdown?

hadicheaito1:
Samuel and I discussed it, we agreed there is no need for drop-down since we can put Delete and Edit on the same page

farhatahmad (approved):

PR #3599, opened 2022-06-30
(fix) Frontend: Room Card height

scouillard:
## Description

RoomCards would pile up if one of the card had a different height - mostly because of the length of the title. 
Add CSS class so RoomCards take 100% of the height of the parent element (and the height of the tallest RoomCard).

image


scouillard:
This is more of a quickfix IMO because the RoomCard layout will have to be re-done entirely to be more responsive.

farhatahmad (approved):

PR #3601, opened 2022-06-30
Added Recording and Format Serializers

farhatahmad:

### DIFF
--- a/app/controllers/api/v1/meetings_controller.rb
+++ b/app/controllers/api/v1/meetings_controller.rb
@@ -33,9 +33,9 @@ module Api
           elsif authorized_as_viewer?
             bbb_role = 'Viewer'
           end
-          render_json data: BigBlueButtonApi.new.join_meeting(room: @room, name: params[:name], role: bbb_role), status: :ok
+          render_json data: BigBlueButtonApi.new.join_meeting(room: @room, name: params[:name], role: bbb_role)
         else
-          render_json status: :unauthorized
+          render_error status: :unauthorized
         end
       end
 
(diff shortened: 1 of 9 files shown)


app/controllers/api/v1/meetings_controller.rb
```diff
@@ -33,9 +33,9 @@ def join
           elsif authorized_as_viewer?
             bbb_role = 'Viewer'
           end
-          render_json data: BigBlueButtonApi.new.join_meeting(room: @room, name: params[:name], role: bbb_role), status: :ok
+          render_json data: BigBlueButtonApi.new.join_meeting(room: @room, name: params[:name], role: bbb_role)
```
KH-Amir-TN, line 36: Aren't we removing `render_json`?

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -15,21 +15,7 @@ def index
...
-            length: recording.length,
-            users: recording.users,
-            visibility: recording.visibility,
-            created_at: recording.created_at.strftime('%A %B %e, %Y %l:%M%P'),
-            formats: recording.formats
-          }
-        end
-
-        render_json data: recordings, status: :ok
+        render_data data: recordings
```
KH-Amir-TN, line 18: I think it may be better to make specifying the Sterilizers in use a pattern to easily jump to them when needed.

app/controllers/api/v1/meetings_controller.rb
KH-Amir-TN, line 36: Aren't we removing `render_json`?

KH-Amir-TN (approved):

app/controllers/api/v1/recordings_controller.rb
farhatahmad: I think it's actually cleaner to let the Serializer do its job and figure it out itself - in the cases where we don't want to use the standard way, we can manually specify it

PR #3610, opened 2022-07-05
Added CurrentRoomSerializer

farhatahmad:

### DIFF
--- a/app/javascript/components/rooms/FeatureTabs.jsx
+++ b/app/javascript/components/rooms/FeatureTabs.jsx
@@ -9,16 +9,16 @@ export default function FeatureTabs() {
   return (
     <Row className="pt-5 mx-0">
       <Tabs defaultActiveKey="recordings">
-        <Tab eventKey="recordings" title="Recordings">
+        <Tab eventKey="recordings" title="Recordings" unmountOnExit>
           <RoomRecordings />
         </Tab>
-        <Tab eventKey="presentation" title="Presentation">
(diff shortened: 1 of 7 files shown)


app/javascript/components/rooms/FeatureTabs.jsx
```diff
@@ -9,16 +9,16 @@ export default function FeatureTabs() {
...
-        <Tab eventKey="presentation" title="Presentation">
+        <Tab eventKey="presentation" title="Presentation" unmountOnExit>
           <Presentation />
         </Tab>
-        <Tab eventKey="access" title="Access">
+        <Tab eventKey="access" title="Access" unmountOnExit>
           <SharedAccess />
         </Tab>
-        <Tab eventKey="settings" title="Settings">
+        <Tab eventKey="settings" title="Settings" unmountOnExit>
```
farhatahmad, line 21: This prevents the queries from firing until the tab comes into view.

What ti was doing before was firing all queries for all tabs as soon as the first tab loads

KH-Amir-TN: Couldn't we achieve the same thing by setting the `unmountOnExit` prop on the `Tabs` level?

KH-Amir-TN (approved):

app/javascript/components/rooms/FeatureTabs.jsx
farhatahmad: Yep you're right

KH-Amir-TN (approved):

PR #3613, opened 2022-07-05
Admin users delete

hadicheaito1:
## Description

- Allows admin to delete specific users
- Fixes horizontal spacing issue on tables
- Adds icons to drop-down
## Testing Steps

- Pull changes
- Go to manage users in admin panel
- Click on three dots on any user row
- Click Delete
- Delete the user and make sure it works

image


app/javascript/components/admin/users/DeleteUser.jsx
```diff
@@ -1,30 +1,66 @@
...
   const { userId } = useParams();
+  const { isLoading, data: user } = useUser(userId);
+  const navigate = useNavigate();
+
+  if (isLoading) return <Spinner />;
+
   return (
     <div id="admin-panel" className="wide-background">
       <h2 className="my-5"> Administrator Panel </h2>
       <Card className="border-0 shadow-sm">
```
KH-Amir-TN, line 24: Is having the `admin-table` `AdminNavSideBar` inside of a container inside of the `admin-panel` `Tab.container` is essential?
If it is, then it should be essential for all of the other admin components IMO.

After choosing and merging #3591, it's important to ensure the consistency and avoid making changes to the admin panel skeleton without generalizing those rectifications into all of the other admin panel components.

The skeleton that I found and used is:
```
   <div id="admin-panel" className="wide-background">
      <h2 className="my-5"> Administrator Panel </h2>
      <Card className="border-0 shadow-sm">
        <Tab.Container activeKey="active-key">
          <Row>
            <Col sm={3}>
              <div id="admin-sidebar">
                <AdminNavSideBar />
              </div>
            </Col>
            <Col sm={9}>
              <Tab.Content className="p-3 ps-0">
                <h1>Content</h1>
              </Tab.Content>
            </Col>
          </Row>
        </Tab.Container>
      </Card>
    </div>
```
If the skeleton is valid then we should be only able to edit what is inside of the `Tab.Content` otherwise there should be a problem with our skeleton that needs to be addressed in its separate PR.

app/javascript/components/admin/users/EditUser.jsx
```diff
@@ -13,13 +13,15 @@ import Spinner from '../../shared/stylings/Spinner';
...
+  const navigate = useNavigate();
 
   if (isLoading) return <Spinner />;
+
   return (
     <div id="admin-panel" className="wide-background">
       <h2 className="my-5"> Administrator Panel </h2>
       <Card className="border-0 shadow-sm">
-        <Tab.Content activekey="users">
+        <Tab.Container activekey="users">
```
KH-Amir-TN, line 24: Good catch!

app/javascript/components/admin/users/DeleteUser.jsx
hadicheaito1: Deleted this file so no need to worry about it anymore :sweat_smile:

app/javascript/components/admin/DeleteUserForm.jsx
```diff
@@ -0,0 +1,52 @@
...
+      <p className="text-center"> Are you sure you want to delete {user.name}&rsquo;s account?
+        <br />
+        If you choose to delete this account, it will NOT be recoverable.
+      </p>
+      <Form methods={methods} onSubmit={deleteUser.mutate}>
+        <Stack direction="horizontal" gap={1} className="float-end">
+          <Button variant="primary-reverse" onClick={handleClose}>
+            Close
+          </Button>
+          <Button variant="danger" type="submit" disabled={deleteUser.isSubmitting}>
```
KH-Amir-TN, line 27: There's no `isSubmitting` mutation state, it should be `deleteUser.isLoading`.

app/javascript/hooks/mutations/admins/useDeleteUser.jsx
```diff
@@ -0,0 +1,20 @@
+import { useMutation, useQueryClient } from 'react-query';
+import axios from 'axios';
+import { toast } from 'react-hot-toast';
+
+export default function useDeleteUser(userId) {
+  const queryClient = useQueryClient();
+
+  return useMutation(
+    (data) => axios.delete(`/api/v1/users/${userId}.json`, data),
```
farhatahmad, line 9: Use the helper axios and remove `/api/v1`

app/javascript/hooks/mutations/admins/useDeleteUser.jsx
```diff
@@ -0,0 +1,20 @@
+import { useMutation, useQueryClient } from 'react-query';
+import { toast } from 'react-hot-toast';
+import axios from '../../../helpers/Axios';
+
+export default function useDeleteUser(userId) {
+  const queryClient = useQueryClient();
+
+  return useMutation(
+    (data) => axios.delete(`/users/${userId}.json`, data),
```
KH-Amir-TN, line 9: Awesome, you will still need the `/admin` path, axios custom instance will only set the `baseUrl` to ``/api/v1`.
 =>`admin/users/${userId}.json`

KH-Amir-TN (changes requested):

app/javascript/hooks/mutations/admins/useDeleteUser.jsx
hadicheaito1: This is actually not using the admin user controller, the /admin was never there

KH-Amir-TN: Oh, I think that we should create an `Admin::UsersController#destroy` API.
The `UsersController#destroy` API will only allow users to destroy there authenticated accounts in the future and not any GL users account.

farhatahmad: Yep

app/controllers/api/v1/admin/users_controller.rb
```diff
@@ -25,6 +25,16 @@ def create
           end
         end
 
+        def destroy
+          # TODO: Will need to add additional logic later
+          user = User.find(params[:id])
+          if user.destroy
+            render_data
+          else
+            render_error errors: user.errors.to_a
```
KH-Amir-TN, line 34: **IMHO**, for `render_data` it's common that a success will have a `:ok` status.
So using it in such cases without explicitly define the status code is natural.
But, for `render_error` there's no common error status code each error is specific and self descriptive.
So, I think  for errors even if the  method default status is `:bad_request` it may be better to explicitly express it when calling it -> `render_error errors: user.errors.to_a, status: :bad_request`.
Having the default status value for `render_error` may be better to be omitted.

scouillard (approved):

app/controllers/api/v1/admin/users_controller.rb
hadicheaito1: I agree - but I find this is something we should do to the whole application sometime in the future to make sure its all consistent.

app/controllers/api/v1/admin/users_controller.rb
```diff
@@ -25,6 +25,16 @@ def create
           end
         end
 
+        def destroy
+          # TODO: Will need to add additional logic later
+          user = User.find(params[:id])
```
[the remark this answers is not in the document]
farhatahmad: I would actually replace the way this works. Make `destroy` use a `before_action` and move the logic for `create_room` directly into the method

farhatahmad (approved):

PR #3614, opened 2022-07-05
More serializing

farhatahmad:

### DIFF
--- a/app/controllers/api/v1/api_controller.rb
+++ b/app/controllers/api/v1/api_controller.rb
@@ -14,17 +14,17 @@ module Api
       # Note: The order of each rescue is important (The highest has the lowest priority).
       rescue_from StandardError do |exception|
         log_exception exception
-        render_json errors: [Rails.configuration.custom_error_msgs[:server_error]], status: :internal_server_error
+        render_error errors: [Rails.configuration.custom_error_msgs[:server_error]], status: :internal_server_error
       end
 
       rescue_from ActionController::ParameterMissing do |exception|
         log_exception exception
-        render_json errors: [Rails.configuration.custom_error_msgs[:missing_params]], status: :bad_request
+        render_error errors: [Rails.configuration.custom_error_msgs[:missing_params]], status: :bad_request
       end
 
       rescue_from ActiveRecord::RecordNotFound do |exception|
         log_exception exception
-        render_json errors: [Rails.configuration.custom_error_msgs[:record_not_found]], status: :not_found
+        render_error errors: [Rails.configuration.custom_error_msgs[:record_not_found]], status: :not_found
       end
 
       # TODO: amir - Better Error handling.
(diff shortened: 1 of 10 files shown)


app/controllers/api/v1/shared_accesses_controller.rb
```diff
@@ -16,48 +16,27 @@ def create
...
-          {
-            id: user.id,
-            name: user.name,
-            email: user.email,
-            avatar: user_avatar(user)
-          }
-        end
-
-        render_json data: shareable_users, status: :ok
+        render_data data: User.with_attached_avatar
```
farhatahmad, line 37: Needed to prevent n+1

scouillard (approved):

app/controllers/api/v1/shared_accesses_controller.rb
scouillard: (for future reviewers) with_attached is an ActiveStorage method: https://api.rubyonrails.org/v5.2/classes/ActiveStorage/Attached/Macros.html#method-i-has_many_attached

app/controllers/api/v1/api_controller.rb
```diff
@@ -34,8 +34,18 @@ def log_exception(exception)
...
+          json: data,
+          root: 'data',
+          status:,
+          serializer:,
+          options:
+        }
+        # Allow ActiveModelSerializer to guess which Serializer to use unless manually specified
+        args[:each_serializer] = each_serializer if each_serializer.present?
+
+        render args
```
scouillard, line 48: I don't understand the difference between `serializer` and `each_serializer`

Doesn't ActiveModelSerializer guess which Serializer to use by default?

When would you use `each_serializer` instead of `serializer`

hadicheaito1 (approved):

app/controllers/api/v1/room_settings_controller.rb
```diff
@@ -8,12 +8,11 @@ class RoomSettingsController < ApiController
 
       # GET /api/v1/room_settings/:friendly_id
       def show
-        options = MeetingOption
-                  .joins(:room_meeting_options)
-                  .where(room_meeting_options: { room_id: @room.id })
-                  .select(:name, :value)
+        options = MeetingOption.joins(:room_meeting_options)
+                               .where(room_meeting_options: { room_id: @room.id })
+                               .select(:name, :value)
```
KH-Amir-TN, line 13: Is the `select(:name, :value)` still required?

app/controllers/api/v1/reset_password_controller.rb
```diff
@@ -13,12 +13,12 @@ class ResetPasswordController < ApiController
 
       def create
         # TODO: Log events.
-        return render_json status: :bad_request unless params[:user]
+        return render_error unless params[:user]
```
KH-Amir-TN, line 16: I think it's better to explicitly write the status code for `render_error` even if it's the default.

KH-Amir-TN (approved):

app/controllers/api/v1/reset_password_controller.rb
farhatahmad: I'm thinking of dropping the `default` in a later PR and always specifying it

PR #3618, opened 2022-07-06
Create server room 2..4: Integrated with `Admin::UsersController#create_server_room` API.

KH-Amir-TN:
## Description

Enable users with the right set of privileges to create rooms for all users.
>Note: API restricting and Role permissions will be introduced in upcoming sprints.

This PR completes **2.**, **3.** & **4.**.  
--
~~1. Add `Admin::UsersController#create_server_room` API.~~
~~2. Add the `Create Room` option for the user rows in the `Manage users` page.~~
~~3. Extend  and Re-use the `Create Room modal` to prompt the user for the new room name.~~
~~4. Add required mutation to create the room for the **selected** user.~~

### User story [Create server room]:
---
0. A user with the right level of permissions authenticates to GL.
1. User opens the organization settings.
2. User selects the manage users page.
3. User filters and selects a user.
4. User clicks on the user more options.
5. User clicks create room.
6. User will have a Create Room modal opened.
7. User will be prompted for the new room name.
8. User can cancel the modal or confirm the operation
>Note: no API request will be issued until the room name is filled.
9. User will have the proper feedback.

## Testing Steps

> 1. Pull the code.
> 2. Install the dependencies `bundle install && npm|yarn install`.
> 3. Clean the previous assets build by running `rm app/assets/builds/*` (This won't remove .keep since it's hidden).
> 4. Clean the database and tmp files for a better isolation by running `rails tmp:clear && rails db:schema:cache:clear && rails db:drop && rails db:create && rails db:migrate:with_data`
> 5. Run the linter and specs `bundle exec rubocop --parallel && bundle exec rspec && npx eslint app/javascript/* --ext .jsx,.js`
> 6. Run `./bin/dev` to run the assets builders processes and the Puma server all at once.

PR #3621, opened 2022-07-06
Initial mutation refactoring

farhatahmad:

### DIFF
--- a/app/javascript/components/forms/DeletePresentationForm.jsx
+++ b/app/javascript/components/forms/DeletePresentationForm.jsx
@@ -12,18 +12,18 @@ import useDeletePresentation from '../../hooks/mutations/rooms/useDeletePresenta
 export default function DeletePresentationForm({ handleClose }) {
   const methods = useForm();
   const { friendlyId } = useParams();
-  const { handleDeletePresentation, isLoading } = useDeletePresentation(friendlyId);
+  const deletePresentation = useDeletePresentation(friendlyId);
   return (
     <>
       <p className="text-center"> Are you sure you want to delete this presentation?</p>
-      <Form methods={methods} onSubmit={handleDeletePresentation}>
+      <Form methods={methods} onSubmit={deletePresentation.mutate}>
         <Stack direction="horizontal" gap={1} className="float-end">
           <Button variant="primary-reverse" onClick={handleClose}>
             Close
           </Button>
-          <Button variant="danger" type="submit" disabled={isLoading}>
+          <Button variant="danger" type="submit" disabled={deletePresentation.isLoading}>
             Delete
-            { isLoading && <Spinner /> }
+            { deletePresentation.isLoading && <Spinner /> }
           </Button>
         </Stack>
       </Form>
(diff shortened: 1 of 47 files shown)


app/javascript/components/forms/DeletePresentationForm.jsx
```diff
@@ -12,18 +12,18 @@ import useDeletePresentation from '../../hooks/mutations/rooms/useDeletePresenta
 export default function DeletePresentationForm({ handleClose }) {
   const methods = useForm();
   const { friendlyId } = useParams();
-  const { handleDeletePresentation, isLoading } = useDeletePresentation(friendlyId);
+  const deletePresentation = useDeletePresentation(friendlyId);
```
KH-Amir-TN, line 15: I think it would be better if we destruct that since we're using that pattern heavily with props.
`const { mutation, isLoading } = useDeletePresentation(...);`
It feels more natural to have `disabled={isLoading}` rather than `deletePresentation.isLoading`.

app/javascript/components/forms/DeletePresentationForm.jsx
```diff
@@ -12,18 +12,18 @@ import useDeletePresentation from '../../hooks/mutations/rooms/useDeletePresenta
...
   return (
     <>
       <p className="text-center"> Are you sure you want to delete this presentation?</p>
-      <Form methods={methods} onSubmit={handleDeletePresentation}>
+      <Form methods={methods} onSubmit={deletePresentation.mutate}>
```
KH-Amir-TN, line 19: If we're gonna use the `mutate` directly I think that we need to:
1. Roll it out everywhere else for consistency.
2. Prepare the data on the `axios` callback level instead on the mutation wrapper that we're doing now, i.e. to send room data from the form it needs to be encapsulated inside of a room property of an object so the `axios` callback should be `(room) => axios.post('endpoint',{ room })`. 
3. Use `mutateAsync` for form when using the `isSubmitting` form state or use `mutate` but use the mutation state instead.

scouillard (approved):

app/javascript/components/forms/DeletePresentationForm.jsx
scouillard: I don't understand #3. We should use mutateAsync for forms, and mutate for other mutations?

KH-Amir-TN: > > Prepare the data on the axios callback level instead of doing that on the mutation wrapper function (that's what we're doing now), i.e. to send room data from the form it needs to be encapsulated inside of a room property of an object so the axios callback should be (room) => axios.post('endpoint',{ room }).
> 
> I'm not sure I understand
> 
Check the changes made to this mutation and how in order to remove the need for the handler wrapping function to encapsulate the room data I'm doing that on the mutation function (axios funtion).
https://github.com/bigbluebutton/greenlight/pull/3618/files#diff-58e8c46468b642313d66ecbee89bb9adc3ef4b4a672495ff26506000267db71cR9

KH-Amir-TN: Sorry for the late replies guys 🙏

PR #3623, opened 2022-07-06
(feat): Add status and participants to ServerRooms

scouillard:


app/serializers/server_room_serializer.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class ServerRoomSerializer < RoomSerializer
+  attributes :owner, :status, :participants
+
+  def owner
+    object.user.name
+  end
+
+  def status
```
KH-Amir-TN, line 10: Instead of iterating over the `active_server_rooms` each time for each room why don't we do something similar to the shared logic that @farhatahmad introduced.
Where we set `participants` and `status` as room instance attributes that can be read `attr_reader` and on the `#index` action level we iterate over our filtered list of rooms and simply check wither there's a matching meeting on the server and set those attributes accordingly.
It feels more natural that serializers be concerned only by representing the data and that actions encapsulate the logic behind providing serialalizers with the right data to represent. 
Also this has some **minor** performance implication.

app/controllers/api/v1/admin/server_rooms_controller.rb
```diff
@@ -4,18 +4,19 @@ module Api
   module V1
     module Admin
       class ServerRoomsController < ApiController
+        # GET /api/v1/admin/server_rooms.json
         def index
-          rooms = Room.all.search(params[:search])
+          rooms = Room.all.includes(:user).search(params[:search])
```
KH-Amir-TN, line 9: Is the `all` here required here?

KH-Amir-TN: Good move using `:includes` here it optimizes a lot this action!

app/javascript/hooks/queries/admin/server_rooms/useActiveServerRooms.jsx
```diff
@@ -0,0 +1,7 @@
+import { useQuery } from 'react-query';
+import axios from '../../../../helpers/Axios';
+
+export default function useActiveServerRooms() {
+  return useQuery('getServerRooms', () => axios.get('/admin/server_rooms/active_server_rooms.json')
+    .then((resp) => resp.data.data));
+}
```
farhatahmad, line 7: This needs to be cleaned up a bit to match the new format

app/serializers/server_room_serializer.rb
farhatahmad: Now that I think about it, that might make more sense. Set `attr_reader` in the controller before being passed into the serializer (similar to how shared rooms is done)

spec/controllers/admin/server_rooms_controller_spec.rb
```diff
@@ -20,5 +20,73 @@
       response_room_ids = JSON.parse(response.body)['data'].map { |room| room['friendly_id'] }
```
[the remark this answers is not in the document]
scouillard: What is the issue?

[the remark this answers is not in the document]
KH-Amir-TN: It seems to lack the mocks for the `bbb_api`.

app/serializers/server_room_serializer.rb
scouillard: Putting the logic in the serializer definitely feels a bit weird.
I'm not familiar with attr_accessor - and not sure of the implications - but I'm reading on it.

hadicheaito1 (approved):

hadicheaito1:
> Is it OK to return room.participants: null if room.status == "Not Running" ? Or maybe return room.participants **only** if room.status == "Active" ?

I think the way you have it right now is fine

spec/controllers/admin/server_rooms_controller_spec.rb
```diff
@@ -17,8 +17,77 @@
...
+    end
+
+    it 'returns the server status as not running if BBB server does not return any active meetings' do
+      create(:room)
+
+      allow_any_instance_of(BigBlueButtonApi).to receive(:active_meetings).and_return([])
+      get :index
+      expect(JSON.parse(response.body)['data'][0]['status']).to eql('Not Running')
+    end
+  end
```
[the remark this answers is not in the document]
scouillard: Participants is null if the room is "Not Running" right now.
IMO this doesn't need to be tested as things are happening only if room.status == "Active".

app/controllers/api/v1/admin/server_rooms_controller.rb
```diff
@@ -4,18 +4,24 @@ module Api
...
+          active_rooms_ids = []
+          participants = []
 
-          rooms.map! do |room|
-            {
-              friendly_id: room.friendly_id,
-              name: room.name,
-              owner: User.find(room.user_id).name
-            }
+          active_rooms.each do |active_room|
```
KH-Amir-TN, line 14: I might create one `active_rooms_hash` data structure  that will store only the `meetingID => participantCount`.
The first advantage is that will have only one data structure that can do it all and the second advantage is that hashes are indexed data structures by nature meaning that finding an element doesn't require the traversal of all of its predecessors.
That means for the same number of active meetings a hash will always find the meeting in a constant time (in theory) while the array resource consumption will increase exponentially depending on the input size (This may not be crucial for our use case).
[More.](https://blog.sessionstack.com/how-javascript-works-arrays-vs-hash-tables-ab769bf84a2d#:~:text=Hash%20tables%20tend%20to%20be,the%20key%20and%20insert%20it.)
image

KH-Amir-TN (approved):

KH-Amir-TN (approved):
# 🔥 🔥 🔥

farhatahmad (approved):

app/controllers/api/v1/admin/server_rooms_controller.rb
```diff
@@ -6,18 +6,22 @@ module Admin
...
-          rooms = Room.all.search(params[:search])
-
-          rooms.map! do |room|
-            {
-              friendly_id: room.friendly_id,
-              name: room.name,
-              owner: User.find(room.user_id).name
-            }
+          rooms = Room.includes(:user).search(params[:search])
+          active_rooms = BigBlueButtonApi.new.active_meetings
```
scouillard: Using the BigBlueButtonApi `active_meetings` method is good because it fetches all the active meetings in a single API call rather than making individual requests for each room, which would be much slower and inefficient.

app/controllers/api/v1/admin/server_rooms_controller.rb
```diff
@@ -6,18 +6,22 @@ module Admin
...
+          active_rooms_hash = {}
+
+          active_rooms.each do |active_room|
+            active_rooms_hash[active_room[:meetingID]] = active_room[:participantCount]
+          end
```
scouillard: The conversion of the `active_rooms` array into an `active_rooms_hash` with the `meetingID` as key is good because it makes the subsequent matching operations much faster as arrays are not efficient for searching.

PR #3626, opened 2022-07-07
Room settings: Enable All users join as moderators setting.

KH-Amir-TN:
## Description

Enable users to join as moderators on rooms with the **All users join as moderators** setting enabled.

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/models/meeting_option.rb
+++ b/app/models/meeting_option.rb
@@ -12,4 +12,13 @@ class MeetingOption < ApplicationRecord
       .where(room_meeting_options: { room_id: })
       .pluck(:name, :value)
   end
+
+  def self.get_value(name:, room_id:)
+    joins(:room_meeting_options)
+      .select(:value)
+      .find_by(
+        name:,
+        room_meeting_options: { room_id: }
+      )
+  end
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -26,6 +26,10 @@ class Room < ApplicationRecord
     all.to_a
   end
 
+  def anyone_joins_as_moderator?
+    MeetingOption.get_value(name: 'glAnyoneJoinAsModerator', room_id: id)&.value == 'true'
+  end
+
   private
 
   def set_friendly_id
--- a/spec/controllers/meetings_controller_spec.rb
+++ b/spec/controllers/meetings_controller_spec.rb
@@ -109,6 +109,24 @@ RSpec.describe Api::V1::MeetingsController, type: :controller do
       expect(response).to have_http_status(:ok)
     end
 
+    it 'joins as moderator if "glAnyoneJoinAsModerator" setting enabled and access code provided' do
+      allow_any_instance_of(Room).to receive(:anyone_joins_as_moderator?).and_return(true)
+      room = create(:room, user:, viewer_access_code: 'AAA', moderator_access_code: 'BBB')
+
+      expect_any_instance_of(BigBlueButtonApi).to receive(:join_meeting).with(room:, name: 'Someone', role: 'Moderator')
+      get :join, params: { friendly_id: room.friendly_id, name: 'Someone', access_code: 'AAA' }
+      expect(response).to have_http_status(:ok)
+    end
+
+    it 'joins as moderator if "glAnyoneJoinAsModerator" setting enabled and access code NOT provided' do
+      allow_any_instance_of(Room).to receive(:anyone_joins_as_moderator?).and_return(true)
+      room = create(:room, user:, viewer_access_code: 'AAA', moderator_access_code: 'BBB')
+
+      expect_any_instance_of(BigBlueButtonApi).to receive(:join_meeting).with(room:, name: 'Someone', role: 'Moderator')
+      get :join, params: { friendly_id: room.friendly_id, name: 'Someone' }
+      expect(response).to have_http_status(:ok)
+    end
+
     it 'returns unauthorized if the access code is wrong' do
       room = create(:room, user:, viewer_access_code: 'AAA', moderator_access_code: 'BBB')
 
--- a/spec/models/room_spec.rb
+++ b/spec/models/room_spec.rb
@@ -29,6 +29,31 @@ RSpec.describe Room, type: :model do
     end
   end
 
+  describe '#anyone_joins_as_moderator?' do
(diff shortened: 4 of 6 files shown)


app/models/room.rb
```diff
@@ -53,4 +57,8 @@ def create_meeting_options
       RoomMeetingOption.create(room: self, meeting_option: option, value: option.default_value)
     end
   end
+
+  def find_meeting_option_by_name(name)
+    room_meeting_options.joins(:meeting_option).find_by('meeting_option.name': name)
+  end
```
farhatahmad, line 63: I would move this method to `MeetingOption`

spec/controllers/meetings_controller_spec.rb
```diff
@@ -109,6 +109,15 @@
...
     end
 
+    it 'joins as moderator if "glAnyoneJoinAsModerator" setting enabled' do
+      allow_any_instance_of(Room).to receive(:anyone_joins_as_moderator?).and_return(true)
+      room = create(:room, user:, viewer_access_code: 'AAA', moderator_access_code: 'BBB')
+
+      expect_any_instance_of(BigBlueButtonApi).to receive(:join_meeting).with(room:, name: 'Someone', role: 'Moderator')
+      get :join, params: { friendly_id: room.friendly_id, name: 'Someone', access_code: 'AAA' }
+      expect(response).to have_http_status(:ok)
+    end
```
farhatahmad, line 119: I would add another test case where there is no access code

app/models/room.rb
KH-Amir-TN: Wouldn't that make `  anyone_joins_as_moderator?`:
```
  def anyone_joins_as_moderator?
    MeetingOption.find_room_meeting_optionby_name(name: 'glAnyoneJoinAsModerator', room_id: id)&.value == 'true'
  end
```
Is that preferred?

spec/controllers/meetings_controller_spec.rb
KH-Amir-TN: To explicitly ensure that passing `access_code` doesn't change the role to viewer?

app/models/meeting_option.rb
```diff
@@ -12,4 +12,11 @@ def self.bbb_options(room_id:)
       .where(room_meeting_options: { room_id: })
       .pluck(:name, :value)
   end
+
+  def self.find_room_meeting_option_by_name(name:, room_id:)
+    joins(:room_meeting_options)
+      .where(room_meeting_options: { room_id: })
+      .select('meeting_options.name', 'room_meeting_options.value')
```
KH-Amir-TN, line 19: This is due to `ActiveRecord` having a hard time to deal with joins so it assumes that the result since it's initiated by the `MeetingOption` model it will result in an instance of `MeetingOption` so it will filter any irrelevant attributes such us `value`.

spec/models/room_spec.rb
```diff
@@ -29,6 +29,45 @@
     end
   end
 
+  describe '#anyone_joins_as_moderator?' do
+    let!(:room) { create(:room) }
+
+    let(:fake_room_meeting_option) do
+      Class.new do
```
KH-Amir-TN, line 36: That may look complex but it was the simplest way I could come up with.
The problem is due to the `anyone_joins_as_moderator?` expecting an object with a value attribute (property).
Mocking the result as simple hash will not doing it because of `Hash` nature in ruby, their properties are not a class attributes that is why we're accessing them through the the brackets `[]`.
And not mocking the `MeetingOption::find_room_meeting_option_by_name` and creating the records on the database will solve this but will make this an integration test rather then a unit test since it will further test the `find_room_meeting_option_by_name` which has its own unit tests.
So, I created a fake anonymous class that has the expected attributes mocking the result of calling `find_room_meeting_option_by_name`.

app/models/meeting_option.rb
```diff
@@ -12,4 +12,11 @@ def self.bbb_options(room_id:)
       .where(room_meeting_options: { room_id: })
       .pluck(:name, :value)
   end
+
+  def self.find_room_meeting_option_by_name(name:, room_id:)
```
farhatahmad, line 16: I think we might want to name this something a little simpler. Maybe `get_value` or something along those lines

app/models/meeting_option.rb
```diff
@@ -12,4 +12,11 @@ def self.bbb_options(room_id:)
...
+    joins(:room_meeting_options)
+      .where(room_meeting_options: { room_id: })
+      .select('meeting_options.name', 'room_meeting_options.value')
+      .find_by(name:)
```
farhatahmad, line 20: Can't we do something like

```
joins(:room_meeting_options)
  .find_by(
    name:,
    room_meeting_options: { room_id: }
  )
  .select(:value)
```

scouillard (dismissed):

app/models/meeting_option.rb
KH-Amir-TN: It's returning more than value, what you think about naming it `get_meeting_option`?

spec/models/room_spec.rb
farhatahmad: You can just mock the request with `OpenStruct`

`allow(MeetingOption).to receive(:get_value).and_return(OpenStruct.new(value: 'true'))`

KH-Amir-TN: > You can just mock the request with `OpenStruct`
> 
> `allow(MeetingOption).to receive(:get_value).and_return(OpenStruct.new({value: 'true'}))`

That's exactly what I needed, didn't know about the `OpenStruct` class.
Thanks for sharing 🙏!
> For anyone that wants to discover this here's an article that I found worthy [HERE](https://medium.com/rubycademy/openstruct-in-ruby-ab6ba3aff9a4).

KH-Amir-TN: Rubocop doesn't allow the user of `OpenStructs` and favorizes the use of much stricter structures such as `Structs`, `Hashes` and `Doubles` [CHECK HERE](https://docs.rubocop.org/rubocop/cops_style.html#styleopenstructuse).
Therefore, switched to `instance_doubles` [MORE HERE](https://relishapp.com/rspec/rspec-mocks/v/3-11/docs/verifying-doubles/using-an-instance-double).

PR #3647, opened 2022-07-13
Meeting Starter now updates last session

scouillard:
## Description

`room.last_session` now gets updated with bbb response `createDate` when a Meeting is started. 
Somewhat in conjunction with the room status, we can now display the date of the last session.

## Testing Steps

1. Create a few Rooms
2. Go to Admin/Server Rooms
3. Start/End some Meetings. 
4. In Admin/Server Rooms, the list will update as the Rooms toggled between active and not running. 

- If a Meeting has never been started in the Room, it should display "No meeting yet"
- If a Meeting is currently active, it should display **Started: datetime**
- If a Room had a Meeting, but it is currently not running, it should display **Ended: datetime**

image


scouillard:
Should we duplicate the bbb server meeting starter response (below) in RSpec for both `controllers/meeting_controller_spec `and `service/meeting_starter_spec` ? Or only give the required object, which is `{ createTime: 123456789 }` 

```
def meeting_starter_response
    {
      returncode: true,
      meetingID: 'hulsdzwvitlk1dbekzxdprshsxmvycvar0jeaszc',
      attendeePW: '12345',
      moderatorPW: '54321',
      createTime: 1_389_464_535_956,
      hasBeenForciblyEnded: false,
      messageKey: '',
      message: ''
    }
end
```

app/serializers/server_room_serializer.rb
```diff
@@ -1,9 +1,16 @@
...
 class ServerRoomSerializer < RoomSerializer
-  attributes :owner, :status, :participants
+  attributes :owner, :status, :participants, :last_session
 
   def owner
     object.user.name
   end
+
+  def last_session
+    return 'No meeting yet' if object.last_session.nil?
```
KH-Amir-TN, line 11: I think we shouldn't be preparing the representation of our data in the backend.
So, `last_session` should have for example `nil|datetime` or `nil|integer`.
The frontend should be the only entity concerned with mapping these values to its adequate represenatation.
If it's falssy then the client app  will show **No meeting yet** and do any I18n otherwise it will be in charge to display the format of the date and time according to the user browser timezone and language (upcoming).

app/serializers/server_room_serializer.rb
```diff
@@ -1,9 +1,16 @@
...
+
+    last_session_date = object.last_session.strftime('%B %e, %Y')
+    object.status == 'Active' ? "Started: #{last_session_date}" : "Ended: #{last_session_date}"
```
KH-Amir-TN, line 14: I know that `status` is out of the scope of this PR.
But, just to mention a question that came to my mind; Is the status non binary? 
Is there a status other than `Active|Not Running`?
If not I think it may make more sense to switch the `status` to `active` Boolean attribute and leave this responsibility for the frontend to map it to an informational output such us `Active`.

app/services/meeting_starter.rb
```diff
@@ -17,7 +17,9 @@ def call
 
     retries = 0
     begin
-      BigBlueButtonApi.new.start_meeting room: @room, options: options, presentation_url: @presentation_url
+      meeting = BigBlueButtonApi.new.start_meeting room: @room, options: options, presentation_url: @presentation_url
+
+      @room.update!(last_session: Time.zone.at(meeting[:createTime] / 1000)) unless meeting[:messageKey] == 'duplicateWarning'
```
KH-Amir-TN, line 22: You don't wanna go that way with `meeting[:messageKey] == 'duplicateWarning'` trust me 😆.
It's safer to check for possible mismatches and update the room, `unless @room.last_session == meeting_create_time`.
[MORE](https://github.com/bigbluebutton/greenlight/pull/3000).

KH-Amir-TN:
> Should we duplicate the whole hardcoded bbb server meeting starter response (below) in RSpec for both `controllers/meeting_controller_spec `and `service/meeting_starter_spec` ? Or only give the required object, which is `{ createTime: 123456789 }` ?
> 
> I think it makes sense to duplicate the whole response in both.
> 
> ```
> def meeting_starter_response
>     {
>       returncode: true,
>       meetingID: 'hulsdzwvitlk1dbekzxdprshsxmvycvar0jeaszc',
>       attendeePW: '12345',
>       moderatorPW: '54321',
>       createTime: 1_389_464_535_956,
>       hasBeenForciblyEnded: false,
>       messageKey: '',
>       message: ''
>     }
> end
> ```

IMO I think it's better to have a stub that returns what we're expecting from **BBB** even if it contains some attributes that we don't need in some tests.
That will give us these major advantages:
1. Several tests under the same test group may share that same mocked data.
2. We can better spot issues in our logic to avoid having a unit/integration tests that works on such data but an e2e that fails on real cases.
3. It's less error prone to do little to no modification to what a real **BBB** API would return to avoid braking its expected format.
 
So, I'd personally go with duplicating it in these two spec files.

app/services/meeting_starter.rb
scouillard: I don't usually take stuff from v2, but I thought this one made sense and was fail proof.

KH-Amir-TN: Yeah, you got reason I'd do that way myself.
That one caused us some headache when working on `v2` 🤕

app/serializers/server_room_serializer.rb
scouillard: I agree.
We should talk about this next meeting.

app/services/meeting_starter.rb
```diff
@@ -17,7 +17,9 @@ def call
 
     retries = 0
     begin
-      BigBlueButtonApi.new.start_meeting room: @room, options: options, presentation_url: @presentation_url
+      meeting = BigBlueButtonApi.new.start_meeting room: @room, options: options, presentation_url: @presentation_url
+
+      @room.update!(last_session: Time.zone.at(meeting[:createTime] / 1000))
```
farhatahmad, line 22: Why the `/ 1000`?

scouillard: We need to convert from milliseconds to seconds to use Time.at

app/serializers/server_room_serializer.rb
scouillard: What do you think about this @farhatahmad ? 
Could there eventually be a room status that is neither Active or Not Running ?
Should we make a switch from status:string to active:boolean ?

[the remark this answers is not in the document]
scouillard: Yep it's weird - my bad.

app/javascript/components/admin/server_rooms/ServerRoomRow.jsx
```diff
@@ -8,26 +8,44 @@ import Modal from '../../shared/Modal';
 import DeleteRoomForm from '../../forms/DeleteRoomForm';
 
 export default function ServerRoomRow({ room }) {
-  const { friendly_id: friendlyId } = room;
+  const {
+    friendly_id: friendlyId, name, owner, last_session: lastSession, status, participants,
+  } = room;
```
scouillard, line 13: I'm not sure about destructuring Room here

app/javascript/components/admin/server_rooms/ServerRoomRow.jsx
```diff
@@ -13,25 +13,37 @@ import useRoomStatus from '../../../hooks/queries/rooms/useRoomStatus';
...
+  };
+
   return (
     <tr className="align-middle text-muted border border-2">
       <td className="border-end-0">
         <Stack>
-          <span className="text-dark fw-bold"> {room.name} </span>
-          <span> Ended: </span>
+          <span className="text-dark fw-bold"> {name} </span>
+          {renderLastSession()}
```
farhatahmad, line 40: I would keep the span here and just return text instead `<span> ${ lastSession } </span>`

app/serializers/server_room_serializer.rb
```diff
@@ -1,9 +1,13 @@
...
 class ServerRoomSerializer < RoomSerializer
-  attributes :owner, :status, :participants
+  attributes :owner, :status, :participants, :last_session
 
   def owner
     object.user.name
   end
+
+  def last_session
+    object.last_session&.strftime('%B %e, %Y')
```
farhatahmad, line 11: I would use the same `strftime` as is used in `created_at`

app/services/meeting_starter.rb
farhatahmad: Hmm not really a fan of that. Here's 2 alternatives:
```
Time.zone.at(0, meeting[:createTime], :millisecond)
DateTime.strptime(meeting[:createTime].to_s, "%Q")
```

scouillard: If we choose to display the createDate in hours/seconds, we should remove the "Started:" and "Ended:" and maybe replace it with "Last Session:" as we are kind of cheating by toggling started/ended depending on the Room status.

app/javascript/components/admin/server_rooms/ServerRoomRow.jsx
scouillard: image

scouillard:
I changed the display from "Started:"/"Ended:" to "Current Session:"/"Last Session:" as we are now displaying the hour.minute of the createDate, and we are not fetching the endDate of the meeting.

KH-Amir-TN (approved):

spec/services/meeting_starter_spec.rb
```diff
@@ -26,9 +26,10 @@
 
   describe '#call' do
     it 'calls BigBlueButtonApi with the right params' do
-      expect_any_instance_of(BigBlueButtonApi)
+      allow_any_instance_of(BigBlueButtonApi)
         .to receive(:start_meeting)
         .with(room:, options:, presentation_url:)
+        .and_return(meeting_starter_response)
```
farhatahmad, line 32: Making this change actually makes this test do nothing. `expect` and `allow` do 2 very different things

spec/services/meeting_starter_spec.rb
```diff
@@ -60,13 +62,40 @@
...
+
+    it 'updates the last session date when a meeting is started' do
+      allow_any_instance_of(BigBlueButtonApi)
+        .to receive(:start_meeting)
+        .with(room:, options:, presentation_url:)
+        .and_return(meeting_starter_response)
+
+      service.call
+
+      expect(room.last_session).to eql(DateTime.strptime(1_389_464_535_956.to_s, '%Q').utc)
```
farhatahmad, line 83: I wouldn't hardcode the startTime here. 

`expect(room.last_session).to eql(DateTime.strptime(meeting_starter_response[:createTime].to_s, '%Q').utc)`

spec/services/meeting_starter_spec.rb
scouillard: It's a cop that made me change it.

`Prefer `allow_any_instance_of` over `expect_any_instance_of` when configuring a response.`

scouillard: I removed the return.

app/services/meeting_starter.rb
```diff
@@ -17,7 +17,9 @@ def call
 
     retries = 0
     begin
-      BigBlueButtonApi.new.start_meeting room: @room, options: options, presentation_url: @presentation_url
+      meeting = BigBlueButtonApi.new.start_meeting room: @room, options: options, presentation_url: @presentation_url
+
+      @room.update!(last_session: DateTime.strptime(meeting[:createTime].to_s, '%Q'))
```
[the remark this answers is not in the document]
scouillard: It was a Time object initially I think but there we had to divide by 1000 - after trying a couple of things, this looks right.

spec/services/meeting_starter_spec.rb
```diff
@@ -26,6 +26,10 @@
 
   describe '#call' do
     it 'calls BigBlueButtonApi with the right params' do
+      allow_any_instance_of(BigBlueButtonApi)
+        .to receive(:start_meeting)
+        .and_return(meeting_starter_response)
+
```
farhatahmad, line 32: This is fine for this PR but we may want to move this to a `before` similar to `MeetingsControllerSpec`

farhatahmad (approved):

spec/services/meeting_starter_spec.rb
scouillard: Right. Might as well.

PR #3649, opened 2022-07-13
Added ability to fetch settings

farhatahmad:
## Description

Added some examples of how it will be used. Tests will come in the next PR

### DIFF
--- a/app/controllers/api/v1/site_settings_controller.rb
+++ b/app/controllers/api/v1/site_settings_controller.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class SiteSettingsController < ApiController
+      # GET /api/v1/site_settings/:name
+      def show
+        render_json data: SettingGetter.new(setting_name: params[:name], provider: 'greenlight').call # TODO: - ahmad: fix provider
+      end
+    end
+  end
+end
--- a/app/javascript/components/rooms/FeatureTabs.jsx
+++ b/app/javascript/components/rooms/FeatureTabs.jsx
@@ -4,20 +4,30 @@ import RoomRecordings from '../recordings/RoomRecordings';
 import Presentation from './Presentation';
 import SharedAccess from '../shared_accesses/SharedAccess';
 import RoomSettings from '../room_settings/RoomSettings';
+import useSiteSetting from '../../hooks/queries/site_settings/useSiteSetting';
 
 export default function FeatureTabs() {
+  const { data: shareRoomEnabled } = useSiteSetting('ShareRooms');
+  const { data: preuploadEnabled } = useSiteSetting('PreuploadPresentation');
+
   return (
     <Row className="pt-5 mx-0">
       <Tabs defaultActiveKey="recordings" unmountOnExit>
         <Tab eventKey="recordings" title="Recordings">
           <RoomRecordings />
         </Tab>
-        <Tab eventKey="presentation" title="Presentation">
-          <Presentation />
-        </Tab>
-        <Tab eventKey="access" title="Access">
-          <SharedAccess />
-        </Tab>
+        { preuploadEnabled
+          && (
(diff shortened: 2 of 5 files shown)


app/javascript/components/rooms/FeatureTabs.jsx
```diff
@@ -4,20 +4,30 @@ import RoomRecordings from '../recordings/RoomRecordings';
...
+          <Tab eventKey="presentation" title="Presentation">
+            <Presentation />
+          </Tab>
+          )}
+        { shareRoomEnabled
+          && (
+          <Tab eventKey="access" title="Access">
+            <SharedAccess />
+          </Tab>
+          )}
```
farhatahmad, line 30: This is an example of how it could be used in the front-end

hadicheaito1 (approved):

scouillard (approved):

app/javascript/components/rooms/FeatureTabs.jsx
KH-Amir-TN: I think that we can approach this differently without having to make a full request for each setting each time we visit a room.

KH-Amir-TN: IMO these settings should be part of the server state given by the `useSessions`, whcih will cache these configs in the `queryCache` and `react-query` will make sure to keep it in sync.

The `AuthProvider` or a new concerned provider will then fetch these configs for us to access when needed through its context.

The user role, its permissions and the site settings (like theme ...) designs a server state also.

Or we can make an endpoint that fetches them in a single request.

app/controllers/api/v1/site_settings_controller.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class SiteSettingsController < ApiController
+      # GET /api/v1/site_settings/:name
+      def show
+        render_json data: SettingGetter.new(setting_name: params[:name], provider: 'greenlight').call # TODO: - ahmad: fix provider
```
KH-Amir-TN, line 8: I feel that this is similar somehow to the `MeetingOption` and `RoomMeetingOption` case from a data point of view.
With that said, I think that we can (even for consistency) do the join in a method inside of `Setting|SiteSetting` model `get_setting(name:, provider:)`.

app/javascript/components/rooms/FeatureTabs.jsx
```diff
@@ -4,20 +4,30 @@ import RoomRecordings from '../recordings/RoomRecordings';
 import Presentation from './Presentation';
 import SharedAccess from '../shared_accesses/SharedAccess';
 import RoomSettings from '../room_settings/RoomSettings';
+import useSiteSetting from '../../hooks/queries/site_settings/useSiteSetting';
 
 export default function FeatureTabs() {
+  const { data: shareRoomEnabled } = useSiteSetting('ShareRooms');
```
KH-Amir-TN, line 10: I believe that `useSiteSetting` will be needed on the `EditSiteSetting` form in the admin panel and it should return a list of all site settings with their `:name` and `:value`.

app/controllers/api/v1/site_settings_controller.rb
KH-Amir-TN: Do we need a `SiteSettingController#show` API?
Because according to the mockups we'd be having a single page where we can edit all site settings individually like with the room settings.
In such case and because the `siteSetting` model isn't that complex to require a specific endpoint to fetch it, we can create only `SiteSettingController#index` that fetches all `siteSettings` for the provider with settings `:name` and `:value`.
And we'd have also the `Admin::SiteSettingController#update` that will update a specific site setting.

farhatahmad: Whoever is creating the page in the admin panel can create a `SiteSettings#index` that will return all those values

PR #3652, opened 2022-07-13
Admin to join or start a server room

scouillard:
## Description

Admin has option to join room if room is active.
Admin has option to start room if room is not running.

Nothing new on the backend, been re-using the hooks from past features.

Note: 
The join feature will not work if the room has `access_code` enabled.
Will need to eventually implement a bypass in MeetingController.
Maybe something like:
```
def authorized_as_moderator?
<current logic...> || @current_user.role_id == 3
end
```

## Testing Steps

- Create a few rooms
- Start a meeting
- Go to Admin/Server Rooms
- Join a meeting if it's already active, if not you should be able to start it.

image


hadicheaito1 (approved):
Looks good!

PR #3657, opened 2022-07-14
Roles Mapping 0: Added Registration UI and dynamic form.

KH-Amir-TN:
## Description

Enable users with the right set of privileges to edit roles mappers.

This PR completes **0.** 
--
~~0. Add Registration UI and dynamic form.~~
1. Add `SiteSettings#index`.
2. Add `Admin::SiteSettings#update` .
3. Integrate with the APIs.
4.Extend `UsersController#create` to infer the user role.

### User story [Role Mapping]:
---
0. A user with the right set of permissions signs in.
1. User navigates to the Manage site settings page in administration panel.
2. User selects the Registration tab.
3. User can edit the list of roles and their corresponding email suffixes.
5. User can update the list and have adequate feedbacks from the client app.
Note: There's no checks for the list content, it's a string comma sperated list
that can be empty or filled.
## Testing Steps

> 1. Pull the code.

image
image

### DIFF
--- a/app/javascript/components/admin/site_settings/Registration.jsx
+++ b/app/javascript/components/admin/site_settings/Registration.jsx
@@ -1,7 +1,37 @@
-import React from 'react';
+import React, { useState } from 'react';
+import { Alert, Container, Row } from 'react-bootstrap';
+import RegistrationForm from '../../forms/admin/RegistrationForm';
+
+const DUMMY_ROLES_MAP = [
+  { name: 'User', suffix: 'users.com' },
+  { name: 'Admin', suffix: 'admins.com' },
+  { name: 'Teacher', suffix: 'teachers.com' },
+  { name: 'Presenter', suffix: 'presenters.com' },
+];
 
 export default function Appearance() {
+  const [showInfo, setShowInfo] = useState(true);
+
   return (
-    <p>Registration</p>
+    <Container>
+      {
+        showInfo && (
+          <Row className="mt-2 mb-0">
+            <Alert variant="light" onClick={() => setShowInfo(false)} dismissible>
+              <Alert.Heading>Roles Mapping By Email</Alert.Heading>
+              <p className="text-muted">
+                Map a user to a role based on their email address suffix.<br />
+                Example:<br />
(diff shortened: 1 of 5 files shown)


KH-Amir-TN:
@farhatahmad @scouillard @hadicheaito1 
https://user-images.githubusercontent.com/29759616/179072467-7dbd9ef7-9b78-4121-a00a-4e17c3a2af1f.mp4

KH-Amir-TN:
:NEW STYLE
@farhatahmad @hadicheaito1 @scouillard 

https://user-images.githubusercontent.com/29759616/179227556-b18ef815-18ae-4288-928e-0195b81e295a.mp4

KH-Amir-TN:
> 

I see what you mean, according to what I understood from @farhatahmad we're fallowing the KISS (Keep It Simple & Stupid) approach with all site settings.
All site settings should have a `:name` and a ``:value` of type string only and share the same APIs (`#index #update...`) without any dependency on the meaning behind any site setting in particular.

So making the role mapping rules fully in sync with the `Roles` table may require braking that simplicity constraint by giving a meaning to roles mapping rules that distinguish it from the rest of site settings.

According to that, we can only at the moment have generic APIs for site settings that fits all of the different settings needs so we cannot afford to doing things relative to some site settings that irrelevant to others.

But we'd have to discuss this more during our next meeting,

KH-Amir-TN:
# REBASED ON ad6fdc65e74ac5aa7f557ace81c71b3152f19470

hadicheaito1 (approved):

KH-Amir-TN:
> Hmm I played around with this a bit and I think I agree with @scouillard - it seems that we might be giving too much control to users here. I like the mockup that @scouillard did - 1 input for each role

I like it either, my only problem is the how?

At the moment and with the current configuration I don't see any clean way of achieving this without introducing a great deal of complexity and conforming with the constraint of site settings which is **all site settings must have the same data representation and should be treated the same way through the same APIs**.

If that's what we're going with (which I personally like) there should be some changes to make that needs to be discussed.

PR #3660, opened 2022-07-15
Fix a spec in Controllers/ServerRooms

scouillard:
## Description

The bbb api call stub had no return, and would break the RSpec if the bbb endpoint environment variable was empty `BIGBLUEBUTTON_ENDPOINT=`

Add an empty array return on the stub + some small layout touch-ups on the test itself. 

The test itself does not seem that useful TBH.

### DIFF
--- a/spec/controllers/admin/server_rooms_controller_spec.rb
+++ b/spec/controllers/admin/server_rooms_controller_spec.rb
@@ -8,18 +8,17 @@ RSpec.describe Api::V1::Admin::ServerRoomsController, type: :controller do
   end
 
   describe '#index' do
-    it 'returns all the Server Rooms' do
+    it 'returns all the Server Rooms from all users' do
       user_one = create(:user)
       user_two = create(:user)
 
-      user_one_rooms = create_list(:room, 5, user_id: user_one.id)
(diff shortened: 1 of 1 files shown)


spec/controllers/admin/server_rooms_controller_spec.rb
```diff
@@ -8,18 +8,17 @@
...
+      create_list(:room, 5, user_id: user_one.id)
+      create_list(:room, 5, user_id: user_two.id)
 
+      allow_any_instance_of(BigBlueButtonApi).to receive(:active_meetings).and_return([])
       get :index
-      allow_any_instance_of(BigBlueButtonApi).to receive(:active_meetings)
-      response_room_ids = JSON.parse(response.body)['data'].map { |room| room['friendly_id'] }
-      expect(response_room_ids).to match_array(rooms.pluck(:friendly_id))
+      expect(JSON.parse(response.body)['data'].map { |room| room['friendly_id'] })
+        .to match_array(Room.all.pluck(:friendly_id))
```
KH-Amir-TN, line 21: Is the `all` required?

KH-Amir-TN (approved):

farhatahmad (approved):

PR #3666, opened 2022-07-18
Share Rooms Site Setting

hadicheaito1:
## Description

- Toggle for Share Rooms site setting now works
- When it is set to off, the Shared Access tab is not shown on to Room page
## Testing Steps

- Create room 
- Go to Admin Panel
- Go to Site Settings
- Toggle on/off
- Check to see that share access tab on room page appears/disappears based on value

image

## TODO
- Some front end touch-ups will be made once more settings are added to page

### DIFF
--- a/app/controllers/api/v1/admin/site_settings_controller.rb
+++ b/app/controllers/api/v1/admin/site_settings_controller.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module Admin
+      class SiteSettingsController < ApiController
+        def update
+          SiteSetting.find_by(setting_id: params[:settingId]).update(value: params[:settingValue].to_s)
+
+          render_json
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/api/v1/admin/site_settings_controller.rb
+++ b/app/controllers/api/v1/admin/site_settings_controller.rb
@@ -17,8 +17,8 @@ module Api
           SiteSetting
             .joins(:setting)
             .where(provider: 'greenlight')
-            .where(setting: { name: params[:settingName] })
-            .update(value: params[:settingValue].to_s)
+            .where(setting: { name: params[:name] })
+            .update(value: params[:value].to_s)
           render_json
         end
       end
--- a/app/controllers/api/v1/admin/site_settings_controller.rb
+++ b/app/controllers/api/v1/admin/site_settings_controller.rb
@@ -4,6 +4,15 @@ module Api
   module V1
     module Admin
       class SiteSettingsController < ApiController
+        def index
+          data = Setting.joins(:site_settings)
+                        .where(site_settings: { provider: 'greenlight' })
+                        .pluck(:name, :value)
+                        .to_h
+
+          render_json data:
+        end
+
         def update
           SiteSetting
             .joins(:setting)
--- a/app/controllers/api/v1/admin/site_settings_controller.rb
+++ b/app/controllers/api/v1/admin/site_settings_controller.rb
@@ -5,20 +5,29 @@ module Api
     module Admin
       class SiteSettingsController < ApiController
         def index
-          data = Setting.joins(:site_settings)
+          site_settings = Setting.joins(:site_settings)
                         .where(site_settings: { provider: 'greenlight' })
                         .pluck(:name, :value)
                         .to_h
 
-          render_json data:
+          return render_error status: :internal_server_error if site_settings.blank?
+          
+          render_json data: site_setting
         end
 
         def update
-          SiteSetting
-            .joins(:setting)
-            .where(provider: 'greenlight')
-            .where(setting: { name: params[:name] })
-            .update(value: params[:value].to_s)
+          return render_error status: :bad_request unless params[:siteSetting] && params[:siteSetting][:value]
+
+          site_setting = SiteSetting.joins(:setting)
+                                    .find_by(
+                                      provider: 'greenlight',
+                                      setting: { name: params[:name] }
+                                    )
+
+          return render_error status: :not_found unless site_setting
+
+          return render_error status: :bad_request unless site_setting.update(value: params[:siteSetting][:value].to_s)
+
           render_json
         end
       end
--- a/app/controllers/api/v1/admin/site_settings_controller.rb
+++ b/app/controllers/api/v1/admin/site_settings_controller.rb
@@ -12,7 +12,7 @@ module Api
 
           return render_error status: :internal_server_error if site_settings.blank?
 
-          render_json data: site_settings, status: :ok
+          render_data data: site_settings, status: :ok
         end
 
         def update
--- a/app/controllers/api/v1/admin/site_settings_controller.rb
+++ b/app/controllers/api/v1/admin/site_settings_controller.rb
@@ -10,8 +10,6 @@ module Api
                                  .pluck(:name, :value)
                                  .to_h
 
-          return render_error status: :internal_server_error if site_settings.blank?
-
           render_data data: site_settings, status: :ok
         end
 
--- a/app/controllers/api/v1/admin/site_settings_controller.rb
+++ b/app/controllers/api/v1/admin/site_settings_controller.rb
@@ -6,29 +6,26 @@ module Api
       class SiteSettingsController < ApiController
         def index
           site_settings = Setting.joins(:site_settings)
-                        .where(site_settings: { provider: 'greenlight' })
-                        .pluck(:name, :value)
-                        .to_h
+                                 .where(site_settings: { provider: 'greenlight' })
+                                 .pluck(:name, :value)
+                                 .to_h
 
           return render_error status: :internal_server_error if site_settings.blank?
-          
-          render_json data: site_setting
+
+          render_json data: site_settings, status: :ok
         end
 
         def update
-          return render_error status: :bad_request unless params[:siteSetting] && params[:siteSetting][:value]
-
           site_setting = SiteSetting.joins(:setting)
                                     .find_by(
                                       provider: 'greenlight',
                                       setting: { name: params[:name] }
                                     )
-
           return render_error status: :not_found unless site_setting
 
           return render_error status: :bad_request unless site_setting.update(value: params[:siteSetting][:value].to_s)
 
-          render_json
+          render_data status: :ok
         end
       end
     end
--- a/app/controllers/api/v1/admin/site_settings_controller.rb
+++ b/app/controllers/api/v1/admin/site_settings_controller.rb
@@ -5,8 +5,11 @@ module Api
     module Admin
       class SiteSettingsController < ApiController
         def update
-          SiteSetting.find_by(setting_id: params[:settingId]).update(value: params[:settingValue].to_s)
-
+          SiteSetting
+            .joins(:setting)
+            .where(provider: 'greenlight')
+            .where(setting: { name: params[:settingName] })
+            .update(value: params[:settingValue].to_s)
           render_json
         end
       end
--- a/app/controllers/api/v1/site_settings_controller.rb
+++ b/app/controllers/api/v1/site_settings_controller.rb
@@ -11,7 +11,7 @@ module Api
           site_setting_data[setting.name] = { value: setting.value, id: setting.id }
         end
 
-        render_json data: site_setting_data, status: :ok
+        render_json data: site_setting_data
       end
 
       # GET /api/v1/site_settings/:name
@@ -22,7 +22,7 @@ module Api
       def update
         SiteSetting.find_by(setting_id: params[:settingId]).update(value: params[:settingValue].to_s)
 
-        render_data status: :ok
+        render_json
       end
     end
   end
--- a/app/controllers/api/v1/site_settings_controller.rb
+++ b/app/controllers/api/v1/site_settings_controller.rb
@@ -3,24 +3,9 @@
 module Api
   module V1
     class SiteSettingsController < ApiController
-      def index
-        data = Setting.joins(:site_settings)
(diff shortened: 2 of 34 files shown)


app/controllers/api/v1/site_settings_controller.rb
```diff
@@ -3,10 +3,27 @@
 module Api
   module V1
     class SiteSettingsController < ApiController
+      def index
+        site_setting_data = {}
+        site_settings = SiteSetting.joins(:setting).select(:id, :name, :value)
```
KH-Amir-TN, line 8: Great idea on returning a `Hash` of `:name => { :setting_data }` that simplifies exponentially the work on the client app 👏
The use of `:id` however isn't mandatory so I'd vote for the use of settings `:name` to identify records instead since it will remove the need for having to manually iterate over the `site_settings` collection to form the `site_settings_data` `Hash` without introducing a lot of complexity to other APIs and it will also provide a much more informative access logs for deployments (having an access log for `PUT /.../site_settings/Terms.json` is much informative then `/.../site_settings/100.json` ).

The `site_setting_data` can be of one level deep only `Hash` `:name => :value` and formed by something like:
```
Setting.joins(:site_settings)
                               .where(site_settings: { provider: 'greenlight' })
                               .pluck(:name, :value)
                               .to_h
``` 
Also, let's avoid to join on child/join tables unless it's mandatory or helps to simplify our workload in a certain aspect.

app/controllers/api/v1/site_settings_controller.rb
```diff
@@ -3,10 +3,27 @@
...
+
+        site_settings.each do |setting|
+          site_setting_data[setting.name] = { value: setting.value, id: setting.id }
+        end
+
+        render_json data: site_setting_data, status: :ok
```
KH-Amir-TN, line 14: We can remove the explicit `status: :ok` for **success** renders.

app/controllers/api/v1/site_settings_controller.rb
```diff
@@ -3,10 +3,27 @@
...
+      end
+
       # GET /api/v1/site_settings/:name
       def show
         render_json data: SettingGetter.new(setting_name: params[:name], provider: 'greenlight').call # TODO: - ahmad: fix provider
       end
+
+      def update
```
KH-Amir-TN, line 22: I think that the  `#update` API should be in `Admin::SiteSettings#update`.

app/controllers/api/v1/site_settings_controller.rb
```diff
@@ -3,10 +3,27 @@
...
+        SiteSetting.find_by(setting_id: params[:settingId]).update(value: params[:settingValue].to_s)
```
KH-Amir-TN, line 23: We should handle cases of not finding the record and providing invalid params while returning adequate feedbacks.

app/controllers/api/v1/site_settings_controller.rb
```diff
@@ -3,10 +3,27 @@
...
+
+        render_data status: :ok
```
KH-Amir-TN, line 25: Since sometimes we'd be using the `render_json` then IMO we should be using `render_data` only when requiring the formatting by one of our serializers.

app/javascript/components/admin/site_settings/Settings.jsx
```diff
@@ -1,7 +1,37 @@
...
+    <div>
+      <Stack direction="horizontal">
+        <Stack>
+          <strong> Allow Users to Share Rooms </strong>
+          <p className="text-muted">
+            Setting to disbaled will remove the button from the room options <br />
+            dropdown, preventing users from sharing rooms
+          </p>
+        </Stack>
+        <div className="form-switch">
```
KH-Amir-TN, line 23: We can use `react-hook-form` here.

app/controllers/api/v1/site_settings_controller.rb
KH-Amir-TN: Also, it may be a good idea to encapsulate the `data` in a root key and leave the query params **only not** encapsulated. 
So, having an `ActionController::Parameters` instance:
`{ id: '' ,search: '', sort: { column: '', direction: '' }, site_setting: { value: '' } }`
May be better to maintain then:
`{ settingId: '' , settingValue: '' ,search: '', sort_column: '', sort_direction: '' } }`

For the record, for GL3 at the moment both are valid since we don't have a lot of `params` but it's worth knowing.

config/routes.rb
```diff
@@ -62,7 +62,8 @@
       resources :verify_account, only: :create do
         post '/activate', to: 'verify_account#activate', on: :collection
       end
-      resources :site_settings, only: :show, param: :name
+      resources :site_settings, only: %i[show index], param: :name
```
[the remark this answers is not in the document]
KH-Amir-TN: The `#update` API itself should be in a controller under the `Admin` namespace IMO, so that alone will solve your issue.
Otherwise for similar cases it's better to have a unified `param` so having the `:id` here may not be a good idea (more on that given above.).

app/javascript/components/admin/site_settings/Settings.jsx
```diff
@@ -1,7 +1,37 @@
 import React from 'react';
+import { Stack } from 'react-bootstrap';
+import useSiteSettings from '../../../hooks/queries/admin/site_settings/useSiteSettings';
+import Spinner from '../../shared/stylings/Spinner';
+import useUpdateSiteSetting from '../../../hooks/mutations/admins/site_settings/useUpdateSiteSetting';
+
+export default function Settings() {
+  const { isLoading, data: siteSettings } = useSiteSettings();
+  const updateSiteSetting = useUpdateSiteSetting();
```
KH-Amir-TN, line 9: Isn't the `useUpdateSiteSetting` mutation expects an `id` argument?
There's no need (may be better to avoid) to pass it manually in the request body the as a `SettingId`.

app/controllers/api/v1/site_settings_controller.rb
hadicheaito1: I see where you are coming from, the reason I used :id though is to cover the case where the provider is not greenlight for example. In that case, you would need to use the name, value, and provider to locate the correct site setting to update. The way I have it now, you only need the :id to update the correct site setting.

That being said, I'm not sure what we want to go with for now. @farhatahmad

app/javascript/components/admin/site_settings/Settings.jsx
hadicheaito1: Yup good catch, I knew something was weird with this :laughing:

hadicheaito1: Actually I remembered why I did this, it is because on the Settings.jsx page, there are multiple settings and different id's, so I can not just pass one id on line 9 since it is not known which id is being updated

hadicheaito1: That being said though, I should remove the id param in `useUpdateSiteSetting`

app/javascript/hooks/queries/admin/site_settings/useSiteSettings.jsx
```diff
@@ -0,0 +1,9 @@
+import { useQuery } from 'react-query';
+import axios from '../../../../helpers/Axios';
+
+export default function useSiteSettings() {
+  return useQuery(
+    'getSiteSettings',
+    async () => axios.get('/site_settings.json').then((resp) => resp.data.data),
```
KH-Amir-TN, line 7: Why having this anonymous `async` function wrapping the `axios.get`?

>Besides that, we use the `async/await` to eliminate the need for using the promise chain syntax `then` (some may refer to as the "thenable syntax").

app/controllers/api/v1/site_settings_controller.rb
```diff
@@ -3,10 +3,25 @@
...
   module V1
     class SiteSettingsController < ApiController
+      def index
+        data = Setting.joins(:site_settings)
+                      .where(site_settings: { provider: 'greenlight' })
+                      .pluck(:name, :value)
+                      .to_h
+
+        render_json data:
+      end
```
farhatahmad, line 13: I don't believe this is needed here - I would move it to the admin controller

app/controllers/api/v1/site_settings_controller.rb
```diff
@@ -3,10 +3,25 @@
...
       # GET /api/v1/site_settings/:name
       def show
         render_json data: SettingGetter.new(setting_name: params[:name], provider: 'greenlight').call # TODO: - ahmad: fix provider
       end
+
+      def update
+        SiteSetting.find_by(setting_id: params[:settingId]).update(value: params[:settingValue].to_s)
+
+        render_json
+      end
```
farhatahmad, line 24: You can remove this from here too

config/routes.rb
```diff
@@ -73,6 +73,11 @@
         end
         resources :server_rooms, only: %i[index destroy], param: :friendly_id
         resources :server_recordings, only: %i[index]
+        resources :site_settings, only: %i[] do
+          collection do 
+            patch '/update', to: 'site_settings#update'
+          end
+        end
```
farhatahmad, line 80: You can just use `edit` here instead of update

farhatahmad: sorry not `edit` but use the `update` resource instead of manually specifying the path

hadicheaito1: The reason I made it manually is because the `update` resource takes in a id as a param. In this case there is no id needed in the param

app/controllers/api/v1/admin/site_settings_controller.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module Admin
+      class SiteSettingsController < ApiController
+        def index
```
KH-Amir-TN, line 7: Do we need the `#index` here?

config/routes.rb
```diff
@@ -73,6 +73,11 @@
         end
         resources :server_rooms, only: %i[index destroy], param: :friendly_id
         resources :server_recordings, only: %i[index]
+        resources :site_settings, only: %i[index] do
+          collection do 
+            patch '/update', to: 'site_settings#update'
```
KH-Amir-TN, line 78: Why not `resources :site_settings, only: :updadte, param: :name`?

KH-Amir-TN:
The `Admin::SiteSettings#update` and `SiteSettings#index` APIs are introduced in other PRs #3663 #3664.
Instead of reinventing these APIs  I'd recommend that we review those PRs, merge and rebase our work on them.

app/controllers/api/v1/admin/site_settings_controller.rb
hadicheaito1: Yeah I was told my Ahmad since retrieving the site settings is only used in the Admin Panel right now

app/controllers/api/v1/admin/site_settings_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+          site_settings = Setting.joins(:site_settings)
+                                 .where(site_settings: { provider: 'greenlight' })
+                                 .pluck(:name, :value)
+                                 .to_h
+
+          return render_error status: :internal_server_error if site_settings.blank?
```
farhatahmad, line 13: Uhhhh why would this return an error?

app/controllers/api/v1/admin/site_settings_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+
+          render_json data: site_settings, status: :ok
```
farhatahmad, line 15: Need the new `render_data`

app/controllers/api/v1/admin/site_settings_controller.rb
KH-Amir-TN: Yup, that was mine xD
What I thought about is what to do when having an  invalid `:provider` that causes the return of an empty data `{}` and I wanted to obscure the error.

But after giving it  a second thought this shouldn't happen, so removing that should be fine.

app/serializers/site_settings_serializer.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class SiteSettingsSerializer < ApplicationSerializer
```
KH-Amir-TN, line 3: Are we using this serializer?
We're returning a `Hash` for the `Admin::SiteSettings#index` and a value on the `SiteSettings#show`.

config/routes.rb
```diff
@@ -62,7 +62,7 @@
       resources :verify_account, only: :create do
         post '/activate', to: 'verify_account#activate', on: :collection
       end
-      resources :site_settings, only: :show, param: :name
+      resources :site_settings, only: %i[show], param: :name
```
KH-Amir-TN, line 65: Is this required?

app/javascript/components/admin/site_settings/SiteSettingsRow.jsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Stack } from 'react-bootstrap';
+import useUpdateSiteSetting from '../../../hooks/mutations/admins/site_settings/useUpdateSiteSetting';
+
+export default function SiteSettingsRow(props) {
```
KH-Amir-TN, line 6:
```suggestion
export default function SiteSettingsRow({ name, title, description, value }) {
```

app/serializers/site_settings_serializer.rb
hadicheaito1: It's not being used, I added it before and forgot to remove it. Removing it now.

config/routes.rb
hadicheaito1: Will change back

farhatahmad (dismissed):

Issue #3342, opened 2022-03-31
Allow User to "Keep me logged in"


scouillard:
**Frontend:**
User, when signing in, has the option to "Be Remembered" by selecting a checkbox.
As such, User's email and password will be filled in by default.

**Backend:**
Storage of a new "Remember Token" inside Cookies?

farhatahmad:
Let's do a `Keep me logged in`

PR #3667, opened 2022-07-18
Session Cookie

scouillard:
## Description

PLEASE GO TO #3671 

EDIT: Just seen this is not what was asked, as per https://github.com/bigbluebutton/greenlight/issues/3342#issuecomment-1180550020

### DIFF
--- a/app/controllers/api/v1/sessions_controller.rb
+++ b/app/controllers/api/v1/sessions_controller.rb
@@ -19,6 +19,7 @@ module Api
 
         # TODO: Add proper error logging for non-verified token hcaptcha
         if user.present? && user.authenticate(session_params[:password])
+          remember user if session_params[:remember_me]
           sign_in user
           render_data data: current_user, serializer: CurrentUserSerializer
         else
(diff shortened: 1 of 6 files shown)


KH-Amir-TN (commented):
The approach is clean and correct.
Thinking about security and ensuring the cryptographic signing of the cookie data is a good catch.
However, I believe that we can approach this with few lines to change and without creating any tokens.

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -35,13 +36,18 @@ def destroy
...
+        params.require(:session).permit(:email, :password, :remember_me)
       end
 
       # Signs In the user
       def sign_in(user)
         session[:user_id] = user.id
       end
+
+      def remember(user)
+        cookie.permanent.signed[:user_id] = user.id
```
KH-Amir-TN, line 48: Permanent cookies are scheduled to expire in 20 years [CHECK HERE](https://apidock.com/rails/ActionDispatch/Cookies/CookieJar/permanent).
I believe that we want the session data to stay in the client browser but not for such a long period (a couple of days may be good).

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -35,13 +36,18 @@ def destroy
...
+        cookie.permanent[:remember_token] = user.generate_remember_token!
```
KH-Amir-TN, line 49: Besides that I don't think that this the simplest way of achieving this.
But even with the current solution the `remember_token` is **NOT** tamper proof and require **at least** cryptographic signature.

PR #3671, opened 2022-07-19
Add Extended Sessions

scouillard:
## Description

Add Extended Session feature.

- If the "Remember Me" checkbox is selected by the User, the `extended_session` cookie (in addition to the session cookie) will be created upon signing in.
- The `extended_session` cookie is set to expire in 2 weeks from it's creation.
- The `current_user` method will check for the `extended_session` cookie before settling for the session cookie.
- On sign out, both cookies are set to nil.

Initially, I tried to overwrite the session[:_APPNAME_session] cookie with an extra { expires: x days } field, but Rails would complain.
I figured creating an extra cookie is not too cumbersome. 

## Testing Steps

NOTE: CHROME BROWSERS HAVE THIS FEATURE BUILT-IN. PLEASE DISABLE IT OR TEST WITH MOZILLA FIREFOX INSTEAD.

**Test if the session cookie has the default behaviour - or the browser does not have the built-in feature:**
1. Sign in without checking the "Remember Me" checkbox. 
2. Close the browser
3. Navigate back to the app, you should be logged out.

**Test the extended session feature:**
1. Sign in with the "Remember Me" checkbox checked.
2. Close the browser
3. Navigate back to the app, you should be still logged in.

### DIFF
--- a/app/controllers/api/v1/sessions_controller.rb
+++ b/app/controllers/api/v1/sessions_controller.rb
@@ -28,19 +28,34 @@ module Api
 
       # DELETE /api/v1/sessions/signout
       def destroy
-        session[:user_id] = nil
+        sign_out
         render_data status: :ok
       end
 
       private
 
       def session_params
-        params.require(:session).permit(:email, :password)
+        params.require(:session).permit(:email, :password, :extend_session)
       end
 
-      # Signs In the user
       def sign_in(user)
-        session[:user_id] = user.id
+        # Creates an extended_session cookie instead of a session cookie if extend_session is selected in sign in form.
+        if session_params[:extend_session]
+          cookies.encrypted[:_extended_session] = {
+            value: {
+              user_id: user.id
+            },
+            expires: 7.days,
+            httponly: true
+          }
+        else
+          session[:user_id] = user.id
+        end
+      end
+
+      def sign_out
+        reset_session
+        cookies.delete :_extended_session
       end
     end
   end
--- a/app/controllers/api/v1/sessions_controller.rb
+++ b/app/controllers/api/v1/sessions_controller.rb
@@ -39,7 +39,7 @@ module Api
       end
 
       def sign_in(user)
-        # Creates an extended_session cookie instead of a session cookie if extend_session is selected in sign in form.
+        # Creates an extended_session cookie if extend_session is selected in sign in form.
         if session_params[:extend_session]
           cookies.encrypted[:_extended_session] = {
             value: {
@@ -48,9 +48,8 @@ module Api
             expires: 7.days,
             httponly: true
           }
-        else
-          session[:user_id] = user.id
         end
+        session[:user_id] = user.id
       end
 
       def sign_out
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,7 +4,7 @@ class ApplicationController < ActionController::Base
   # Returns the current signed in User (if any)
   def current_user
     # Overwrites the session cookie if an extended_session cookie exists
-    session[:user_id] = cookies.encrypted[:_extended_session]['user_id'] if cookies.encrypted[:_extended_session].present?
+    session[:user_id] ||= cookies.encrypted[:_extended_session]['user_id'] if cookies.encrypted[:_extended_session].present?
     @current_user ||= User.find_by(id: session[:user_id])
   end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,6 +3,8 @@
 class ApplicationController < ActionController::Base
   # Returns the current signed in User (if any)
   def current_user
+    # Overwrites the session cookie if an extended_session cookie exists
+    session[:user_id] = cookies.encrypted[:_extended_session]['user_id'] if cookies.encrypted[:_extended_session].present?
     @current_user ||= User.find_by(id: session[:user_id])
   end
 
--- a/app/javascript/components/forms/SigninForm.jsx
+++ b/app/javascript/components/forms/SigninForm.jsx
@@ -46,9 +46,13 @@ export default function SigninForm() {
       <FormControl field={fields.password} type="password" />
       <Row>
         <Col>
-          <BootstrapForm.Group className="mb-2" controlId="rememberMeCheckbox">
-            <BootstrapForm.Check type="checkbox" label="Remember me" className="small" {...methods.register('extend_session')} />
-          </BootstrapForm.Group>
+          <FormControl
+            control={BootstrapForm.Check}
+            field={fields.extend_session}
+            label={fields.extend_session.label}
+            type="checkbox"
+            noLabel
+          />
         </Col>
         <Col>
           <Link to="/forget_password" className="text-link float-end small"> Forgot password? </Link>
--- a/app/javascript/components/forms/SigninForm.jsx
+++ b/app/javascript/components/forms/SigninForm.jsx
@@ -1,3 +1,5 @@
+/* eslint-disable react/jsx-props-no-spreading */
+
 import React, { useState, useRef } from 'react';
 import {
   Button, Col, Row, Stack, Form as BootstrapForm,
@@ -44,8 +46,8 @@ export default function SigninForm() {
       <FormControl field={fields.password} type="password" />
       <Row>
         <Col>
-          <BootstrapForm.Group className="mb-2" controlId="formBasicCheckbox">
-            <BootstrapForm.Check type="checkbox" label="Remember me" className="small" />
+          <BootstrapForm.Group className="mb-2" controlId="rememberMeCheckbox">
+            <BootstrapForm.Check type="checkbox" label="Remember me" className="small" {...methods.register('extend_session')} />
           </BootstrapForm.Group>
         </Col>
         <Col>
--- a/spec/controllers/sessions_controller_spec.rb
+++ b/spec/controllers/sessions_controller_spec.rb
@@ -45,8 +45,8 @@ RSpec.describe Api::V1::SessionsController, type: :controller do
 
       delete :destroy
 
+      expect(cookies.encrypted[:_extended_session]).to be_nil
       expect(session[:user_id]).to be_nil
     end
   end
 end
-
--- a/spec/controllers/sessions_controller_spec.rb
+++ b/spec/controllers/sessions_controller_spec.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::V1::SessionsController, type: :controller do
+  before do
+    request.headers['ACCEPT'] = 'application/json'
+  end
+
+  describe '#create' do
+    it 'creates a regular session if the remember me checkbox is not selected' do
+      user = create(:user, email: 'email@email.com', password: 'password', password_confirmation: 'password')
+
+      post :create, params: {
+        session: {
+          email: 'email@email.com',
+          password: 'password',
+          extend_session: false
+        }
+      }, as: :json
+
+      expect(cookies.encrypted[:_extended_session]).to be_nil
+      expect(session[:user_id]).to eq(user.id)
+    end
+
+    it 'creates an extended session if the remember me checkbox is selected' do
+      user = create(:user, email: 'email@email.com', password: 'password', password_confirmation: 'password')
+      post :create, params: {
+        session: {
+          email: 'email@email.com',
+          password: 'password',
+          extend_session: true
+        }
+      }, as: :json
+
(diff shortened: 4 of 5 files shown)


app/controllers/api/v1/sessions_controller.rb
```diff
@@ -15,11 +15,14 @@ def create
...
 
         # TODO: Add proper error logging for non-verified token hcaptcha
-        return render_error if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
+        # return render_error if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
         # TODO: Add proper error logging for non-verified token hcaptcha
         if user.present? && user.authenticate(session_params[:password])
-          sign_in user
+          # Creates an extra cookie if extend_session is selected by the user.
+          cookies.encrypted[:_extended_session] = { value: user.id, expires: 14.days } if session_params[:extend_session]
```
KH-Amir-TN, line 23: For security purposes we'd like to mark the `_extended_session` as `httponly` to prevent scripts from accessing it.
Also, I think you could leverage that `sign_in user` to do all of this and further clean the action.

KH-Amir-TN:
```
      # Signs In the user
      def sign_in(user:, extended:)
        # Creates an extra cookie if extend_session is selected by the user.
        if extended
          cookies.encrypted[:_extended_session] = {
            value: user.id,
            expires: 14.days,
            httponly: true
          }
        end

        session[:user_id] = user.id
      end
```

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -29,18 +32,14 @@ def create
       # DELETE /api/v1/sessions/signout
       def destroy
         session[:user_id] = nil
+        cookies.encrypted[:_extended_session] = nil
```
KH-Amir-TN, line 35: Same here you can create `sign_out` private method to clean a bit the action and be consistent with the `sign_in`.
Besides, setting the `cookies.encrypted[:_extended_session] = nil` will not remove the cookie you can use the `cookies.delete(:_COOKIE_KEY)` API for that.

KH-Amir-TN: Something like:
```
      # Signs Out the user
      def sign_out
        reset_session
        cookies.delete(:_extended_session)
      end
````

File: app/controllers/application_controller.rb [hunk 3]
```diff
@@ -3,7 +3,9 @@
 class ApplicationController < ActionController::Base
   # Returns the current signed in User (if any)
   def current_user
-    @current_user ||= User.find_by(id: session[:user_id])
+    # Returns the extended session cookie value if present instead of the session cookie that expires on browser exit
+    user_id = cookies.encrypted[:_extended_session].presence || session[:user_id]
```
KH-Amir-TN, line 7:
```suggestion
    user_id = cookies.encrypted[:_extended_session] || session[:user_id]
```

KH-Amir-TN: Besides, I don't think that having two sources of truth is good it causes confusing TBH and may raise issues.

I'm a strong fallower of the separation of concerns design principle [CHECK HERE](https://en.wikipedia.org/wiki/Separation_of_concerns).

When the user enables `remember_me`, the session cookie itself isn't enough to persist its data so we need another persistent placeholder or store.

So, the `_extended_session` concern/responsibility is to hold that data and provide it to the session when it needs it and the `session` cookie is the only one that should be the source of truth for the user data while the other is a simple store.

It's the same concept in computer software, we may store data on the disk (our `_extended_session`) to persist it but when we start the software it will load it first to the RAM (our `session`) that will use it afterwards as its source of truth.

KH-Amir-TN: With being said we can do something like:
```
  # Returns the current signed in User (if any)
  def current_user
    # Returns the extended session cookie value if present instead of the session cookie that expires on browser exit
    session[:user_id] ||= cookies.encrypted[:_extended_session]
  end
```

KH-Amir-TN (changes requested):

app/javascript/components/forms/SigninForm.jsx
```diff
@@ -45,7 +45,8 @@ export default function SigninForm() {
       <Row>
         <Col>
           <BootstrapForm.Group className="mb-2" controlId="formBasicCheckbox">
-            <BootstrapForm.Check type="checkbox" label="Remember me" className="small" />
+            {/* eslint-disable-next-line react/jsx-props-no-spreading */}
+            <BootstrapForm.Check type="checkbox" label="Remember me" className="small" {...methods.register('extend_session')} />
```
hadicheaito1, line 49: I'm curious how the `{...methods.register('extend_session')}` works?

scouillard: It's from React Hook Form https://react-hook-form.com/api/useform/register

It's basically the thing that let's you build the object that is being sent in the request.

Ex:
register("firstName") --> {firstName: 'value'}

app/javascript/components/forms/SigninForm.jsx
```diff
@@ -45,7 +45,8 @@ export default function SigninForm() {
       <Row>
         <Col>
           <BootstrapForm.Group className="mb-2" controlId="formBasicCheckbox">
-            <BootstrapForm.Check type="checkbox" label="Remember me" className="small" />
+            {/* eslint-disable-next-line react/jsx-props-no-spreading */}
```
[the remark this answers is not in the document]
scouillard: It's to disable the JSX - it doesn't approve the ...register but it's needed with React Hook Form.

app/controllers/application_controller.rb
scouillard: That's a very good use of the ||= operator. Thank you. I'm on it.

app/javascript/components/forms/SigninForm.jsx
KH-Amir-TN: > It's from React Hook Form https://react-hook-form.com/api/useform/register
> 
> It's basically the method that builds the object that is being sent in the request.
> 
> Ex: register("firstName") --> {firstName: 'value'} register("name.firstName") --> {name: { firstName: 'value' }}

Exactly!
Good explanation @scouillard 

It achieves that by returning a prop object defined according to `react-hook-form` config made when invoking its hook `useForm`.
The register props registers any kind of HTML inputs or select elements to `react-hook-form` so that it can control and keep hold of it.
It returns the fallowing props -> `name:, ref:, onChange, onBlur`.

Internally the registration process is a mapping of the registration name to a ref object.
More on that [HERE](https://react-hook-form.com/api/useform/register).

app/javascript/components/forms/SigninForm.jsx
```diff
@@ -45,7 +45,8 @@ export default function SigninForm() {
       <Row>
         <Col>
           <BootstrapForm.Group className="mb-2" controlId="formBasicCheckbox">
```
KH-Amir-TN, line 47: Can't we use the `FormControl` component?

app/javascript/components/forms/SigninForm.jsx
hadicheaito1: Thanks for the explanation guys :+1:

app/controllers/application_controller.rb
KH-Amir-TN: `cookies.encrypted[:_greenlight_session]` is managed by Rails itself so it makes since that it will overwrite your changes.

app/javascript/components/forms/SigninForm.jsx
KH-Amir-TN: >I'd rather not use the Form template we have because I don't feel in control when doing so.
Oh, is there a problem with the Form framework?

KH-Amir-TN:
```
          <FormControl
            control={BootstrapForm.Check}
            field={fields.extend_session}
            label={fields.extend_session.label}
            type="checkbox"
            noLabel
          />
```
```
export const signinFormFields = {
  email: { ... },
  password: { ... },
  extend_session: {
    label: 'Remember me',
    controlId: 'signinFormRememberMe',
    hookForm: {
      id: 'extend_session',
    },
  },
};
```

```
const validationSchema = yup.object({
  ...,
  extend_session: yup.bool(),
});

export const signinFormConfig = {
  defaultValues: {
    email: '',
    password: '',
    extend_session: false,
  },
  resolver: yupResolver(validationSchema),
};
```

KH-Amir-TN: Avoiding the use of the consistent form template can have its drawbacks.

Changing how forms and form inputs behave or appear can become somehow impossible to manage if we're not having well known templates/frameworks in use. 

React hook form as a lib offers flexibility that is good to a certain extent but flexibility without consistency becomes a problem therefore the form framework is in place to first simplify the complexity of `react-hook-form` (any lib that offers a lot of flexibility has more learning curve), ensure the correct use of the lib and offer a consistent way for working with forms and displaying them.

hadicheaito1 (dismissed):

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -28,19 +28,34 @@ def create
...
+        if session_params[:extend_session]
+          cookies.encrypted[:_extended_session] = {
+            value: {
+              user_id: user.id
+            },
+            expires: 7.days,
+            httponly: true
+          }
+        else
+          session[:user_id] = user.id
```
KH-Amir-TN, line 52: You can keep using the keep setting the ` session[:user_id] = user.id` always.

app/controllers/application_controller.rb
```diff
@@ -3,6 +3,8 @@
 class ApplicationController < ActionController::Base
   # Returns the current signed in User (if any)
   def current_user
+    # Overwrites the session cookie if an extended_session cookie exists
+    session[:user_id] = cookies.encrypted[:_extended_session]['user_id'] if cookies.encrypted[:_extended_session].present?
```
KH-Amir-TN, line 7: You can do this only once when the session is empty and the `extended_session` isn't.

app/javascript/components/forms/SigninForm.jsx
scouillard: IMO the template adds much more complexity to React Hook Form.

In here, for example, we will be adding multiples lines of codes, in 2 extra files, to add a checkbox in our form.

When using the template, the problem comes if something is not working the way I expect it to work.
I find it really hard to debug, expand, or dealing with the Form template.
I'm not saying the problem is the template itself, it's my experience with it.

You are absolutely right that the template should be used in this example for homogeneity, as it's already been used for the Sign In Form.

KH-Amir-TN (commented):
It's better to keep the normal session flow intact and just add a small layer on top of it.

So, session used to be set on login and reset on sign-out.
After our changes (unless it's mandatory) we should try to keep the same flow where we set the session data to the session cookie, we read it from it only and on sign-out we reset it.

The only exception is when the session data is empty then we'd check our **reserve** `_extended_session` and populate the session and the flow continues as expected.  

On login we have an extra step that creates us a **reserve** that we can leverage when needed but that should be an extra step **irrelevant** (does not affect) to the session flow in any ways it just help it carrying on in certain cases that normally it would not.

app/javascript/components/forms/SigninForm.jsx
scouillard: I'll give the template another try.

spec/controllers/sessions_controller_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  describe '#destroy' do
+    it 'signs off the user' do
+      user = create(:user)
+      session[:user_id] = user.id
+
+      delete :destroy
+
+      expect(session[:user_id]).to be_nil
```
KH-Amir-TN, line 48: We'd need to assert that the **reserve** cookie is deleted.

app/controllers/application_controller.rb
```diff
@@ -3,6 +3,8 @@
 class ApplicationController < ActionController::Base
   # Returns the current signed in User (if any)
   def current_user
+    # Overwrites the session cookie if an extended_session cookie exists
+    session[:user_id] ||= cookies.encrypted[:_extended_session]['user_id'] if cookies.encrypted[:_extended_session].present?
```
KH-Amir-TN, line 7: The details 👏 👏 
I wouldn't spot this little optimization myself, good catch!

app/javascript/components/forms/SigninForm.jsx
KH-Amir-TN: Will discuss further see what's the problem with the framework on our next meeting,
If it's causing problems we should fix it or remove it and find something else.
If it's raising complexity then differently  it has a problem that we need to address.

KH-Amir-TN (approved):
# 🔥 🔥 🔥

PR #3672, opened 2022-07-20
More render_data stuff

farhatahmad:

### DIFF
--- a/app/controllers/api/v1/api_controller.rb
+++ b/app/controllers/api/v1/api_controller.rb
@@ -36,6 +36,8 @@ module Api
       end
 
       def render_data(status:, serializer: nil, data: {}, options: {})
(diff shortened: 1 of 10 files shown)


app/controllers/api/v1/api_controller.rb
```diff
@@ -36,6 +36,8 @@ def log_exception(exception)
       end
 
       def render_data(status:, serializer: nil, data: {}, options: {})
+        data = { data: } if data.is_a?(Hash) # Manually add root if not serializing
```
farhatahmad, line 39: I have to manually add the `data` at the root if the object is a hash since ActiveModelSerializer will only add it if it can find a matching serializer

Before this, any hash that wasnt serialized was being returned as:
```
{
  key: value,
  key: value
}
```

After this change: 
```
{
  data: {
    key: value,
    key: value
  }
}
```

PR #3682, opened 2022-07-21
Site Settings: Integrated `RoleMapping` and `Administration` settings with the APIs.

KH-Amir-TN:
## Description

This PR integrates Integrated `RoleMapping` and `Administration` settings with the site settings APIs.
## Testing Steps

> 1. Pull the code.
> 2. Install the dependencies `bundle install && npm|yarn install`.
> 3. Clean the previous assets build by running `rm app/assets/builds/*` (This won't remove .keep since it's hidden).
> 4. Clean the database and tmp files for a better isolation by running `rails tmp:clear && rails db:schema:cache:clear && rails db:drop && rails db:create && rails db:migrate:with_data`
> 5. Run the linter and specs `bundle exec rubocop --parallel && bundle exec rspec && npx eslint app/javascript/* --ext .jsx,.js`
> 6. Run `./bin/dev` to run the assets builders processes and the Puma server all at once.
## Screenshots (if appropriate):

image
image

PR #3683, opened 2022-07-21
Role mapping 2: Extended `UsersController#create` API and Added `UserCreator` svc.

KH-Amir-TN:
## Description

Enable users with the right set of privileges to edit roles mappers.

This PR completes **2.** 
--
~~0. Add Registration UI.~~
~~1. Integrate with the `Admin::SiteSettings` APIs.~~
~~2.Extend `UsersController#create` to infer the user role.~~
### User story [Role Mapping]:
---
0. A user with the right set of permissions signs in.
1. User navigates to the Manage site settings page in administration panel.
2. User selects the Registration tab.
3. User can edit the list of roles and their corresponding email suffixes.
5. User can update the list and have adequate feedbacks from the client app.
Note: There's no checks for the list content, it's a string comma speparated list
that can be empty or filled.

## Testing Steps

> 1. Pull the code.

image

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -18,10 +18,7 @@ module Api
         # TODO: amir - ensure accessibility for unauthenticated requests only.
         params[:user][:language] = I18n.default_locale if params[:user][:language].blank?
 
-        user = User.new({
-          provider: 'greenlight',
-          role: Role.find_by(name: 'User') # TODO: - Ahmad: Move to service
-        }.merge(user_params)) # TMP fix for presence validation of :provider
+        user = UserCreator.new(user_params:, provider: 'greenlight').call
 
         # TODO: Add proper error logging for non-verified token hcaptcha
         return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
--- a/spec/requests/users_controller_spec.rb
+++ b/spec/requests/users_controller_spec.rb
@@ -70,5 +70,29 @@ RSpec.describe Api::V1::UsersController, type: :request do
         expect(JSON.parse(response.body)['errors']).to be_present
       end
     end
+
+    context 'Role mapping' do
+      before do
+        setting = create(:setting, name: 'RoleMapping')
+        create(:site_setting, setting:, provider: 'greenlight', value: 'Decepticons=@decepticons.cybertron,Autobots=autobots.cybertron')
+      end
+
+      it 'Creates a User and assign a role if a rule matches their email' do
+        autobots = create(:role, name: 'Autobots')
+        user_params = {
+          name: 'Optimus Prime',
+          email: 'optimus@autobots.cybertron',
+          password: 'Autobots',
+          password_confirmation: 'Autobots',
+          language: 'teletraan'
+        }
+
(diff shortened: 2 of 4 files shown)


KH-Amir-TN:
# DEPENDS ON #3682.

app/controllers/api/v1/users_controller.rb
```diff
@@ -18,10 +18,7 @@ def create
         # TODO: amir - ensure accessibility for unauthenticated requests only.
         params[:user][:language] = I18n.default_locale if params[:user][:language].blank?
 
-        user = User.new({
-          provider: 'greenlight',
-          role: Role.find_by(name: 'User') # TODO: - Ahmad: Move to service
-        }.merge(user_params)) # TMP fix for presence validation of :provider
+        user = RoleSetter.new(user_params:, provider: 'greenlight').call
```
farhatahmad, line 21: This isn't actually setting the role - this is creating the user

farhatahmad: I would rename this to `UserCreator`

app/services/role_setter.rb
```diff
@@ -0,0 +1,26 @@
...
+  def call
+    email = @user_params[:email]
+
+    matched_rule = if @roles_mappers
+                     rules = @roles_mappers.split(',').map { |rule| rule.split('=') }
+
+                     rules.find { |rule| email.ends_with? rule.second if rule.second }
+                   end
+
+    role = Role.find_by(name: matched_rule&.first) || Role.find_by(name: 'User')
```
farhatahmad, line 19: I would move this to a private method inside the service (after making the changes above)

spec/requests/users_controller_spec.rb
```diff
@@ -70,5 +70,33 @@
         expect(JSON.parse(response.body)['errors']).to be_present
       end
     end
+
+    context 'Role mapping' do
+      let(:user_creator) { instance_double(UserCreator) }
```
KH-Amir-TN, line 75: I'm a bit skeptical about this and believe it may introduce complexity for you guys.
The reason behind that was me not wanting to call the `:new` original method because it will require a lot of mocking for the `SettingGetter` service.
So, I wanted a way to mock the `UserCreator` instance without doing any of that.
That appears to be the recommended way of doing it according to [RuboCop](https://www.rubydoc.info/gems/rubocop-rspec/1.7.0/RuboCop/Cop/RSpec/AnyInstance).

KH-Amir-TN: Another problem was trying to assert the params given to the `UserCreator.new`.
The `user_params` is coming from the `user_controller#user_params` function that returns an instance of` `ActionController::Parameters` after requiring and permitting its values.

To avoid having to do some complex logic inside of the test `expect(user_creator).to receive(user_params: ActionController::Parameters.new(user_params).permit(...))` I decided to omit that assertion in favor of asserting the expected outcome by checking the `email` value to match what was given to the API.

farhatahmad: I think we could simplify this down to solve our problem. I would just test that the service is called properly and that's it. You are already testing the service in it's own file. 

Just expect that the params and provider are passed through

KH-Amir-TN: There's no clean way that I can think of to make this as a unit test.
So, I'm changing it into a functional one.

farhatahmad (approved):

PR #3691, opened 2022-07-25
Back-end: Pagination

hadicheaito1:
## Description

- Installing pagy gem for back-end pagination
- Added `pagy.rb` initialization file
- Adding pagination code to all back-end API endpoints that need it


config/initializers/pagy.rb
```diff
@@ -0,0 +1,235 @@
...
+# Pagy::DEFAULT[:searchkick_search] = :search
+# require 'pagy/extras/searchkick'
+# uncomment if you are going to use Searchkick.pagy_search
+# Searchkick.extend Pagy::Searchkick
+
+# Frontend Extras
+
+# Bootstrap extra: Add nav, nav_js and combo_nav_js helpers and templates for Bootstrap pagination
+# See https://ddnexus.github.io/pagy/extras/bootstrap
+require 'pagy/extras/bootstrap'
```
KH-Amir-TN, line 119: I believe we don't need this.

Gemfile.lock
```diff
@@ -389,3 +392,9 @@ DEPENDENCIES
   tzinfo-data
   web-console
   webdrivers
+
+RUBY VERSION
+   ruby 3.1.0p0
+
+BUNDLED WITH
+   2.3.10
```
farhatahmad, line 400: Remove these

farhatahmad (approved):

config/initializers/pagy.rb
```diff
@@ -0,0 +1,235 @@
...
+# See https://ddnexus.github.io/pagy/extras/items
+# require 'pagy/extras/items'
+# set to false only if you want to make :items_extra an opt-in variable
+# Pagy::DEFAULT[:items_extra] = false    # default true
+# Pagy::DEFAULT[:items_param] = :items   # default
+# Pagy::DEFAULT[:max_items]   = 100      # default
+
+# Overflow extra: Allow for easy handling of overflowing pages
+# See https://ddnexus.github.io/pagy/extras/overflow
+# require 'pagy/extras/overflow'
```
KH-Amir-TN, line 170: It may be a good idea to set this to:
```
require 'pagy/extras/overflow'
Pagy::DEFAULT[:overflow] = :last_page
```
That will not raise an exception when trying to access an out of bound page number and rather return the last page instead.

config/initializers/pagy.rb
```diff
@@ -0,0 +1,235 @@
...
+# Pagy::DEFAULT[:meilisearch_search] = :ms_search
+# require 'pagy/extras/meilisearch'
+
+# Metadata extra: Provides the pagination metadata to Javascript frameworks like Vue.js, react.js, etc.
+# See https://ddnexus.github.io/pagy/extras/metadata
+# you must require the shared internal extra (BEFORE the metadata extra) ONLY if you need also the :sequels
+# require 'pagy/extras/shared'
+require 'pagy/extras/metadata'
+# For performance reasons, you should explicitly set ONLY the metadata you use in the frontend
+# Pagy::DEFAULT[:metadata] = %i[scaffold_url page prev next last]   # example
```
KH-Amir-TN, line 102: It may be a good idea to define **only** what we need from this list `[ :scaffold_url, :first_url, :prev_url, :page_url, :next_url, :last_url, :count, :page, :items, :vars, :pages, :last, :from, :to, :prev, :next, :series ]`.

It's even suggested in pagy official docs [CHECK HERE](https://ddnexus.github.io/pagy/extras/metadata.html#variables).

config/initializers/pagy.rb
hadicheaito1: This is needed to use `pagy_metadata()` function

config/initializers/pagy.rb
hadicheaito1: Yes I'm aware, I will do this after I find exactly what is needed and what is not (while working on the front-end)

config/initializers/pagy.rb
hadicheaito1: Good to know, I'd like to add something like this (if needed) as part of the next PR when I actually create the front-end

farhatahmad (approved):

PR #3692, opened 2022-07-25
Brand Theming: Primary/Brand color

scouillard:
## Description

Remove all usage of "primary" to style the elements with the brand color.
Create new "brand" classes to use instead.
ie: text-primary -> text-brand, variant="primary" -> variant="brand"

## Testing Steps

1. Pick a new color in the color picker, and press on submit.
2. The brand color will propagate in the whole app.

image

### DIFF
--- a/app/javascript/App.jsx
+++ b/app/javascript/App.jsx
@@ -1,14 +1,21 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { Container } from 'react-bootstrap';
 import { Outlet } from 'react-router-dom';
 import { Toaster } from 'react-hot-toast';
 import Header from './components/shared/Header';
 import { useAuth } from './contexts/auth/AuthProvider';
 import Footer from './components/shared/Footer';
+import useSiteSettings from './hooks/queries/admin/site_settings/useSiteSettings';
 
 export default function App() {
   const currentUser = useAuth();
   const containerHeight = currentUser?.signed_in ? 'full-height' : 'h-100';
+  const { data: siteSettings } = useSiteSettings();
+
+  useEffect(() => {
+    document.documentElement.style.setProperty('--brand-color', siteSettings?.PrimaryColor);
+  }, [siteSettings]);
+
   return (
     <>
       {currentUser?.signed_in && <Header /> }
--- a/app/javascript/App.jsx
+++ b/app/javascript/App.jsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React from 'react';
 import { Container } from 'react-bootstrap';
 import { Outlet } from 'react-router-dom';
 import { Toaster } from 'react-hot-toast';
@@ -12,9 +12,7 @@ export default function App() {
   const containerHeight = currentUser?.signed_in ? 'full-height' : 'h-100';
   const { data: siteSettings } = useSiteSettings();
 
-  useEffect(() => {
-    document.documentElement.style.setProperty('--brand-color', siteSettings?.PrimaryColor);
-  }, [siteSettings]);
+  document.documentElement.style.setProperty('--brand-color', siteSettings?.PrimaryColor);
 
   return (
     <>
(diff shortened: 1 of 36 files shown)


app/javascript/components/admin/site_settings/Appearance.jsx
```diff
@@ -1,7 +1,36 @@
...
 export default function Appearance() {
+  const { register, handleSubmit } = useForm();
+  const updateSiteSetting = useUpdateSiteSetting('PrimaryColor');
+  const { data: siteSettings } = useSiteSettings();
+
   return (
-    <p>Appearance</p>
+    <>
+      <h6>Primary Color</h6>
+      <Form onSubmit={handleSubmit(updateSiteSetting.mutate)}>
```
KH-Amir-TN, line 19: I'd go with making a `PrimaryColorForm` component that contains the form fields and expects the `value` to be provided from its parent.

The `Appearance` component will only be in charge of extracting the required settings from the API and provide it to each form and create their layout.

For the layout I'd go with bootstrap's Grid system `<Row></Row> <Col></Col>` unless its necessary to do otherwise.

And I'd use the Form framework when making the forms.

app/javascript/App.jsx
```diff
@@ -5,10 +5,15 @@ import { Toaster } from 'react-hot-toast';
 import Header from './components/shared/Header';
 import { useAuth } from './contexts/auth/AuthProvider';
 import Footer from './components/shared/Footer';
+import useSiteSettings from './hooks/queries/admin/site_settings/useSiteSettings';
 
 export default function App() {
   const currentUser = useAuth();
   const containerHeight = currentUser?.signed_in ? 'full-height' : 'h-100';
+  const { data: siteSettings } = useSiteSettings();
```
farhatahmad, line 13: You should use `useSiteSetting` here and specify the name of the setting that you want - not `useSiteSettings`

app/javascript/components/admin/site_settings/Appearance.jsx
scouillard: I didn't put much time on frontend/UI - this will be for next PR. 

This is just a prototype to show how the theming would work and if you guys agree with the way it's done.

I'm sorry I should have specified, I will keep your comments in mind.

KH-Amir-TN: Oh, makes sense.

app/javascript/App.jsx
scouillard: We will need to use PrimaryColorLight and PrimaryColorDark here as well - I'm not sure if that is what you mean?

hadicheaito1: `useSiteSetting('PrimaryColor')`

farhatahmad: There already is one in the SiteSettings controller

farhatahmad: Maybe we can extend the controller to potentially take an array?

hadicheaito1: The existing one only works for settings with true or false I believe ? He would need to change how it returns for it to work for all settings

scouillard: I have been using the Admin/SiteSettingsController - not the "regular" SiteSettingsController. I need to double check those.

app/javascript/hooks/queries/site_settings/useSiteSettingAsync.jsx
```diff
@@ -0,0 +1,12 @@
+import { useQuery } from 'react-query';
+import axios from '../../../helpers/Axios';
+
+export default function useSiteSetting(name) {
```
farhatahmad, line 4: Can we name this `useSiteSettingAsync` to make sure its clear?

scouillard: Good catch, thank you.

farhatahmad (approved):

PR #3693, opened 2022-07-25
Brand Theming: basic UI + endpoints hookup

scouillard:
## Description

User can now set a Brand/Primary color.
Basic form using Bootstrap color picker for now.

This PR just shows off the mechanic to propagate the Brand/Primary color value to the application.

Please use #3692 instead to test the feature.

The line below basically does all the work for the theming.
`document.documentElement.style.setProperty('--brand-color', siteSettings?.PrimaryColor);`

It adds styling to the root of the app, at `<html> ... </html>`
I'm not sure how it works - I assume the whole DOM gets re-loaded if the value of --brand-color changes?
Some say it goes against the principle of React to act on the DOM itself, some says it makes sense. 
So far, it's working pretty good IMO.
The other solution would be to add inline styling everywhere the brand color is needed.

## Testing Steps

For the full experience, you should test with #3692.

1. Choose a new color and click on submit.
2. Go to http://localhost:3000/api/v1/admin/site_settings and make sure the PrimaryColor value has changed accordingly.

### DIFF
--- a/app/javascript/components/admin/site_settings/Appearance.jsx
+++ b/app/javascript/components/admin/site_settings/Appearance.jsx
@@ -1,7 +1,36 @@
+/* eslint-disable react/jsx-props-no-spreading */
+
 import React from 'react';
+import { Form } from 'react-bootstrap';
+import Button from 'react-bootstrap/Button';
+import { useForm } from 'react-hook-form';
+import useUpdateSiteSetting
+  from '../../../hooks/mutations/admins/site_settings/useUpdateSiteSetting';
+import useSiteSettings from '../../../hooks/queries/admin/site_settings/useSiteSettings';
 
 export default function Appearance() {
+  const { register, handleSubmit } = useForm();
+  const updateSiteSetting = useUpdateSiteSetting('PrimaryColor');
+  const { data: siteSettings } = useSiteSettings();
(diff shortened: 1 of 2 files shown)


app/javascript/components/admin/site_settings/Appearance.jsx
```diff
@@ -1,7 +1,36 @@
...
+import Button from 'react-bootstrap/Button';
+import { useForm } from 'react-hook-form';
+import useUpdateSiteSetting
+  from '../../../hooks/mutations/admins/site_settings/useUpdateSiteSetting';
+import useSiteSettings from '../../../hooks/queries/admin/site_settings/useSiteSettings';
 
 export default function Appearance() {
+  const { register, handleSubmit } = useForm();
+  const updateSiteSetting = useUpdateSiteSetting('PrimaryColor');
+  const { data: siteSettings } = useSiteSettings();
```
hadicheaito1, line 14: I'm seeing two different approaches with this. Sometimes we just do 
`const { siteSettings } = useSiteSettings();` and then do `siteSettings.data.PrimaryColor' for example.

What should be the way that we stick to ?
@scouillard @farhatahmad @KH-Amir-TN

PR #3699, opened 2022-07-27
Room Settings 1: Added `RoomSettingsGetter` service.

KH-Amir-TN:
## Description

Synchronize room settings with rooms configurations.

This PR completes **1.** 
--
~~1. Create `RoomSettingsGetter` service that can take a room and return its settings as a `Hash: :name => :value` making sense of each setting and its config.~~
2. Integrate the service in `RoomSettingsController#show`.
3. Extend the `RoomSettings` and `RoomSettingsRow` UIs.

### User story [Room Settings show]:
---
1. A user authenticates.
2. A user creates and accesses a room and its settings.

[If the setting config is "true" for the room setting]:
4. User will have the setting enabled and user cannot disable it.
>NOTE: The API will set the setting value as if it was enabled, the front end only will disable the setting for editing but will not change its value.

[If the setting config is "false" for the room setting]:
3. User will not see the setting.
>NOTE: The frontend will not mount the setting component to the DOM.

[If the setting config is "optional" for the room setting]:
3. User will have their setting with its registered status 'enabled|disbaled'.
5. User can alter the status of the setting and have proper feedbacks.

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/services/room_settings_getter.rb
+++ b/app/services/room_settings_getter.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+class RoomSettingsGetter
+  # Special options are meeting options that have different values other than `true|false` to represent postives and negatives.
+  # The `special_options` hash is a registry that keeps hold of all of the special options used in GL3 along with their postive and negative values.
+  # When adding a new special option just register it in this Hash respecting the fallowing format:
+  # Hash(`<option_name> => {'true' => <Postive>, 'false' => <Negative>})`
+  SPECIAL_OPTIONS = { 'guestPolicy' => { 'true' => 'ASK_MODERATOR', 'false' => 'ALWAYS_ACCEPT' } }.freeze
+
+  def initialize(room_id:, provider:, only_bbb_options: false)
+    @room_id = room_id
+    @only_bbb_options = only_bbb_options
+    # Fetching only rooms configs that are not optional to overwrite the settings values.
+    @rooms_configs = MeetingOption.joins(:rooms_configurations)
+                                  .where(rooms_configurations: { provider: })
+                                  .where.not(rooms_configurations: { value: 'optional' })
+                                  .pluck(:name, :value)
+                                  .to_h
+  end
+
+  def call
+    room_settings = MeetingOption.joins(:room_meeting_options).where(room_meeting_options: { room_id: @room_id })
+    room_settings = room_settings.where.not('name ILIKE :prefix', prefix: 'gl%') if @only_bbb_options
+    room_settings = room_settings.pluck(:name, :value).to_h
+
+    room_settings.merge!(@rooms_configs) # Merging rooms settings with its none optional configurations prioritizing forced configs over settings.
+    room_settings_special_options = SPECIAL_OPTIONS.slice(*room_settings.keys) # Extracting the room settings special options to minimize iterations.
+
+    # Configs are enum of ['true', 'false', 'optional'], after merging none optional settings with their configs only the room
+    # special options have to map positive/negative values.
+    room_settings_special_options.each_key do |name|
+      config = room_settings[name]
+      room_settings[name] = SPECIAL_OPTIONS[name][config] if %w[true false].include? config # Config values are expected to be 'true'|'false'
+    end
+
+    room_settings
+  end
+end
--- a/spec/services/room_settings_getter_spec.rb
+++ b/spec/services/room_settings_getter_spec.rb
@@ -1,0 +1,85 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe RoomSettingsGetter, type: :service do
+  describe '#call' do
+    context 'Normal room settings' do
+      it 'returns a Hash("name" => "value") of room settings according to the configurations' do
+        room = create(:room)
+        setting1 = create(:meeting_option, name: 'glSetting1')
+        setting2 = create(:meeting_option, name: 'glSetting2')
+        setting3 = create(:meeting_option, name: 'setting3')
+
+        create(:room_meeting_option, room:, meeting_option: setting1, value: 'value1')
+        create(:room_meeting_option, room:, meeting_option: setting2, value: 'value2')
+        create(:room_meeting_option, room:, meeting_option: setting3, value: 'value3')
+
+        create(:rooms_configuration, meeting_option: setting1, provider: 'greenlight', value: 'true')
+        create(:rooms_configuration, meeting_option: setting2, provider: 'greenlight', value: 'optional')
+        create(:rooms_configuration, meeting_option: setting3, provider: 'greenlight', value: 'false')
+
+        res = described_class.new(room_id: room.id, provider: 'greenlight').call
+
+        expect(res).to eq({
+                            'glSetting1' => 'true',
+                            'glSetting2' => 'value2',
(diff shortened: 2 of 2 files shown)


app/services/room_settings_getter.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+class RoomSettingsGetter
+  # Special options are meeting options that have different values other than `true|false` to represent postives and negatives.
+  # The `special_options` hash is a registry that keeps hold of all of the special options used in GL3 along with their postive and negative values.
+  # When adding a new special option just register it in this Hash respecting the fallowing format:
+  # Hash(`<option_name> => {'true' => <Postive>, 'false' => <Negative>})`
+  SPECIAL_OPTIONS = { 'guestPolicy' => { 'true' => 'ASK_MODERATOR', 'false' => 'ALWAYS_ACCEPT' } }.freeze
+
+  def initialize(room_id:, provider:, prefix: '')
```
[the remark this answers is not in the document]
KH-Amir-TN: Fair enough, will change Inshallah

app/services/room_settings_getter.rb
```diff
@@ -0,0 +1,43 @@
...
+    @room_id = room_id
+    @only_bbb_options = only_bbb_options
+    @rooms_configs = MeetingOption.joins(:rooms_configurations)
+                                  .where(rooms_configurations: { provider: })
+                                  .pluck(:name, :value)
+                                  .to_h
+  end
+
+  def call
+    room_settings = MeetingOption.joins(:room_meeting_options).where(room_meeting_options: { room_id: @room_id })
```
KH-Amir-TN, line 20: This is a tiny optimization that simplifies the logic by removing the need to duplicate two complete queries per case or to use multiple variables for each query relation.

app/services/room_settings_getter.rb
```diff
@@ -0,0 +1,43 @@
...
+    room_settings = room_settings.where.not('name ILIKE :prefix', prefix: 'gl%') if @only_bbb_options
+    room_settings = room_settings.pluck(:name, :value).to_h
+
```
KH-Amir-TN, line 23: This is a tiny optimization that simplifies the logic by removing the need to duplicate two complete queries per case or to use multiple variables for each query relation.

KH-Amir-TN: Inspired from [THIS ARTICLE](https://www.justinweiss.com/articles/search-and-filter-rails-models-without-bloating-your-controller/).

KH-Amir-TN:
# REBASED ON 1209f83f7530f5d3d46eaaa3221c3effbaab0e24

scouillard:
This will overwrite the all the Rooms RoomSettings in case the Admin edits the Room Configuration, is this correct?

app/services/room_settings_getter.rb
scouillard: Great article

scouillard (dismissed):

KH-Amir-TN:
> The scope of those PRs is to overwrite the all the Rooms RoomSettings in case the Admin edits the Room Configuration, is this correct?

Yup, the service will be the only source of truth to fetch room settings.
Which in turn will take into consideration the rooms configurations and the setting itself, so a force enabled setting will have a positive value while a force disabled will have a negative value other despite what the setting has in the database.
When the setting is optional it will have its registered value.

app/services/room_settings_getter.rb
```diff
@@ -0,0 +1,43 @@
...
+  # Hash(`<option_name> => {'true' => <Postive>, 'false' => <Negative>})`
+  SPECIAL_OPTIONS = { 'guestPolicy' => { 'true' => 'ASK_MODERATOR', 'false' => 'ALWAYS_ACCEPT' } }.freeze
+
+  def initialize(room_id:, provider:, only_bbb_options: false)
+    @room_id = room_id
+    @only_bbb_options = only_bbb_options
+    @rooms_configs = MeetingOption.joins(:rooms_configurations)
+                                  .where(rooms_configurations: { provider: })
+                                  .pluck(:name, :value)
+                                  .to_h
```
farhatahmad, line 16: Might be worth only loading room configs that are not optional

app/services/room_settings_getter.rb
```diff
@@ -0,0 +1,43 @@
...
+    if %w[true false].include?(config)
+      forced_value = if SPECIAL_OPTIONS.key? name
+                       SPECIAL_OPTIONS[name][config]
+                     else
+                       config
+                     end
+      return [name, forced_value]
+    end
+
+    [name, value]
```
farhatahmad, line 41: I feel like this could be simplified quite a bit. Maybe merge the `room_configs` wit h`room_settings` before calling this function and then make this function only handle the `SPECIAL_OPTIONS`?

KH-Amir-TN: Happy to read this, of course we can 😆 
Tried to make it as simple and straightforward as I could,

app/services/room_settings_getter.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    room_settings.merge!(@rooms_configs) # Merging rooms settings with its none optional configurations prioritizing forced configs over settings.
+    room_settings_special_options = SPECIAL_OPTIONS.slice(*room_settings.keys) # Extracting the room settings special options to minimize iterations.
+
+    # Configs are enum of ['true', 'false', 'optional'], after merging none optional settings with their configs only the room
+    # special options have to map positive/negative values.
+    room_settings_special_options.each_key do |name|
+      config = room_settings[name]
+      room_settings[name] = SPECIAL_OPTIONS[name][config] if %w[true false].include? config # Config values are expected to be 'true'|'false'
+    end
```
farhatahmad, line 34: This is still super confusing to me - I'm having a hard time understanding the complexity behind this. What's the goal of this? To replace `ASK_MODERATOR` with `true`?

KH-Amir-TN: That's the result of our little optimization.

The service will:
1. Fetch **only** the rooms configs  that are marked  as **not** optional (`true|false`) as a `Hash(<setting_name>=><'true'|'false'>)`.
2. Fetch **all** of the the room settings as a `Hash(<setting_name> => <value>)`.
3. Merge the room settings hash with the rooms configs.
This step will overwrite all none optional configured settings with their forced value `true|false` (including but not limited to `ASK_MODERATOR`).
4. Filter only the room special settings as a `Hash(<setting_name> => Hash('true'=><POSITIVE>, 'false' => <NEGATIVE>))`.
5. Since only the room special settings aren't ready by nature and require mapping of the forced values into their expected output, the room special settings hash will be mapped where each setting (including but not limited to `ASK_MODERATOR`) existing on the room settings and the  `SPECIAL_OPTIONS` will map the setting to its **positive/negative** value.

So, when we merge the rooms settings with the rooms configs we will overwrite their values.
That optimizes our work because only a handful of  settings that we call "special" are only those which requires an extra step to set properly.
All optional settings (even the special ones) will conserve their values, while regular none optional settings will have their forced binary value and none optional special settings will map will map their values to its final output as registered in the `SPECIAL_OPTIONS` constant and according to its small specification.

Hope that clarifies the changes,

farhatahmad (approved):

PR #3702, opened 2022-07-27
Pagination - Non-Admin pages

hadicheaito1:
## Description

- Pagination front-end for all the pages that do not require admin rights

image

### DIFF
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -8,6 +8,7 @@
 @import 'shared_accesses';
 @import 'presentation';
 @import 'admin_panel';
+@import '//cdnjs.cloudflare.com/ajax/libs/semantic-ui/2.2.2/semantic.min.css';
 
 html,
 body {
--- a/app/javascript/components/recordings/Recordings.jsx
+++ b/app/javascript/components/recordings/Recordings.jsx
@@ -1,6 +1,6 @@
 import React, { useState } from 'react';
 import {
-  Card, Button, Stack,
+  Card, Stack,
 } from 'react-bootstrap';
 import useRecordings from '../../hooks/queries/recordings/useRecordings';
 import SearchBarQuery from '../shared/SearchBarQuery';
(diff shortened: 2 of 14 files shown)


app/javascript/components/recordings/RoomRecordings.jsx
```diff
@@ -11,10 +12,17 @@ import RoomsRecordingRow from './RoomsRecordingRow';
 
 export default function RoomRecordings() {
   const [input, setInput] = useState('');
+  const [page, setPage] = useState();
   const { friendlyId } = useParams();
-  const roomRecordings = useRoomRecordings(friendlyId, input);
+  const { isLoading, data: roomRecordings } = useRoomRecordings(friendlyId, input, page);
```
farhatahmad, line 17: Try `  const { isLoading, data: { data: recordings, meta: recordingsMeta } } = useRoomRecordings(friendlyId, input, page);`

hadicheaito1: I tried this before. The issue with it is that the first `data` is still null when it tries to de-structure it and access the inner `data`, resulting in a `Cannot read properties of undefined (reading 'data')` error.

That's why I did it the way I did right now since it allows us to use the `?.` safe negation operator before trying to access the inner `data`.

I agree though that there should be a way to do what we are trying to do

app/javascript/components/recordings/Recordings.jsx
```diff
@@ -25,11 +31,25 @@ export default function Recordings() {
...
+            defaultActivePage={recordings.meta.page}
+            totalPages={recordings.meta.pages}
+            onPageChange={handlePage}
+            firstItem={null}
+            lastItem={null}
+            pointing
+            secondary
+          />
+        </Container>
+        )}
```
farhatahmad, line 52: We should move this to its own component

app/assets/stylesheets/application.bootstrap.scss
```diff
@@ -8,6 +8,7 @@
 @import 'shared_accesses';
 @import 'presentation';
 @import 'admin_panel';
+@import '//cdnjs.cloudflare.com/ajax/libs/semantic-ui/2.2.2/semantic.min.css';
```
farhatahmad, line 11: This is fine for now - but we definitely shouldn't import the entire css file for semantic

hadicheaito1: I believe it's one of these, but I can't seem to find which one exactly contains the css for Pagination component

https://cdnjs.com/libraries/semantic-ui/2.2.2

farhatahmad (approved):

farhatahmad (approved):

PR #3703, opened 2022-07-27
Add Brand Color UI/frontend (NEW UI)

scouillard:
## Description

Add the UI/frontend for the Brand Color picker.

## Testing Steps

Play with the color(s) picker(s).

image


scouillard:
TODO (next PR):

I've been interchanging variables names for "brand" or "primary" colors - this needs to be fixed.

Right now, it's set as "PrimaryColor" in the DB.
I prefer using "BrandColor" than "PrimaryColor" personnally, in part because primary is used by Bootstrap already (blue color).
I would call it something like "BrandColor" and "BrandSecondaryColor".

Please LMK what you think.

app/javascript/components/admin/site_settings/appearance/BrandColor.jsx
```diff
@@ -1,15 +1,38 @@
...
   const { data: siteSettings } = useSiteSettings();
 
+  const brandDefaultColors = [
+    '#B80000', '#DB3E00', '#FCCB00', '#008B02', '#006B76', '#1273DE', '#004DCF', '#5300EB',
+  ];
+
+  const brandLightDefaultColors = [
+    '#EB9694', '#FAD0C3', '#FEF3BD', '#C1E1C5', '#BEDADC', '#C4DEF6', '#BED3F3', '#D4C4FB',
+  ];
+
```
farhatahmad, line 16: I think we should drop the default colors - they aren't useful for users

scouillard: Would you rather have only an input for the hex value?
IMO this is better than the saturation/color field.
I think it looks good and do no harm.

scouillard: I could do a full custom block without the default colors.
image
I'm not sure this will be ready for the demo tomorrow tho.

farhatahmad (approved):

PR #3704, opened 2022-07-27
Fix UI missing Brand colors (BASED ON #3703)

scouillard:
## Description

Fix some small issues regarding the Brand colors, mostly the "switch" and "nav-link".
Remove $primary and $primary-light variable from CSS.
Add $muted variable for bootstrap text-muted color.

## Screenshot
image

### DIFF
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -115,10 +115,31 @@ input.search-bar {
   }
 }
 
+.nav-link {
+  color: $muted;
+
+  &:hover {
+    color: var(--brand-color);
+  }
+}
+
+#nav-user-dropdown {
+  color: $black !important;
+  &:hover {
+    color: $black !important;
+  }
+  &:active {
+    color: $black !important;
+  }
+  &:focus {
+    color: $black !important;
+  }
+}
+
 //Brand
 :root {
   --brand-color: '';
-  --brand-color-light: whitesmoke;
+  --brand-color-light: '';
 }
 
 .text-brand {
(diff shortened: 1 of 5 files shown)


scouillard:
I will spend some time to clean up our CSS files after the demo.
Need to find a way to scope our classes properly.

app/assets/stylesheets/application.bootstrap.scss
```diff
@@ -146,3 +166,16 @@ input.search-bar {
     opacity: 0.8;
   }
 }
+
+.form-switch .form-check-input:focus {
```
KH-Amir-TN, line 170: I'm just worried about stepping on Bootstrap's toes (if he got some 😅) https://getbootstrap.com/docs/5.0/forms/checks-radios/

scouillard: Yeah changing the switch default color is super annoying for some reason.

farhatahmad (approved):

app/assets/stylesheets/application.bootstrap.scss
KH-Amir-TN: We shouldn't be using raw bootstrap in GL3.
Those `form-switch` will be refactored to use `react-bootstrap` [CHECK HERE](https://react-bootstrap.netlify.app/forms/checks-radios/#switches).

scouillard: What are you linking to? It's the only way to remove the blue color om switches as far as I know. 

I don't think it's a big deal to overwrite bootstrap components colors. I would not overwrite paddings, widths, etc.

PR #3706, opened 2022-07-28
Greenlight 3: Minor bug fixes.

KH-Amir-TN:
## Description

This PR fixes some minor bugs.
 ~~- AdminPanel->Manage Users-Create 400 Bad request (bad encapsulation on the mutation).~~
~~- Reset password is giving 400 Bad request on send reset token on reset page (bad encapsulation on the mutations).~~

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/javascript/hooks/mutations/admins/useAdminCreateUser.jsx
+++ b/app/javascript/hooks/mutations/admins/useAdminCreateUser.jsx
@@ -14,7 +14,7 @@ export default function useAdminCreateUser({ onSettled }) {
   };
 
   return useMutation(
-    (data) => axios.post('/admin/users.json', data),
+    (user) => axios.post('/admin/users.json', { user }),
     {
       onMutate: addInferredLanguage,
       onSuccess: () => {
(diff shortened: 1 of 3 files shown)


app/services/recordings_sync.rb
```diff
@@ -6,7 +6,7 @@ def initialize(params)
   end
 
   def call
-    rooms = @user.rooms
+    rooms = Room.all # TODO: Revisit this.
```
farhatahmad, line 9: Uhhh I dont think this is a good idea

KH-Amir-TN: Me either, it's just temporary.
Implementing it correctly requires more work since it may introduce its own infrastructure (ActiveJobs infrastructure, bulked batch updates,...).

farhatahmad: Let's leave it as is then

KH-Amir-TN: The server recordings will only display the user recordings that clicked the re-sync button and not all of the rooms recordings.
Is that okay for now?

farhatahmad: That's fine

farhatahmad (approved):

PR #3709, opened 2022-07-28
Moving re-sync button to Server Recordings page

hadicheaito1:
## Description

- Moving recording re-sync to admin server recordings page

image

### DIFF
--- a/app/javascript/hooks/queries/recordings/useRecordingsReSync.jsx
+++ b/app/javascript/hooks/queries/recordings/useRecordingsReSync.jsx
@@ -7,7 +7,7 @@ export default function useRecordingsReSync() {
 
   return useQuery(
     'getRecordingsResync',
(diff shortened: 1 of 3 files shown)


farhatahmad (approved):

app/javascript/hooks/queries/recordings/useRecordingsReSync.jsx
```diff
@@ -7,7 +7,7 @@ export default function useRecordingsReSync() {
 
   return useQuery(
     'getRecordingsResync',
-    () => axios.get('/recordings/resync.json').then(() => queryClient.invalidateQueries('getRecordings')),
+    () => axios.get('/recordings/resync.json').then(() => queryClient.invalidateQueries('getServerRecordings')),
```
KH-Amir-TN, line 10: Sorry for the late reply.
Just to note, I think we should do the cache invalidation in an `onSuccess` callback and maybe make this `resync` **API** more of a `POST` request because its causes a side effect and not fetches data.

PR #3713, opened 2022-07-29
Room Settings update 1: Added `MeetingOption::get_config_value`.

KH-Amir-TN:
## Description

Synchronize room settings with rooms configurations.

This PR completes **1.** 
--
~~1. Create `RoomsConfigGetter` service that can take a setting name and return its value.~~
2. Integrate the service in `RoomSettingsController#update` API.

### User story [Room setting update]:
---
1. A user authenticates.
2. A user creates a Room and access it.
[If the setting config is "optional" for the room setting]:
3. user can edit the setting status and have proper feedbacks.
[If the setting config is **not** "optional" for the room setting]:
3. user should not be able to edit the setting (by not seing the setting if it's disabled or by having the setting edit disabled when it's force enabled).

>NOTE: update API calls and access codes removal for a setting that is not configured as optional should fail with adequate error.

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/models/meeting_option.rb
+++ b/app/models/meeting_option.rb
@@ -14,12 +14,21 @@ class MeetingOption < ApplicationRecord
       .pluck(:name, :value)
   end
 
-  def self.get_value(name:, room_id:)
+  def self.get_setting_value(name:, room_id:)
     joins(:room_meeting_options)
       .select(:value)
       .find_by(
         name:,
         room_meeting_options: { room_id: }
+      )
+  end
+
+  def self.get_config_value(name:, provider:)
+    joins(:rooms_configurations)
+      .select(:value)
+      .find_by(
+        name:,
+        rooms_configurations: { provider: }
       )
   end
 end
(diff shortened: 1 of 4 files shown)


hadicheaito1 (approved):

KH-Amir-TN:
It will be used wherever we need a room config so we don't have to manually do the joining each time.
The rooms settings and configurations in this PRs will be related where one completes the other so having a dedicated service to fetch a room config will simplify the job.

It's used in #3714 and will be used in future PRs for the access codes generation and removal along with the rooms creation and fetching.

app/services/rooms_config_getter.rb
```diff
@@ -0,0 +1,17 @@
...
+    @provider = provider
+  end
+
+  def call
+    config = MeetingOption.joins(:rooms_configurations)
+                          .select(:value)
+                          .where(rooms_configurations: { provider: @provider })
+                          .find_by(name: @setting_name)
+
+    config&.value
```
farhatahmad, line 15: I'm not sure having a service here makes sense. I think this should be an action in the model instead

KH-Amir-TN: A method on the model?

farhatahmad: Yep

farhatahmad (approved):

PR #3722, opened 2022-07-29
Adding created_at to user serializer

hadicheaito1:
## Description

- Created at in manage users was not showing since it did not exist in user serializer

### DIFF
--- a/app/serializers/user_serializer.rb
+++ b/app/serializers/user_serializer.rb
@@ -3,9 +3,13 @@
 class UserSerializer < ApplicationSerializer
   include Avatarable
 
-  attributes :id, :name, :email, :provider, :language, :avatar
+  attributes :id, :name, :email, :provider, :language, :avatar, :created_at
 
   def avatar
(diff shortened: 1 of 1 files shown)


KH-Amir-TN (approved):

app/serializers/user_serializer.rb
```diff
@@ -3,9 +3,13 @@
 class UserSerializer < ApplicationSerializer
   include Avatarable
 
-  attributes :id, :name, :email, :provider, :language, :avatar
+  attributes :id, :name, :email, :provider, :language, :avatar, :created_at
```
farhatahmad, line 6: Just add this - there already is a `created_at` method in `ApplicationSerializer`

hadicheaito1: Yeah but it displays the time at the end, but in this case I don't want to display the time, so they are a bit different

farhatahmad (approved):

PR #3725, opened 2022-08-02
Access codes rectification 5: Access Codes: Added required tests to `room_spec`.

KH-Amir-TN:
## Description

Rectifying the rooms access codes to become a room setting that can be configured like any other setting.

This PR completes **5.**
--
~~1. Create and populate `glViewerAccessCode` and `glModeratorAccessCode` in `PopulateRoomsConfigurations` and `PopulateMeetingOptions`.~~
~~2. Remove `:viewer_access_code` and `:moderator_access_code` migrations.~~
~~3. Create `Room#viewer_access_code`, `Room#moderator_access_code`, `Room#generate_viewer_access_code`,
`Room#generate_moderator_access_code`, `Room#remove_viewer_access_code`, `Room#remove_moderator_access_code`.~~
~~4. Update `RoomsController (#generate_access_code, #remove_access_code)` and tests `(meeting_controller_spec and rooms_controller_spec)`.~~
~~5. Update and extend model spec `room_spec`.~~
6. Update `RoomSettings#update` **API** to restrict edit to codes.

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/spec/models/room_spec.rb
+++ b/spec/models/room_spec.rb
@@ -3,9 +3,9 @@
 require 'rails_helper'
 
 RSpec.describe Room, type: :model do
-  describe 'before_validations' do
-    let!(:room) { create(:room) }
+  let!(:room) { create(:room) }
 
+  describe 'before_validations' do
     describe '#set_friendly_id' do
       it 'sets a rooms friendly_id before creating' do
         expect(room.friendly_id).to be_present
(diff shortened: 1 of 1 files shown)


KH-Amir-TN:
# DEPENDS ON #3725

spec/models/room_spec.rb
```diff
@@ -54,6 +54,111 @@
     end
   end
 
+  describe 'private methods' do
+    describe '#generate_code' do
+      before do
+        allow_any_instance_of(described_class).to receive(:generate_code).and_call_original
```
KH-Amir-TN, line 60: Some may not like this because they believe  a private method should remain private even in tests and only functional test it through public methods that uses it.

While others will simply say a private method isn't that private after all, if not suspect to frequent changes in its API and unit testing it is possible without introducing a lot of complexity it may make sense and doesn't do any harm.

Don't know what school you like guys 😆 
Tell me what you think,

@farhatahmad @scouillard @hadicheaito1

spec/models/room_spec.rb
```diff
@@ -54,6 +54,111 @@
...
+        allow(SecureRandom).to receive(:alphanumeric).and_return('TEST')
+        expect(SecureRandom).to receive(:alphanumeric).with(6)
+        expect(room.generate_code).to eq('test')
+      end
+    end
+  end
+
+  describe '#viewer_access_code' do
+    it 'calls #get_setting with name "glViewerAccessCode" and returns the value' do
+      allow_any_instance_of(described_class).to receive(:get_setting).and_return(instance_double(RoomMeetingOption, value: 'CODE'))
```
farhatahmad, line 73: Can we stay consistent with the other tests and not use `instance_double`?

KH-Amir-TN: I'm afraid that this is different,
It's the same case such as [THIS](https://github.com/bigbluebutton/greenlight/blob/v3/spec/models/room_spec.rb#L36) and [THIS](https://github.com/bigbluebutton/greenlight/blob/v3/spec/models/room_spec.rb#L43).
We're expecting an object with a method `value`.

KH-Amir-TN: More is listed in #3626

farhatahmad (approved):

PR #3729, opened 2022-08-03
Add PermissionsChecker service

scouillard:
## Description

Add a PermissionsChecker service to be called in a controllers `before_action`.

The PermissionsChecker returns a bool.

The PermissionsChecker will return true if the request to the endpoint either contains:

1. a current_user that has the appropriate permission to act on a resource
2. a params[:resource_id] and a current_user that is trying to act on its own resource as per the matching of current_user.resource_id == params[:ressource_id]

Else, the PermissionsChecker will return false and an unauthorized status.

## Testing Steps

- Try to edit the users in the Admin/ManageUsers tab without having the "ManageUsers" permission
- Try to edit the users in the Admin/ManageUsers tab while having the "ManageUsers" permission
- Try to edit your own profile via the Admin/ManageUsers tab and the Profile tab.

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -5,7 +5,7 @@ module Api
     class UsersController < ApiController
       skip_before_action :ensure_authenticated, only: %i[create]
 
-      before_action only: %i[update destroy purge_avatar] do
+      before_action only: %i[update purge_avatar] do
         ensure_authorized('ManageUsers')
       end
 
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -5,6 +5,10 @@ module Api
     class UsersController < ApiController
       skip_before_action :ensure_authenticated, only: %i[create]
 
+      before_action only: %i[update destroy purge_avatar] do
+        ensure_authorized('ManageUsers')
+      end
+
       def show
         user = User.find(params[:id])
 
--- a/app/controllers/concerns/authorizable.rb
+++ b/app/controllers/concerns/authorizable.rb
@@ -13,6 +13,17 @@ module Authorizable
     return render_error status: :unauthorized unless current_user
   end
 
+  # PermissionsChecker service will return a true or false depending on whether the current_user's role has the provided permission_name
+  def ensure_authorized(permission_name)
+    return render_error status: :unauthorized unless PermissionsChecker.new(
+      current_user:,
+      permission_name:,
+      user_id: params[:id],
+      friendly_id: params[:friendly_id],
+      record_id: params[:record_id]
+    ).call
+  end
+
   private
 
   # Ensures that requests to the API are explicit enough.
--- a/app/controllers/concerns/authorizable.rb
+++ b/app/controllers/concerns/authorizable.rb
@@ -16,7 +16,7 @@ module Authorizable
   # PermissionsChecker service will return a true or false depending on whether the current_user's role has the provided permission_name
   def ensure_authorized(permission_name)
     # TODO: - should this return a status: :not_found instead of :unauthorized if the user doesn't have the permission AND the param doesn't exists?
-    return render_error status: :unauthorized unless PermissionsChecker.new(
+    return render_error status: :forbidden unless PermissionsChecker.new(
       current_user:,
       permission_name:,
       user_id: params[:id],
--- a/app/controllers/concerns/authorizable.rb
+++ b/app/controllers/concerns/authorizable.rb
@@ -15,6 +15,7 @@ module Authorizable
 
   # PermissionsChecker service will return a true or false depending on whether the current_user's role has the provided permission_name
   def ensure_authorized(permission_name)
+    # TODO: - should this return a status: :not_found instead of :unauthorized if the user doesn't have the permission AND the param doesn't exists?
     return render_error status: :unauthorized unless PermissionsChecker.new(
       current_user:,
       permission_name:,
--- a/app/services/permissions_checker.rb
+++ b/app/services/permissions_checker.rb
@@ -11,8 +11,7 @@ class PermissionsChecker
 
   def call
     return true if RolePermission.joins(:permission)
-                                 .where(role_id: @current_user.role_id)
-                                 .find_by(permission: { name: @permission_name })&.value == 'true'
+                                 .find_by(role_id: @current_user.role_id, permission: { name: @permission_name })&.value == 'true'
 
     case @permission_name
     when 'ManageUsers'
--- a/app/services/permissions_checker.rb
+++ b/app/services/permissions_checker.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class PermissionsChecker
-  def initialize(current_user:, permission_name:, user_id:, friendly_id:, record_id:)
+  def initialize(current_user:, permission_name:, user_id: '', friendly_id: '', record_id: '')
     @current_user = current_user
     @permission_name = permission_name
     @user_id = user_id
@@ -29,7 +29,7 @@ class PermissionsChecker
   private
 
   def authorize_manage_user
-    @current_user.id.to_s == @user_id
+    @current_user.id.to_s == @user_id.to_s
   end
 
   def authorize_manage_rooms
--- a/app/services/permissions_checker.rb
+++ b/app/services/permissions_checker.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+class PermissionsChecker
+  def initialize(current_user:, permission_name:, user_id:, friendly_id:, record_id:)
+    @current_user = current_user
+    @permission_name = permission_name
+    @user_id = user_id
+    @friendly_id = friendly_id
+    @record_id = record_id
+  end
+
+  def call
+    return true if RolePermission.joins(:permission)
+                                 .where(role_id: @current_user.role_id)
+                                 .find_by(permission: { name: @permission_name })&.value == 'true'
+
+    case @permission_name
+    when 'ManageUsers'
+      return authorize_manage_user
+    when 'ManageRooms'
+      return authorize_manage_rooms
+    when 'ManageRecordings'
+      return authorize_manage_recordings
+    end
+
+    false
+  end
+
+  private
+
+  def authorize_manage_user
+    @current_user.id.to_s == @user_id
+  end
+
+  def authorize_manage_rooms
+    @current_user.rooms.find_by(friendly_id: @friendly_id).present?
+  end
+
+  def authorize_manage_recordings
+    @current_user.recordings.find_by(record_id: @record_id).present?
+  end
+end
--- a/spec/controllers/admin/users_controller_spec.rb
+++ b/spec/controllers/admin/users_controller_spec.rb
@@ -60,7 +60,7 @@ RSpec.describe Api::V1::Admin::UsersController, type: :controller do
       room_valid_params = { name: 'Awesome Room' }
       create(:role_permission, permission: manage_users_permission, role:, value: 'false', provider: 'greenlight')
       expect { post :create_server_room, params: { user_id: user.id, room: room_valid_params } }.not_to(change { user.rooms.count })
-      expect(response).to have_http_status(:unauthorized)
+      expect(response).to have_http_status(:forbidden)
     end
 
     it 'returns :not_found for unfound users' do
--- a/spec/controllers/admin/users_controller_spec.rb
+++ b/spec/controllers/admin/users_controller_spec.rb
@@ -6,6 +6,13 @@ RSpec.describe Api::V1::Admin::UsersController, type: :controller do
   let(:role) { create(:role) }
   let(:user) { create(:user, role:) }
   let(:manage_users_permission) { create(:permission, name: 'ManageUsers') }
+  let!(:manage_users_role_permission) do
+    create(:role_permission,
+           role_id: user.role_id,
+           permission_id: manage_users_permission.id,
+           value: 'true',
+           provider: 'greenlight')
+  end
 
   before do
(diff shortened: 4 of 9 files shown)


app/services/permissions_checker.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class PermissionsChecker
+  def initialize(current_user, permission_name)
+    @current_user = current_user
+    @permission = Permission.find_by(name: permission_name)
+  end
+
+  def call
+    return true if @current_user.role.role_permissions.find_by(permission_id: @permission.id).value == 'true'
```
hadicheaito1, line 10: Just learned that this is actually firing multiple queries.

I believe that something like this is the preferred method (one query):
 ```
RolePermission.joins(:permission)
  .where(provider: 'greenlight')
  .where(role_id: current_user.role_id)
  .find_by(permission: { name: @permission.name }).value == 'true' 
```

Comment (scouillard) [reply app/services/permissions_checker.rb]: Indeed, this query is almost twice as fast

File: app/controllers/concerns/authorizable.rb [hunk 2]
```diff
@@ -13,6 +13,11 @@ def ensure_authenticated
     return render_error status: :unauthorized unless current_user
   end
 
+  # PermissionsChecker service will return a true or false depending on whether the current_user's role has the provided permission_name
+  def ensure_authorized(permission_name)
+    return render_error status: :unauthorized unless ::PermissionsChecker.new(current_user, permission_name, params).call
```
KH-Amir-TN, line 18: I believe we can omit the `return`.

KH-Amir-TN: For consistency only, we're making our services use named not positional arguments.

KH-Amir-TN: Why the explicit `::PermissionsChecker` and not `PermissionsChecker`?

app/services/permissions_checker.rb
```diff
@@ -0,0 +1,30 @@
...
+
+class PermissionsChecker
+  def initialize(current_user, permission_name, params)
+    @current_user = current_user
+    @permission = Permission.find_by(name: permission_name)
+    @params = params
+  end
+
+  def call
+    return true if RolePermission.joins(:permission).where(role_id: @current_user.role_id).find_by(permission_id: @permission.id).value == 'true'
```
KH-Amir-TN, line 11:
```suggestion
    return true if RolePermission.joins(:permission).where(role_id: @current_user.role_id).find_by(permission: { name: @permission_name})&.value == 'true'
```

app/services/permissions_checker.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class PermissionsChecker
+  def initialize(current_user, permission_name, params)
+    @current_user = current_user
+    @permission = Permission.find_by(name: permission_name)
```
KH-Amir-TN, line 6:
```suggestion
    @permission_name = permission_name
```

app/services/permissions_checker.rb
```diff
@@ -0,0 +1,30 @@
...
+  def call
+    return true if RolePermission.joins(:permission).where(role_id: @current_user.role_id).find_by(permission_id: @permission.id).value == 'true'
+    return true if authorize_manage_user || authorize_manage_rooms || authorize_manage_recordings
+
+    false
+  end
+
+  private
+
+  def authorize_manage_user
```
KH-Amir-TN, line 19: I don't think that this what we're going with,
APIs restriction will only be used for APIs scoped under `Admin` namespace.
All of the other, regular APIs will be rectified to work only on the authenticated user resources such as its rooms, recordings,... 
The regular APIs will expect authorize the user if  it's authenticated or not depending on the case.

With that being said all of these private methods functionality may not be essential in the upcoming weeks ,

KH-Amir-TN: Also, the way of using the `params` as is, fitting all possible scenarios at once is a bit error prone,

app/controllers/api/v1/users_controller.rb
```diff
@@ -5,6 +5,10 @@ module V1
     class UsersController < ApiController
       skip_before_action :ensure_authenticated, only: %i[create]
 
+      before_action only: %i[update destroy purge_avatar] do
```
KH-Amir-TN, line 8: I think that `ensure_authorized('ManageUsers')` should concern only all o the actions of  `Admin::UsersController`,
The explanation is given down below,

app/services/permissions_checker.rb
KH-Amir-TN: This will save us an extra query (CHECK BELOW),

app/controllers/concerns/authorizable.rb
scouillard: My IDE was complaining for the ::
Every method in Authorizable.rb has an explicit `return`

app/services/permissions_checker.rb
scouillard: Nice

app/controllers/api/v1/users_controller.rb
scouillard: Admin seems to be using regular user endpoints.

app/services/permissions_checker.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class PermissionsChecker
+  def initialize(current_user, permission_name, params)
```
farhatahmad, line 4: As Amir mentioned - switch these to named arguments

app/controllers/api/v1/users_controller.rb
farhatahmad, line 8: `before_action :ensure_authorized('ManageUsers`), only: %i[update destroy purge_avatar]

app/services/permissions_checker.rb
```diff
@@ -0,0 +1,30 @@
...
+    @current_user = current_user
+    @permission = Permission.find_by(name: permission_name)
+    @params = params
+  end
+
+  def call
+    return true if RolePermission.joins(:permission).where(role_id: @current_user.role_id).find_by(permission_id: @permission.id).value == 'true'
+    return true if authorize_manage_user || authorize_manage_rooms || authorize_manage_recordings
```
farhatahmad, line 12: I think I would move the `@permission.name` check to a switch statement here

```
case @permission.name
when `ManageUsers`
  authorize_manage_user
....
```

app/services/permissions_checker.rb
scouillard: Isn't it what we are doing here?

app/controllers/api/v1/users_controller.rb
scouillard: Can't pass an arg to :ensure_authorized, it needs to be in a block.

app/services/permissions_checker.rb
scouillard: I'll double check but I think we are using regular endpoint for most things in the admin panel.

scouillard: I agree tho, I think they should be separate endpoints.

scouillard: I've been going through your comment a couple of time, and I think we have a different vision of how auth could/should be. Maybe you could provide an example for tomorrow's (5 aug) meeting?

app/services/permissions_checker.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+class PermissionsChecker
+  def initialize(current_user:, permission_name:, user_id:, friendly_id:, record_id:)
```
hadicheaito1, line 4: Maybe `permission_name:, user_id:, friendly_id:, record_id:` should each have default values set

hadicheaito1: Now that I think of it, it isn't needed so nvm

farhatahmad (dismissed):

farhatahmad:
@scouillard fix the specs and I'll merge

scouillard:
> 

Im on it

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -3,7 +3,9 @@
 require 'rails_helper'
 
 RSpec.describe Api::V1::Admin::UsersController, type: :controller do
-  let(:user) { create(:user) }
+  let(:role) { create(:role) }
+  let(:user) { create(:user, role:) }
+  let(:manage_users_permission) { create(:permission, name: 'ManageUsers') }
```
scouillard, line 8: @farhatahmad 
Following our call, I tried to add
`let(:role_permission) { create(:role_permission, permission: manage_users_permission, role:, value: 'true', provider: 'greenlight') }` but it would not work. I tried to debug but I had to let it go. It has been added to every endpoint the role_permission is needed.

app/controllers/concerns/authorizable.rb
```diff
@@ -13,6 +13,18 @@ def ensure_authenticated
...
+  def ensure_authorized(permission_name)
+    # TODO - should this return a status: :not_found instead of :unauthorized if the user doesn't have the permission AND the param doesn't exists?
+    return render_error status: :unauthorized unless PermissionsChecker.new(
+      current_user:,
+      permission_name:,
+      user_id: params[:id],
+      friendly_id: params[:friendly_id],
+      record_id: params[:record_id]
+    ).call
+  end
```
scouillard, line 26: The PermissionsChecker works this way:
The request either has:
1. a current_user that has the appropriate permission to act on a resource
2. the current_user is trying to act on its own resource as per the matching of current_user.resource_id == params[:ressource_id]

An awkward situation that could arrive is if the request does not have:
1. a current user with the appropriate permission
3. the provided param is not a valid user_id, friendly_id, or record_id

spec/controllers/users_controller_spec.rb
```diff
@@ -32,59 +32,52 @@
...
-      user = create(:user)
       expect(response).to have_http_status(:ok)
       expect { delete :destroy, params: { id: user.id } }.to change(User, :count).by(-1)
     end
 
     it 'does not delete any user if the user id is invalid' do
       expect { delete :destroy, params: { id: 'invalid-id' } }.not_to change(User, :count)
-      expect(response).to have_http_status(:not_found)
+      # TODO - maybe ensure_authorized should return a :not_found if the user is not authorized AND the param doesn't exists?
+      expect(response).to have_http_status(:unauthorized)
```
scouillard, line 80: Related to https://github.com/bigbluebutton/greenlight/pull/3729/files#r938277438

app/services/permissions_checker.rb
KH-Amir-TN: Sure, will sync in our meeting

app/services/permissions_checker.rb
```diff
@@ -0,0 +1,42 @@
...
+    @permission_name = permission_name
+    @user_id = user_id
+    @friendly_id = friendly_id
+    @record_id = record_id
+  end
+
+  def call
+    return true if RolePermission.joins(:permission)
+                                 .where(role_id: @current_user.role_id)
+                                 .find_by(permission: { name: @permission_name })&.value == 'true'
```
KH-Amir-TN, line 15: I think you can further remove the `where` and use only the `find_by(role_id: ..., permission: { ... })`

app/controllers/concerns/authorizable.rb
```diff
@@ -13,6 +13,18 @@ def ensure_authenticated
     return render_error status: :unauthorized unless current_user
   end
 
+  # PermissionsChecker service will return a true or false depending on whether the current_user's role has the provided permission_name
+  def ensure_authorized(permission_name)
+    # TODO: - should this return a status: :not_found instead of :unauthorized if the user doesn't have the permission AND the param doesn't exists?
+    return render_error status: :unauthorized unless PermissionsChecker.new(
```
KH-Amir-TN, line 19: Isn't the status `forbidden` more adequate  than `unauthorized` or we're obscuring the responses?

scouillard: Why would we obscure this response?
I was hesitating between `unauthorized` and `not_found`

app/controllers/concerns/authorizable.rb
hadicheaito1: I feel that :unauthorized is fine

hadicheaito1: It's a bit more specific to what the error actually is

scouillard:
The fix for the duplication of the destroy method in users_controller and admin/users_controller - and the associated RSpec - will be in another PR as this needs to be merged ASAP.

spec/controllers/admin/users_controller_spec.rb
hadicheaito1: This should be possible :
` create(:role_permission, role_id: user.role.id, permission_id: permission.id, value: 'true', provider: 'greenlight')`.

This way we won't need to add the role_permission every test, but only to the test where you need to change the value to 'false'

scouillard: It doesn't seem to work. It should tho.

image

hadicheaito1: Are you adding it in the before do block?

scouillard: No, I put it on top of everything, like the other let(:<something>)

image

scouillard: So I just learnt the let(:something) does not instantiate the object if the variable is not called.

hadicheaito1: To force the initiation, use let!.

I think rubocop complains tho when using let!, that's why play around with it and see, but I think it has to be in before do

scouillard: let! works, thanks for the help.

hadicheaito1 (approved):

PR #3736, opened 2022-08-08
Access Codes: Added Access codes auto generation on `after_find` and `after_create` for certain configs.

KH-Amir-TN:
## Description

This PR:

- Edits `Rooms` model to add in the capability to lazy generate access codes on rooms creation and first interaction after having forced enabled configs.

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -16,7 +16,13 @@ class Room < ApplicationRecord
   validates :meeting_id, presence: true, uniqueness: true
 
   before_validation :set_friendly_id, :set_meeting_id, on: :create
+
   after_create :create_meeting_options
+  after_create :generate_viewer_access_code, if: :auto_generate_viewer_access_code?
+  after_create :generate_moderator_access_code, if: :auto_generate_moderator_access_code?
+
+  after_find :generate_viewer_access_code, if: :auto_generate_viewer_access_code?
+  after_find :generate_moderator_access_code, if: :auto_generate_moderator_access_code?
 
   attr_accessor :shared, :active, :participants
 
@@ -89,5 +95,17 @@ class Room < ApplicationRecord
 
   def generate_code
     SecureRandom.alphanumeric(6).downcase
+  end
+
+  def auto_generate_viewer_access_code?
+    config = MeetingOption.get_config_value(name: 'glViewerAccessCode', provider: 'greenlight')&.value
+
+    viewer_access_code.blank? && config == 'true'
+  end
+
+  def auto_generate_moderator_access_code?
+    config = MeetingOption.get_config_value(name: 'glModeratorAccessCode', provider: 'greenlight')&.value
+
+    moderator_access_code.blank? && config == 'true'
   end
 end
--- a/spec/models/room_spec.rb
+++ b/spec/models/room_spec.rb
@@ -5,26 +5,216 @@ require 'rails_helper'
 RSpec.describe Room, type: :model do
   let!(:room) { create(:room) }
 
-  describe 'before_validations' do
-    describe '#set_friendly_id' do
-      it 'sets a rooms friendly_id before creating' do
-        expect(room.friendly_id).to be_present
+  context 'callbacks' do
+    context 'before_validations' do
+      describe '#set_friendly_id' do
+        it 'sets a rooms friendly_id before creating' do
+          expect(room.friendly_id).to be_present
+        end
+
+        it 'prevents duplicate friendly_ids' do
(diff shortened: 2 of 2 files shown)


app/models/room.rb
```diff
@@ -16,7 +16,13 @@ class Room < ApplicationRecord
   validates :meeting_id, presence: true, uniqueness: true
 
   before_validation :set_friendly_id, :set_meeting_id, on: :create
+
   after_create :create_meeting_options
+  after_create :generate_viewer_access_code, if: :auto_generate_viewer_access_code?
+  after_create :generate_moderator_access_code, if: :auto_generate_moderator_access_code?
+
+  after_find :generate_viewer_access_code, if: :auto_generate_viewer_access_code?
+  after_find :generate_moderator_access_code, if: :auto_generate_moderator_access_code?
```
scouillard, line 25: Could we use `after_initialize` instead of both callbacks?

> The [after_initialize](https://api.rubyonrails.org/v7.0.3.1/classes/ActiveRecord/Callbacks/ClassMethods.html#method-i-after_initialize) callback will be called whenever an Active Record object is instantiated, either by directly using new or when a record is loaded from the database.

KH-Amir-TN: Here's an inclusive response hope that you don't get bored 🤣 

The problem is due to the nature of `after_initialize`.
The callback will be called whenever a new instance of the Room model is created, wither after a fetching operation or when attempting to create a new record.
When attempting to create a new record the `after_initialize` will be one of the first if not the first callback to be called for a record that still not `:persisted?` in the DB.

The access codes are room meeting options records where each refers to a room and a meeting option.
Auto generating the access codes is part of the fallowing pipeline:
1. Room model gets instantiated and populated.
2. Room gets its unique meeting and friendly ids set -> `before_validation :set_friendly_id, :set_meeting_id, on: :create`.
3. Room record gets persisted to the DB.
4. Room meeting options records gets auto generated and persisted to the DB with their corresponding default values -> `after_create :create_meeting_options`.
5. Room access codes meeting options will **update** their default values with an auto generated codes for certain scenarios.

So, with such pipeline the access codes auto generation on create requires having the room meeting options records in place, which in turn requires having the room record itself persisted in the DB.

That was the reasoning behind avoiding the `after_initialize` since it will cause us additional queries without any benefit (**performance loss**).

What we can do instead is to leave only the `after_find` callback which will give the expected outcome but that means in some expected **frequent** cases additional waiting time on any feature in GL3 that fetches a room record (like the join page, rooms page, room page, admin server rooms,... ) and that's why we wanted to push that waiting time whenever possible to the creation process for a better UX.

app/models/room.rb
```diff
@@ -16,7 +16,13 @@ class Room < ApplicationRecord
   validates :meeting_id, presence: true, uniqueness: true
 
   before_validation :set_friendly_id, :set_meeting_id, on: :create
+
   after_create :create_meeting_options
+  after_create :generate_viewer_access_code, if: :auto_generate_viewer_access_code?
+  after_create :generate_moderator_access_code, if: :auto_generate_moderator_access_code?
```
farhatahmad, line 22: Why wouldn't we just move this logic into the `create_meeting_options`?

app/models/room.rb
farhatahmad, line 25: I'm not sure how much I like this - this adds an extra query for a lot of our endpoints

app/models/room.rb
KH-Amir-TN: `create_meeting_options` will launch only after room creation.
The access codes auto generation case isn't limited to creation but also to the first fetching opertation (find).
Besides the access codes generation on room creation requires having the room meeting options in place therefore splitting that logic into a separate callback will reduce the complexity of `create_meeting_options` and improve readability.

More explanation on this is given above.

app/models/room.rb
KH-Amir-TN: Yup, access codes aren't room properties now.
So, no matter what we'd have to make an extra query to fetch the room codes.
I haven't payed optimizing this logic any of my time,

We could create one callback concerned with the access codes auto generation, that callback will fetch both of the room codes with one query instead of having to make two with this current trivial implementation.
But, at least one query per room on each find will still be issued with the current design.

To eliminate the need for that we may consider falling back to access codes properties but we pay its design more time to reduce spatial (memory) wastage.

KH-Amir-TN: This logic was optimized in #3762

PR #3737, opened 2022-08-08
Access Codes: Extended the access codes fetching logic to hide the codes for forced disabled configs.

KH-Amir-TN:
## Description

This PR:

- Edits `Rooms` model to extend the access codes fetching logic to hide their values when having a forced disabled config.

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -31,10 +31,16 @@ class Room < ApplicationRecord
   end
 
   def viewer_access_code
+    config = MeetingOption.get_config_value(name: 'glViewerAccessCode', provider: 'greenlight')&.value
+    return nil if config == 'false'
+
     get_setting(name: 'glViewerAccessCode')&.value
   end
 
   def moderator_access_code
+    config = MeetingOption.get_config_value(name: 'glModeratorAccessCode', provider: 'greenlight')&.value
+    return nil if config == 'false'
+
     get_setting(name: 'glModeratorAccessCode')&.value
   end
 
--- a/spec/models/room_spec.rb
+++ b/spec/models/room_spec.rb
@@ -69,20 +69,54 @@ RSpec.describe Room, type: :model do
   end
 
   describe '#viewer_access_code' do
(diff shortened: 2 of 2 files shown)


scouillard (dismissed):

app/models/room.rb
```diff
@@ -31,10 +31,16 @@ def anyone_joins_as_moderator?
...
+    return nil if config == 'false'
+
     get_setting(name: 'glViewerAccessCode')&.value
   end
 
   def moderator_access_code
+    config = MeetingOption.get_config_value(name: 'glModeratorAccessCode', provider: 'greenlight')&.value
+    return nil if config == 'false'
+
     get_setting(name: 'glModeratorAccessCode')&.value
```
farhatahmad, line 44: Wouldn't you just call the RoomSettingGetter instead?

KH-Amir-TN: Because this is a config **not** a room list of settings.
- `RoomSettingGetter` will return a `Hash(<name> => <value>)` of a given room settings values while taking into consideration the available rooms configurations.
- `MeetingOption.get_config_value` which was introduced on #3713 will return the value of a rooms configuration i.e. a forced enabled ``"record"` configuration.

KH-Amir-TN: At the moment of this writing, we don't have a `RoomSettingGetter` that fetches by name a room setting.
We have a `RoomSettingsGetter` service that fetches a room `Hash(<name> => <value>)` of settings while considering the application rooms configuration and another service `SettingGetter` that returns the value for a given site setting while binarizing it.

Having a service that fetches only a room setting does not have to be on a service since it does not exceed in terms of functionality what is handled by the data layer.

Isn't the same reason why we removed the `RoomConfigGetter` service in favor of `MeetingOption.get_config_value`?

KH-Amir-TN:
# REBASED ON e35a8820c41bdf1560c8e65438e8ccae4c415bc2

PR #3740, opened 2022-08-08
Add Pagination Styling

scouillard:
## Description

Add Pagination Styling.

NOTE:

- Recordings table/card HTML layout needs to be fixed - there should not be borders all around.
- Admin Tables (manage users, server recordings, server rooms, etc) HTML layout need to be uniformized.

## Testing Steps

Play with pagination

image


app/javascript/components/shared/Pagination.jsx
```diff
@@ -0,0 +1,60 @@
...
+import PaginationButton from './PaginationPrevButton';
+
+export default function Pagination({
+  page, totalPages, setPage, borders,
+}) {
+  const handlePage = (e, { activePage }) => {
+    const gotopage = { activePage };
+    const pagenum = gotopage.activePage;
+    setPage(pagenum);
+  };
```
KH-Amir-TN, line 13:
```suggestion
  const handlePage = (e, { activePage }) => { setPage(activePage); };
```

app/javascript/components/shared/Pagination.jsx
```diff
@@ -0,0 +1,60 @@
...
+  const paginationWrapper = () => {
+    if (borders === true) {
+      return 'pagination-wrapper';
+    }
+    return '';
+  };
+
+  if (totalPages > 0) {
+    return (
+      <div className={paginationWrapper()}>
```
KH-Amir-TN, line 25:
```suggestion
      <div className={borders ? 'pagination-wrapper' : ''}>
```

app/javascript/components/shared/Pagination.jsx
scouillard: This is from a previous PR, will this work @hadicheaito1

db/seeds.rb
```diff
@@ -1,8 +1,21 @@
 # frozen_string_literal: true
+
 # This file should contain all the record creation needed to seed the database with its default values.
 # The data can then be loaded with the bin/rails db:seed command (or created alongside the database with db:setup).
 #
 # Examples:
 #
 #   movies = Movie.create([{ name: "Star Wars" }, { name: "Lord of the Rings" }])
 #   Character.create(name: "Luke", movie: movies.first)
+require 'faker'
```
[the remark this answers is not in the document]
KH-Amir-TN:
```
require 'factory_bot_rails'
require 'faker'

include FactoryBot::Syntax::Methods

create_list(:user, 5, password: 'password', provider: 'Amir')
```

app/javascript/components/shared/Pagination.jsx
hadicheaito1: i think you would need to do `const handlePage = (e, { activePage }) => { setPage(activePage.activePage); };`. It might not work though since right now `gotopage = {activePage} `and not just `activePage`

db/seeds.rb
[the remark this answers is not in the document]
hadicheaito1: I agree with Amir

app/javascript/components/shared/Pagination.jsx
scouillard: Amir was right.

farhatahmad:
Considering the fact that we're planning to remove Semantic UI - I wouldn't put in any effort to this. The css file is too big and will cause us many problems in the future - we need to find a way around it

scouillard:
> Considering the fact that we're planning to remove Semantic UI - I wouldn't put in any effort to this. The css file is too big and will cause us many problems in the future - we need to find a way around it

I didn't know we were not keeping Semantic UI

scouillard:
I scoped the CSS classes. It should not cause us issues down the road I think?

KH-Amir-TN:
> > Considering the fact that we're planning to remove Semantic UI - I wouldn't put in any effort to this. The css file is too big and will cause us many problems in the future - we need to find a way around it
> 
> I didn't know we were not keeping Semantic UI

Me either,

app/javascript/components/admin/server_rooms/ServerRooms.jsx
```diff
@@ -58,15 +58,18 @@ export default function ServerRooms() {
                           )}
                       </tbody>
                     </Table>
+                    {!isLoading
+                      && (
+                        <div className="table-pagination-wrapper">
```
KH-Amir-TN, line 63: At this step of development, Let's start considering using `react-bootstrap` components instead of raw html elements whenever possible,
A `Container` will render a div by default with the supervision of our design lib.

So, `<Container className="table-pagination-wrapper"> ... </Container>` will do the trick,

app/javascript/components/admin/server_recordings/ServerRecordings.jsx
```diff
@@ -51,15 +51,18 @@ export default function ServerRecordings() {
...
                           RecordingRow={ServerRecordingRow}
                         />
+                        {!isLoading
+                          && (
+                            <div className="table-pagination-wrapper">
+                              <Pagination
+                                page={serverRecordings.meta.page}
+                                totalPages={serverRecordings.meta.pages}
+                                setPage={setPage}
+                                borders
```
KH-Amir-TN, line 61: With the `table-pagination-wrapper` css class in place I believe that the borders property is absurd,

app/javascript/components/admin/manage_users/ActiveUsers.jsx
```diff
@@ -13,11 +13,14 @@ export default function ActiveUsers({ input }) {
...
-          page={activeUsers.meta.page}
-          totalPages={activeUsers.meta.pages}
-          setPage={setPage}
-        />
+          <div className="table-pagination-wrapper">
+            <Pagination
+              page={activeUsers.meta.page}
+              totalPages={activeUsers.meta.pages}
+              setPage={setPage}
+              borders
```
KH-Amir-TN, line 21: There's no use of `borders` prop on `Pagination` component,

app/javascript/components/admin/server_rooms/ServerRooms.jsx
```diff
@@ -58,15 +58,18 @@ export default function ServerRooms() {
...
+                          <Pagination
+                            page={serverRooms.meta.page}
+                            totalPages={serverRooms.meta.pages}
+                            setPage={setPage}
+                            borders
```
KH-Amir-TN, line 68: Same here,

app/assets/stylesheets/application.bootstrap.scss
```diff
@@ -190,3 +190,54 @@ input.search-bar {
 .form-control:focus {
   box-shadow: none;
 }
+
+
+//Semantic UI pagination
+.table-pagination-wrapper {
```
KH-Amir-TN, line 196: Since this has nothing to do with a table wrapping the pagination component but rather a bordered container we might call this something more informational i.e. `.container-bordered` and maybe make it a ` border: 2px solid whitesmoke;` if the extra border bottom isn't a problem.

KH-Amir-TN: `<Container className="border border-light border-2 border-top-0"> ... </Container>`

app/javascript/components/admin/server_rooms/ServerRooms.jsx
scouillard: I disagree with this.

I want only the wrapping class, not a container.

Stacking layers of containers - or whatever Bootstrap layout classes - will cause problems (extra paddings, margins, etc), especially with responsiveness.

app/javascript/components/shared_components/PaginationPrevButton.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import Button from 'react-bootstrap/Button';
+import { ArrowLeftIcon } from '@heroicons/react/outline';
+import { ArrowRightIcon } from '@heroicons/react/solid';
+import PropTypes from 'prop-types';
+
+export default function PaginationButton({ direction }) {
+  if (direction === 'Previous') {
+    return (
+      <Button variant="brand-backward">
```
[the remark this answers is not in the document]
scouillard: I think it's fine as it currently is.
This button style is used thoroughly, not only for pagination.

app/assets/stylesheets/application.bootstrap.scss
scouillard: I can change the name of the class.
Using bootstrap border classes if feasible, but not in this PR.
We need to clean/uniformize the Admin panel tables, as every table has been made a different way.

KH-Amir-TN: That's good,
For the divergence of all admin panel tabs, that is due to the decision we took by reducing complexity and duplicating the panels avoiding the use of `<Outlet />`.
This might help #3590

app/javascript/components/shared_components/PaginationPrevButton.jsx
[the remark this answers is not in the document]
KH-Amir-TN: Oh, It's just the naming scheme,
I was wondering if the word **backward** is adequate as a variant name, it gave me an insight that it's the variant for a backward (previous) button and not a generic button variant,
It's not an issue, it may be because I don't fully understand the meaning behind **backward**,

app/javascript/components/admin/server_rooms/ServerRooms.jsx
KH-Amir-TN: The `Container` will only have `container` class as I recall, the `Row` component will have the `row` class.

For responsiveness I believe it's quite the opposite, when using raw `div` elements it's true that we will have the default `div` with no extra stuff but that means that bootstrap will not handle that element and therefore we may expect issues on responsiveness.
 As set by the bootstrap doc _"Containers are the most basic layout element in Bootstrap and are **required when using our default grid system**."_ 
 But for the record since we have a top level container it's not required to nest containers as set in the docs  _" While containers can be nested, most layouts do not require a nested container."_ 
But, it still a best practice because with react each component encapsulates its UI and can be used in other projects,

[More](https://getbootstrap.com/docs/5.0/layout/containers/#how-they-work)

scouillard: I avoid nesting layout elements.

I don't see why we shouldn't use regular <div> because we are using bootstrap.

I am confident using a <container> in this particular place will mess up the layout responsiveness.

app/assets/stylesheets/application.bootstrap.scss
scouillard: Exactly, I have no problem with the duplication.

They should be uniformized, not DRY-ed.

It seems like every table has been built differently, and now behaves differently.

app/javascript/components/shared_components/PaginationPrevButton.jsx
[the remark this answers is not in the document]
scouillard: Yeah, maybe "brand-outline" would be more adequate.

app/assets/stylesheets/application.bootstrap.scss
KH-Amir-TN: I think if it's gonna take sometime then this may be better to be done at the end of development otherwise after uniformizing them now they may diverge again in the upcoming weeks,

scouillard: Good point.

farhatahmad (approved):

PR #3749, opened 2022-08-10
ManageRoles Permission

hadicheaito1:
## Description

- Adding check for ManageRoles permission before accessing roles endpoints
- Added rspec tests to check for forbidden status return
- Removes the Roles tab from AdminNavSideBar if current user doesn't have permission

### DIFF
--- a/spec/controllers/admin/roles_controller_spec.rb
+++ b/spec/controllers/admin/roles_controller_spec.rb
@@ -3,13 +3,13 @@
 require 'rails_helper'
 
 RSpec.describe Api::V1::Admin::RolesController, type: :controller do
-  let(:role) { create(:role, name: 'User') }
-  let(:user) { create(:user, role:) }
+  let(:manage_roles_role) { create(:role, name: 'User') }
+  let(:user) { create(:user, role: manage_roles_role) }
   let(:manage_roles_permission) { create(:permission, name: 'ManageRoles') }
   let!(:manage_roles_role_permission) do
     create(:role_permission,
-           role_id: user.role_id,
-           permission_id: manage_roles_permission.id,
+           role: manage_roles_role,
+           permission: manage_roles_permission,
            value: 'true',
            provider: 'greenlight')
   end
@@ -29,12 +29,6 @@ RSpec.describe Api::V1::Admin::RolesController, type: :controller do
       expect(JSON.parse(response.body)['data'].pluck('id')).to match_array(roles.pluck(:id))
     end
 
-    it 'user without ManageRoles permission cannot return the list of roles' do
-      manage_roles_role_permission.update!(value: 'false')
-      get :index
-      expect(response).to have_http_status(:forbidden)
-    end
-
     it 'returns the roles according to the query' do
       search_roles = [create(:role, name: 'Role 1'), create(:role, name: 'ROLE 2'), create(:role, name: 'role 3')]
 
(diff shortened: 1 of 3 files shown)


spec/controllers/admin/roles_controller_spec.rb
```diff
@@ -20,6 +29,14 @@
       expect(JSON.parse(response.body)['data'].pluck('id')).to match_array(roles.pluck(:id))
     end
 
+    it 'admin without ManageRoles permission cannot return the list of roles' do
+      roles = [create(:role, name: 'Hokage'), create(:role, name: 'Jonin'), create(:role, name: 'Chunin')]
+      roles << user.role
```
scouillard, line 34: I'm not sure this does anything?

spec/controllers/admin/roles_controller_spec.rb
```diff
@@ -20,6 +29,14 @@
       expect(JSON.parse(response.body)['data'].pluck('id')).to match_array(roles.pluck(:id))
     end
 
+    it 'admin without ManageRoles permission cannot return the list of roles' do
```
KH-Amir-TN, line 32: Since GL3 is permission based, there's no meaning behind `admin` role,
I'd change the message to `user without "ManageRoles" ...` instead to avoid giving a meaning to the admin role,

spec/controllers/admin/roles_controller_spec.rb
```diff
@@ -3,7 +3,16 @@
...
-  let(:user) { create(:user, role: create(:role, name: 'User')) }
+  let(:role) { create(:role, name: 'User') }
+  let(:user) { create(:user, role:) }
+  let(:manage_roles_permission) { create(:permission, name: 'ManageRoles') }
+  let!(:manage_roles_role_permission) do
+    create(:role_permission,
+           role_id: user.role_id,
+           permission_id: manage_roles_permission.id,
+           value: 'true',
+           provider: 'greenlight')
```
KH-Amir-TN, line 14:
```suggestion
    create(:role_permission,
           role: user.role,
           permission: manage_roles_permission,
           # if we rename manage_roles_permission -> permission that will further simplify it to `permission:`
           value: 'true',
           provider: 'greenlight')
```

scouillard: We have been using the wording "manage_roles_permission" instead of "permission" because in some cases, we need to add more than one permission to the test suite (ex "manage_users_permissions)

KH-Amir-TN: Gotcha, thank you!

scouillard (approved):

KH-Amir-TN (commented):
If possible, let us try to be consistent with the spec template introduced by @scouillard ,

farhatahmad (approved):

hadicheaito1:
> If possible, let us try to be consistent with the spec template introduced by @scouillard ,

@KH-Amir-TN This is following the same template. The only difference is that it is using the context to remove the permission instead since most of the tests in the file require the permission. We agreed that it what we will do in this case

KH-Amir-TN:
> > If possible, let us try to be consistent with the spec template introduced by @scouillard ,
> 
> @KH-Amir-TN This is following the same template. The only difference is that it is using the context to remove the permission instead since most of the tests in the file require the permission. We agreed that it what we will do in this case

Gotcha,
My bad, I didn't pay enough attention 😅

PR #3752, opened 2022-08-10
MangeRecordings Permission

hadicheaito1:
## Description

- Before action check for ManageRecordings permission where needed
- Moved resync to server_recordings admin
- Server Recordings tab in Admin Panel only shows if current user has ManageRecordings permission

### DIFF
--- a/app/controllers/api/v1/admin/server_recordings_controller.rb
+++ b/app/controllers/api/v1/admin/server_recordings_controller.rb
@@ -4,7 +4,7 @@ module Api
   module V1
     module Admin
       class ServerRecordingsController < ApiController
-        before_action only: %i[index resync] do
+        before_action do
           ensure_authorized('ManageRecordings')
         end
 
--- a/app/controllers/api/v1/admin/server_recordings_controller.rb
+++ b/app/controllers/api/v1/admin/server_recordings_controller.rb
@@ -4,6 +4,10 @@ module Api
   module V1
     module Admin
       class ServerRecordingsController < ApiController
+        before_action only: %i[index resync] do
+          ensure_authorized('ManageRecordings')
+        end
+
         # POST /api/v1/admin/server_recordings.json
         # Expects: {}
         # Returns: { data: Array[serializable objects] , errors: Array[String] }
@@ -15,6 +19,12 @@ module Api
           pagy, recordings = pagy(Recording.order(sort_config)&.search(params[:search]))
 
           render_data data: recordings, meta: pagy_metadata(pagy), status: :ok
+        end
+
+        def resync
+          RecordingsSync.new(user: current_user).call
+
+          render_data status: :ok
         end
       end
     end
--- a/spec/controllers/admin/server_recordings_controller_spec.rb
+++ b/spec/controllers/admin/server_recordings_controller_spec.rb
@@ -10,8 +10,7 @@ RSpec.describe Api::V1::Admin::ServerRecordingsController, type: :controller do
     create(:role_permission,
            role: manage_recordings_role,
            permission: manage_recordings_permission,
-           value: 'true',
-           provider: 'greenlight')
+           value: 'true')
   end
 
   before do
--- a/spec/controllers/admin/server_recordings_controller_spec.rb
+++ b/spec/controllers/admin/server_recordings_controller_spec.rb
@@ -3,13 +3,13 @@
 require 'rails_helper'
 
 RSpec.describe Api::V1::Admin::ServerRecordingsController, type: :controller do
-  let(:role) { create(:role) }
-  let(:user) { create(:user, role:) }
+  let(:manage_recordings_role) { create(:role) }
+  let(:user) { create(:user, role: manage_recordings_role) }
   let(:manage_recordings_permission) { create(:permission, name: 'ManageRecordings') }
   let!(:manage_recordings_role_permission) do
     create(:role_permission,
-           role_id: user.role_id,
-           permission_id: manage_recordings_permission.id,
+           role: manage_recordings_role,
+           permission: manage_recordings_permission,
            value: 'true',
            provider: 'greenlight')
   end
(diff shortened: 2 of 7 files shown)


scouillard (approved):

spec/controllers/admin/server_recordings_controller_spec.rb
```diff
@@ -3,7 +3,16 @@
 require 'rails_helper'
 
 RSpec.describe Api::V1::Admin::ServerRecordingsController, type: :controller do
-  let(:user) { create(:user) }
+  let(:role) { create(:role) }
+  let(:user) { create(:user, role:) }
+  let(:manage_recordings_permission) { create(:permission, name: 'ManageRecordings') }
+  let!(:manage_recordings_role_permission) do
```
KH-Amir-TN, line 9: Please refer to #3749 review,

spec/controllers/admin/server_recordings_controller_spec.rb
```diff
@@ -57,4 +72,22 @@
       end
     end
   end
+
+  describe '#resync' do
+    it 'calls the RecordingsSync service correctly' do
+      expect_any_instance_of(RecordingsSync).to receive(:call)
+      get :resync
```
KH-Amir-TN, line 79: We should always assert our responses in actions tests,

spec/controllers/admin/server_recordings_controller_spec.rb
```diff
@@ -57,4 +72,22 @@
...
+    end
+
+    it 'calls the RecordingsSync service with correct params' do
+      expect(RecordingsSync).to receive(:new).with(user:)
+      get :resync
```
KH-Amir-TN, line 84: Same here, we should assert the response

spec/controllers/admin/server_recordings_controller_spec.rb
```diff
@@ -57,4 +72,22 @@
       end
     end
   end
+
+  describe '#resync' do
+    it 'calls the RecordingsSync service correctly' do
+      expect_any_instance_of(RecordingsSync).to receive(:call)
```
KH-Amir-TN, line 78: I think we may want to assert that the service was initialized with the right parameters and then assert that `:call` was called in the test,

spec/controllers/admin/server_recordings_controller_spec.rb
```diff
@@ -57,4 +72,22 @@
...
+      get :resync
+    end
+
+    it 'calls the RecordingsSync service with correct params' do
+      expect(RecordingsSync).to receive(:new).with(user:)
```
KH-Amir-TN, line 83: After asserting the response this will fail, we should either call the original `:new` method implantation or stub the the `:new` method returning an instance double (verified double) when we don't want to trigger the original implentation,

spec/controllers/recordings_controller_spec.rb
```diff
@@ -91,6 +100,14 @@
       expect(response).to have_http_status(:ok)
     end
 
+    it 'admin without ManageRecordings permission cannot update the recordings name' do
+      manage_recordings_role_permission.update!(value: 'false')
+      expect { post :update, params: { recording: { name: 'My Awesome Recording!' }, id: recording.record_id } }.not_to(change do
```
KH-Amir-TN, line 105: I think that we can simplify this line a bit?

spec/controllers/recordings_controller_spec.rb
```diff
@@ -115,26 +132,18 @@
   #     recording = create(:recording)
   #     expect { delete :destroy, params: { id: recording.id } }.to change(Recording, :count).by(-1)
   #   end
-
+  #   it 'admin without ManageRecordings permission cannot delete recording from the database' do
```
KH-Amir-TN, line 135: I know that this irrelevant to this PR,
But, what's the reasoning behind these test group been commented?

KH-Amir-TN (commented):
This PR is addressing more then a single concern,
I'd suggest that we extract the migration of the `re-sync`  API into its separate PR if possible,
We still have some work to do on the `re-sync` feature so it will have more PRs in the upcoming weeks,

spec/controllers/recordings_controller_spec.rb
hadicheaito1: It was to do with the `BigBlueButtonApi.new.delete_recordings(record_ids: params[:id])` having to be moved out of the destroy method and be put in the Recording model in an `after_destroy`.

I think either way we would need to just mock the response of `BigBlueButtonApi.new.delete_recordings(record_ids: params[:id])`? @KH-Amir-TN

spec/controllers/admin/server_recordings_controller_spec.rb
hadicheaito1: I think all this stuff should be done in a different PR. I only moved the resync method since it affected the ManageRecordings permission

Could you explain this one more though, I don't understand why it's failing

spec/controllers/recordings_controller_spec.rb
KH-Amir-TN: That makes sense,
In such case IMO we'd like to have the `#destroy` tests that like any other API will assert the removal of recordings records in the DB while expecting certain responses for certain cases,
The action tests shouldn't make sense of the model `callbacks` but to avoid compromising the rigidity of the tests we'd stub the  `BigBlueButtonApi#delete_recordings` like you said,

app/controllers/api/v1/admin/server_recordings_controller.rb
```diff
@@ -4,6 +4,10 @@ module Api
   module V1
     module Admin
       class ServerRecordingsController < ApiController
+        before_action only: %i[index resync] do
```
KH-Amir-TN, line 7: Conceptually, admin controllers if found should concern one area of GL3 admin panel and `ensure_authorized(...)` for all actions by default,

KH-Amir-TN (commented):
For specs if possible let's try to be consistent with the permission template,

farhatahmad:
Rebate on latest so that this passes

hadicheaito1:
> Rebase on latest so that this passes

@farhatahmad Done

farhatahmad (approved):

PR #3756, opened 2022-08-11
Remove admin::user#create endpoint duplication

scouillard:
## Description

Remove the admin::user#create endpoint.
Put the logic inside the user#create controller.

Small fix: the logic added to user#create controller is to make sure the current_user is not replaced by the created user if the user creation is made via a logged in user.

Adding #create to the `before_action ensure_authorized` creates a problem with the PermissionsChecker because the service checks if you are either a current_user or have a permission (and a new user is/have neither).

But, ultimately, you don't need a special permission to create an account, so I think it's not needed anyways - correct me if I am wrong.

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -30,7 +30,7 @@ module Api
         return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
         if user.save
-          session[:user_id] = user.id
+          session[:user_id] ||= user.id
           token = user.generate_activation_token!
           render_data data: { token: }, status: :created # TODO: enable activation email sending.
         else
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -23,16 +23,16 @@ RSpec.describe Api::V1::UsersController, type: :controller do
       }
     end
 
-    it 'creates a current user if the new user is created' do
+    it 'creates a current_user if a new user is created' do
       session[:user_id] = nil
       create(:role, name: 'User') # Needed for admin#create
-      post :create, params: user_params
-      expect(session[:user_id]).to eql(User.last.id)
+      expect { post :create, params: user_params }.to change(User, :count).by(1)
+      expect(session[:user_id]).to be_present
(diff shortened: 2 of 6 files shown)


app/controllers/api/v1/users_controller.rb
```diff
@@ -30,7 +30,7 @@ def create
         return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
         if user.save
-          session[:user_id] = user.id
+          session[:user_id] ||= user.id
```
KH-Amir-TN, line 33: Good catch!

spec/controllers/users_controller_spec.rb
```diff
@@ -10,6 +10,33 @@
...
+          language: 'language'
+        }
+      }
+    end
+
+    it 'creates a current user if the new user is created' do
+      session[:user_id] = nil
+      create(:role, name: 'User') # Needed for admin#create
+      post :create, params: user_params
+      expect(session[:user_id]).to eql(User.last.id)
```
KH-Amir-TN, line 30: Um, this seems to be okay but it somehow makes this test fragile since anything that can cause `ActiveRecord` to return the `user` record as the last in the `User` table will cause this to fail,

I'd go with something much safer maybe like:
```
 expect(session[:user_id]).to be_present
 expect(session[:user_id]).not_to (user.id)
```

spec/controllers/users_controller_spec.rb
```diff
@@ -10,6 +10,33 @@
...
+    end
+
+    it 'creates a new user without changing the current user if the new user is created from a logged in user' do
```
KH-Amir-TN, line 33: The tests description is setting that we're creating of a new user but we're not asserting that,

KH-Amir-TN (commented):
I think we need to update `routes.rb`

KH-Amir-TN:
> But, ultimately, you don't need a special permission to create an account, so I think it's not needed anyways - correct me if I am wrong.
For creating a user account **only** I think that it doesn't require any permission.
The feature itself is only for GL administrators convenience it helps them creating accounts without leaving the administrator panel,

KH-Amir-TN (approved):

PR #3757, opened 2022-08-11
Remove admin::user#destroy endpoint duplication + "admin" specs template proposal

scouillard:
## Description

Remove the duplicated user#destroy endpoint.

**Propose a new template for "admin" specs:**
In specs, it can be difficult to differentiate regular user specs from "admin" users specs.
1. The roles/permissions are set on top of the resource spec files (this can be moved in another module eventually)

```
  let(:<permission>_role) { create(:role) }
  let(:<permission>_permission) { create(:permission, name: '<permission>') }
  let(:<permission>_role_permission) do
    create(:role_permission,
           role: <permission>_role,
           permission: <permission>_permission,
           value: 'true',
           provider: 'greenlight')
  end
```

2. A context is created inside the `describe <endpoint>`

```
describe 'some#endpoint' do
  it 'does something' do 
    test something
  end

  context 'user with <permission> is doing something' do
     before do 
       <permission>_role_permission ## instantiate the role_permission
       user.update!(role: <permission>_role) ## assign current_user the permission
     end

     it 'does something' do 
       test something
     end 
end
```

### DIFF
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -105,13 +105,8 @@ RSpec.describe Api::V1::UsersController, type: :controller do
     end
   end
 
-<<<<<<< HEAD
-  describe 'DELETE users#destroy' do
-    it 'deletes the user' do
-=======
   describe 'users#destroy' do
     it 'deletes the current_user account' do
->>>>>>> e0266e2d... Fix RSpec + Routes
       expect(response).to have_http_status(:ok)
       expect { delete :destroy, params: { id: user.id } }.to change(User, :count).by(-1)
     end
@@ -124,13 +119,10 @@ RSpec.describe Api::V1::UsersController, type: :controller do
 
     it 'returns status code forbidden if the user id is invalid' do
       expect { delete :destroy, params: { id: 'invalid-id' } }.not_to change(User, :count)
-<<<<<<< HEAD
-      expect(response).to have_http_status(:not_found)
-=======
       expect(response).to have_http_status(:forbidden)
     end
 
-    context 'current_user with ManageUsers permission deletes another user' do
+    context 'current_user with ManageUsers permission' do
       before do
         manage_users_role_permission
         user.update!(role: manage_users_role)
@@ -145,7 +137,6 @@ RSpec.describe Api::V1::UsersController, type: :controller do
         expect { delete :destroy, params: { id: 'invalid-id' } }.not_to change(User, :count)
         expect(response).to have_http_status(:not_found)
       end
->>>>>>> e0266e2d... Fix RSpec + Routes
     end
   end
 
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -7,7 +7,7 @@ RSpec.describe Api::V1::UsersController, type: :controller do
 
   let(:manage_users_role) { create(:role) }
   let(:manage_users_permission) { create(:permission, name: 'ManageUsers') }
-  let(:manage_users_role_permission) do
+  let!(:manage_users_role_permission) do
     create(:role_permission,
            role: manage_users_role,
            permission: manage_users_permission,
@@ -124,7 +124,6 @@ RSpec.describe Api::V1::UsersController, type: :controller do
 
     context 'current_user with ManageUsers permission' do
       before do
-        manage_users_role_permission
         user.update!(role: manage_users_role)
       end
 
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -117,11 +117,6 @@ RSpec.describe Api::V1::UsersController, type: :controller do
       expect(response).to have_http_status(:forbidden)
     end
 
-    it 'returns status code forbidden if the user id is invalid' do
-      expect { delete :destroy, params: { id: 'invalid-id' } }.not_to change(User, :count)
-      expect(response).to have_http_status(:forbidden)
-    end
-
     context 'current_user with ManageUsers permission' do
       before do
         user.update!(role: manage_users_role)
@@ -132,7 +127,7 @@ RSpec.describe Api::V1::UsersController, type: :controller do
         expect { delete :destroy, params: { id: new_user.id } }.to change(User, :count).by(-1)
       end
 
-      it 'returns status code not found if the user id is invalid' do
+      it 'returns status code not found if the user does not exists' do
         expect { delete :destroy, params: { id: 'invalid-id' } }.not_to change(User, :count)
         expect(response).to have_http_status(:not_found)
       end
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -5,6 +5,16 @@ require 'rails_helper'
 RSpec.describe Api::V1::UsersController, type: :controller do
   let(:user) { create(:user) }
 
+  let(:manage_users_role) { create(:role) }
+  let(:manage_users_permission) { create(:permission, name: 'ManageUsers') }
+  let(:manage_users_role_permission) do
+    create(:role_permission,
+           role: manage_users_role,
+           permission: manage_users_permission,
+           value: 'true',
+           provider: 'greenlight')
+  end
+
   before do
     request.headers['ACCEPT'] = 'application/json'
     session[:user_id] = user.id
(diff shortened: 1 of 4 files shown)


spec/controllers/users_controller_spec.rb
```diff
@@ -4,6 +4,15 @@
 
 RSpec.describe Api::V1::UsersController, type: :controller do
   let(:user) { create(:user) }
+  let(:admin_role) { create(:role) }
```
KH-Amir-TN, line 7: Can we call this something other than admin?
For example `users_manager_role`?

spec/controllers/users_controller_spec.rb
```diff
@@ -68,15 +77,31 @@
     end
   end
 
-  describe 'DELETE users#destroy' do
+  describe 'users#destroy' do
     it 'deletes the user' do
```
KH-Amir-TN, line 81: I think that `deletes the current_user account` is more informational,

spec/controllers/users_controller_spec.rb
```diff
@@ -68,15 +77,31 @@
...
       expect(response).to have_http_status(:ok)
       expect { delete :destroy, params: { id: user.id } }.to change(User, :count).by(-1)
     end
 
     it 'does not delete any user if the user id is invalid' do
```
KH-Amir-TN, line 86: I think that this should be scoped inside of a test group like `context 'current_user has NOT 'ManageUsers' permission'`,
And maybe we can add two tests one for `it 'returns status code :forbidden for a NONE current_user :id '` targeting an existing user and another `it 'returns status code :forbidden for inexistent :id '` targeting an inexistent user,

KH-Amir-TN (commented):
Shouldn't we update the `routes.rb`?

spec/controllers/users_controller_spec.rb
scouillard: IMO qualifying an admin as someone with ALL the permissions makes sense.

Then, in negative specs, if you want to test something if the admin does not have a certain permission, you remove it manually.

I've been refactoring those for the past two days and IMO this is the cleaner way to do this.

scouillard: I have yet to see a case where we need to discriminate between the admin permissions, for example a case where we need to test if an admin has a certain permission but does not have a certain other.

spec/controllers/users_controller_spec.rb
```diff
@@ -68,15 +77,31 @@
...
       expect { delete :destroy, params: { id: 'invalid-id' } }.not_to change(User, :count)
-      expect(response).to have_http_status(:not_found)
+      expect(response).to have_http_status(:forbidden)
+    end
+
+    context 'current_user is an admin' do
```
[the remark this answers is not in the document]
scouillard: It gets really complicated and messy - unless you find a better solution than I did.

Please look at the whole file.

We create an admin_role at the head of the file.

And we link all the permissions to the admin.

```
  let(:manage_rooms_permission) { create(:permission, name: 'ManageRooms') }
  let!(:manage_rooms_role_permission) do
    create(:role_permission,
           role_id: admin_role.id,
           permission_id: manage_rooms_permission.id,
           value: 'true',
           provider: 'greenlight')
  end

  let(:manage_users_permission) { create(:permission, name: 'ManageUsers') }
  let!(:manage_users_role_permission) do
    create(:role_permission,
           role_id: admin_role.id,
           permission_id: manage_users_permission.id,
           value: 'true',
           provider: 'greenlight')
  end
```

And, in the admin context **before do**, we simply have to set the current_user as an admin with `user.update!(role_id: admin_role.id)`.

Again, after spending two days on this, it's the cleanest way I've found.

**I have yet to see a case where we need to discriminate between the admin permissions, for example a case where we need to test if an admin has a certain permission but does not have a certain other.**

In other words, the user/admin might as well have ALL the permissions - it doesn't change anything.

If needed, you can always manually remove a permission.

hadicheaito1:
The only issue that I see now is that some test files have the case where most of their tests require a permission (such as `roles_controller_spec`). In this case I find it cleaner in a way to just give the user the permission at the beginning and then disable it where needed.

I feel your proposal works best in the cases where majority of the tests in the file don't need the permission being tested.

^ That being said, I'm not sure what the best approach would be atm, but just thought I should share my thinking

scouillard:
> The only issue that I see now is that some test files have the case where most of their tests require a permission (such as `roles_controller_spec`). In this case I find it cleaner in a way to just give the user the permission at the beginning and then disable it where needed.
> 
> I feel your proposal works best in the cases where majority of the tests in the file don't need the permission being tested.
> 
> ^ That being said, I'm not sure what the best approach would be atm, but just thought I should share my thinking

Absolutely - in those cases, you would do the opposite.

In most Specs tho I think this template will work great.

scouillard:
It was obvious after two days of trying to refactor the Specs that something needed to be done for the user "permissions" specs.
I like working with this template, but of course this solution can be improved or even replaced.

spec/controllers/users_controller_spec.rb
```diff
@@ -68,15 +78,38 @@
...
+      expect(response).to have_http_status(:forbidden)
+    end
+
+    it 'returns status code forbidden if the user id is invalid' do
       expect { delete :destroy, params: { id: 'invalid-id' } }.not_to change(User, :count)
-      expect(response).to have_http_status(:not_found)
+      expect(response).to have_http_status(:forbidden)
+    end
+
+    context 'current_user with ManageUsers permission deletes another user' do
```
KH-Amir-TN, line 98: This can be further simplified to something like "current_user with "ManageUsers" permission",

spec/controllers/users_controller_spec.rb
```diff
@@ -68,15 +78,38 @@
...
     end
 
-    it 'does not delete any user if the user id is invalid' do
+    it 'returns status code forbidden if the user tries to delete another user' do
+      new_user = create(:user)
+      expect { delete :destroy, params: { id: new_user.id } }.not_to change(User, :count)
+      expect(response).to have_http_status(:forbidden)
+    end
+
+    it 'returns status code forbidden if the user id is invalid' do
```
KH-Amir-TN, line 93: The word "invalid" may be misleading since we use it to indicate invalid params,
Do you find this more informational "returns status code :forbidden for inexistent :id"?

spec/controllers/users_controller_spec.rb
```diff
@@ -68,15 +78,38 @@
...
       expect { delete :destroy, params: { id: 'invalid-id' } }.not_to change(User, :count)
-      expect(response).to have_http_status(:not_found)
+      expect(response).to have_http_status(:forbidden)
+    end
+
+    context 'current_user with ManageUsers permission deletes another user' do
+      before do
+        manage_users_role_permission
```
KH-Amir-TN, line 100: Is this line required?

scouillard: Yes, it's needed to instantiate the object.
I would use let! instead but RSpec is complaining that the object is not being directly referenced - maybe we should remove that rule?

hadicheaito1: Yeah I agree that rule is annoying

spec/controllers/users_controller_spec.rb
```diff
@@ -95,15 +105,47 @@
     end
   end
 
+<<<<<<< HEAD
```
hadicheaito1, line 108: These needs to be removed

spec/controllers/users_controller_spec.rb
scouillard: It is kind of invalid, no?
It's a string and it's expecting an id.
`params: { id: 'invalid-id' }`

farhatahmad (dismissed):

spec/controllers/users_controller_spec.rb
farhatahmad: Yeah I would remove it and use `let!`

spec/controllers/users_controller_spec.rb
KH-Amir-TN: Aren't we testing here the case when we're trying to remove a user that does not exist?
This gives the impression that our API is asserting the data type of the inputs our their values considering some as invalid and returning an errored response.

spec/controllers/users_controller_spec.rb
```diff
@@ -95,15 +105,37 @@
...
+      before do
+        user.update!(role: manage_users_role)
+      end
+
+      it 'deletes a user' do
+        new_user = create(:user)
+        expect { delete :destroy, params: { id: new_user.id } }.to change(User, :count).by(-1)
+      end
+
+      it 'returns status code not found if the user id is invalid' do
```
KH-Amir-TN, line 135: I think this will clearify what I'm trying to say,
This is testing a `:not_found` case, an inexistant id isn't invalid.

scouillard: Right. Agreed. I don't find this test particularly useful neither.

scouillard: I changed it to `'returns status code not found if the user does not exists'`

farhatahmad (approved):

PR #3760, opened 2022-08-12
Importing Pagination css file only

hadicheaito1:
## Description

- Importing only the css file that is needed for the pagination component from Semantic UI intead of importing all the css files


KH-Amir-TN (approved):
Cool, now it's gonna take 5 seconds only to load 🤣

scouillard (approved):

PR #3762, opened 2022-08-15
Access Codes: Added Access codes auto generation on certain configs [OPTIMIZED].

KH-Amir-TN:
## Description

This PR:

- Edits `Rooms` model to add in the capability to auto generate access codes on rooms creation and first interaction after having forced enabled configs.

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -8,6 +8,13 @@ class Room < ApplicationRecord
 
   has_many :recordings, dependent: :destroy
   has_many :room_meeting_options, dependent: :destroy
+  has_one :viewer_access_code_setting, lambda {
+                                         joins(:meeting_option)
+                                           .where(meeting_option: { name: 'glViewerAccessCode' })
+                                       }, class_name: 'RoomMeetingOption', inverse_of: :room, dependent: :destroy
+  has_one :moderator_access_code_setting, lambda {
+                                            joins(:meeting_option).where(meeting_option: { name: 'glModeratorAccessCode' })
+                                          }, class_name: 'RoomMeetingOption', inverse_of: :room, dependent: :destroy
 
   has_one_attached :presentation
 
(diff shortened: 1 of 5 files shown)


app/models/room.rb
```diff
@@ -82,12 +92,28 @@ def set_meeting_id
 
   # Autocreate all meeting options using the default values
   def create_meeting_options
-    MeetingOption.all.find_each do |option|
-      RoomMeetingOption.create(room: self, meeting_option: option, value: option.default_value)
+    access_codes_names = %w[glViewerAccessCode ModeratorAccessCode]
+    access_configs = MeetingOption.access_codes_configs(provider: 'greenlight').to_h
+
+    MeetingOption.find_each do |option|
+      value = if access_codes_names.include?(option.name) && access_configs[option.name] == 'true'
```
KH-Amir-TN, line 99: `access_codes_names.include ?(...)` check was added as a safe guard in case the `MeetingOption.access_codes_configs` returns some irrelevant settings because of an intended update to its implementation ensuring the determinism of this action,

app/models/room.rb
```diff
@@ -82,12 +92,28 @@ def set_meeting_id
...
   end
 
   def generate_code
     SecureRandom.alphanumeric(6).downcase
   end
+
+  def auto_generate_access_codes
+    access_configs = MeetingOption.access_codes_configs(provider: 'greenlight').to_h
+
+    generate_viewer_access_code if access_configs['glViewerAccessCode'] == 'true' && viewer_access_code.blank?
```
KH-Amir-TN, line 116: The order ensures that access codes records will be lazy loaded,
A possible optimization is to invert the flow of this logic eager loading the access codes (because they are more likely to be accessed) and only lazy load the configs when needed (if the access code is `:blank?` then we check if the config value is force enabled).

I've chosen this flow because it's a bit simpler but the other still worth mentioning for future refactoring targeting optimization,

KH-Amir-TN:
# REBASED ON e35a8820c41bdf1560c8e65438e8ccae4c415bc2

PR #3765, opened 2022-08-16
Add named arguments to the ensure_authorized method

scouillard:
## Description

Add user_id, friendly_id, and record_id as named argument for the `ensure_authorized` method.
This change lets us pass the <user resource> param id into the `before_action` callback.

### DIFF
--- a/app/controllers/concerns/authorizable.rb
+++ b/app/controllers/concerns/authorizable.rb
@@ -14,14 +14,14 @@ module Authorizable
   end
 
   # PermissionsChecker service will return a true or false depending on whether the current_user's role has the provided permission_name
-  def ensure_authorized(permission_name)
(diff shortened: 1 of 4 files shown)


app/controllers/concerns/authorizable.rb
```diff
@@ -14,14 +14,14 @@ def ensure_authenticated
   end
 
   # PermissionsChecker service will return a true or false depending on whether the current_user's role has the provided permission_name
-  def ensure_authorized(permission_name)
+  def ensure_authorized(permission_name, user_id: '', friendly_id: '', record_id: '')
```
farhatahmad, line 17: I would default these to `nil` not `''`

farhatahmad (approved):

PR #3766, opened 2022-08-16
SharedList Permission

hadicheaito1:
## Description

- Removes Users with SharedList permission == false
  - Gets all the Role ids of the roles that have permission SharedList == true and then only gets the users with role_ids in that list

### DIFF
--- a/spec/controllers/shared_accesses_controller_spec.rb
+++ b/spec/controllers/shared_accesses_controller_spec.rb
@@ -4,6 +4,8 @@ require 'rails_helper'
 
 RSpec.describe Api::V1::SharedAccessesController, type: :controller do
   let(:user) { create(:user) }
+  let(:shared_list_role) { create(:role) }
+  let(:shared_list_permission) { create(:permission, name: 'SharedList') }
 
   before do
     request.headers['ACCEPT'] = 'application/json'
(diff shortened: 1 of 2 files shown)


scouillard:
Would it be faster to have a negative logic instead ie fetch the users without the SharedList permission?
As the "default" for SharedList is true.

spec/controllers/shared_accesses_controller_spec.rb
```diff
@@ -55,27 +57,47 @@
...
-      room.shared_users = create_list(:user, 5)
-      shareable_users = create_list(:user, 5, name: 'Jane Doe')
+    context 'users are given SharedList permission' do
+      before do
+        user.update!(role: shared_list_role)
+        create(:role_permission,
+               role: shared_list_role,
+               permission: shared_list_permission,
+               value: 'true',
+               provider: 'greenlight')
```
scouillard, line 78: Could we put this on top of the file like the others specs please

hadicheaito1: It's a bit different since it isn't being assigned to a variable in this file. That's why I didn't put it at the top like the rest of the files, rubocop complains. I could put it in the `before do` at the very top but I felt it would be better to put it in this more specific `before do`

farhatahmad (approved):

PR #3767, opened 2022-08-16
Big refactor related to roles provider

farhatahmad:
## Description

1- Moved the `provider` attribute to roles instead of role permissions (since that makes more sense)
2- Cleaned up tests by hardcoding the provider for ease of us 
3- Fixed annoying warnings when running tests

### DIFF
--- a/spec/models/role_spec.rb
+++ b/spec/models/role_spec.rb
@@ -6,36 +6,38 @@ RSpec.describe Role, type: :model do
   describe 'validations' do
     subject { create(:role) }
 
+    it { is_expected.to validate_presence_of(:provider) }
+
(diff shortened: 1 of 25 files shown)


spec/models/role_spec.rb
```diff
@@ -6,36 +6,38 @@
...
-          expect(role).to be_valid
-          expect(role.color).to eq('#ffffff')
-        end
+      it 'overwrites any color value when provided on :create' do
+        allow(SecureRandom).to receive(:hex).and_return('ffffff')
+        role = build(:role, color: '#HEXTET')
+        expect(role).to be_valid
+        expect(role.color).to eq('#ffffff')
       end
     end
```
farhatahmad, line 42: This is all just reordering

scouillard (approved):

PR #3768, opened 2022-08-16
Remove admin::server_rooms#destroy endpoint duplication

scouillard:
## Description

Remove admin::server_rooms#destroy endpoint duplication.
Rework rooms#destroy endpoint, specs.

## Testing Steps

- Delete one of your room
- Delete someone else's room as an admin
- (try to) delete someone else's room as a regular user

### DIFF
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -4,16 +4,18 @@ module Api
   module V1
     class RoomsController < ApiController
       before_action :find_room,
-                    only: %i[show update recordings recordings_processing
+                    only: %i[show update destroy recordings recordings_processing
                              purge_presentation access_codes
                              generate_access_code remove_access_code]
 
       skip_before_action :ensure_authenticated, only: %i[show]
-      before_action only: %i[show] do
-        ensure_authorized('ManageRooms', friendly_id: params[:friendly_id])
-      end
+
       before_action only: %i[create] do
         ensure_authorized('ManageUsers', user_id: room_params[:user_id])
+      end
+
+      before_action only: %i[show destroy] do
+        ensure_authorized('ManageRooms', friendly_id: params[:friendly_id])
       end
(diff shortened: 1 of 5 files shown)


farhatahmad (dismissed):

farhatahmad:
@scouillard needs a rebase

spec/controllers/rooms_controller_spec.rb
```diff
@@ -106,6 +106,31 @@
...
+    it 'cannot delete the room of another user' do
+      new_user = create(:user)
+      room = create(:room, user: new_user)
+      expect { delete :destroy, params: { friendly_id: room.friendly_id } }.not_to change(Room, :count)
+      expect(response).to have_http_status(:forbidden)
+    end
+
+    context 'user with ManageRooms permission' do
+      before do
+        user.update!(role_id: manage_rooms_role.id)
```
KH-Amir-TN, line 119: We can do `role: manage_rooms_role`

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -4,18 +4,20 @@ module Api
   module V1
     class RoomsController < ApiController
       before_action :find_room,
-                    only: %i[show update recordings recordings_processing
+                    only: %i[show update destroy recordings recordings_processing
```
KH-Amir-TN, line 7: I'm starting to have a hard time reading this,
We can invert the logic flow by using `except` instead of `only` 
```
      before_action :find_room, except: %i[index create]
```

KH-Amir-TN (commented):
Shouldn't we update the `routes.rb`?

farhatahmad (approved):

app/controllers/api/v1/rooms_controller.rb
farhatahmad: I prefer that we manually specify rather than invert and assume for all methods

PR #3769, opened 2022-08-16
Room settings: Upgraded `RoomSettingsController#update` API.

KH-Amir-TN:
+ Handle access codes generation and removal.

### DIFF
--- a/app/controllers/api/v1/room_settings_controller.rb
+++ b/app/controllers/api/v1/room_settings_controller.rb
@@ -15,12 +15,20 @@ module Api
       # PATCH /api/v1/room_settings/:friendly_id
       def update
         config = MeetingOption.get_config_value(name: room_setting_params[:settingName], provider: current_provider)&.value
+        return render_error status: :bad_request unless config
 
-        return render_error status: :forbidden unless config == 'optional'
+        name = room_setting_params[:settingName]
+        value = room_setting_params[:settingValue].to_s
 
-        option = @room.get_setting(name: room_setting_params[:settingName])
+        is_access_code = %w[glViewerAccessCode glModeratorAccessCode].include? name
 
-        return render_error status: :bad_request unless option&.update(value: room_setting_params[:settingValue].to_s)
+        return render_error status: :forbidden unless config == 'optional' || (config == 'true' && is_access_code && value != 'false')
+
+        value = infer_access_code(value:) if is_access_code # Handling access code update.
+
+        option = @room.get_setting(name:)
+
+        return render_error status: :bad_request unless option&.update(value:)
 
         render_data status: :ok
       end
@@ -33,6 +41,16 @@ module Api
 
       def room_setting_params
         params.require(:room_setting).permit(:settingValue, :settingName)
+      end
+
+      def infer_access_code(value:)
+        # Access code update logic maps 'false' -> removing a code & <OTHER> -> generating a code.
+        value == 'false' ? '' : generate_code
+      end
+
+      # TODO: Check if we could extract all GL3 codes generation into a service.
+      def generate_code
+        SecureRandom.alphanumeric(6).downcase
       end
     end
   end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -76,8 +76,8 @@ RSpec/AnyInstance:
   Enabled: false
 
 Metrics/CyclomaticComplexity:
-  Max: 8
+  Max: 10
 
 Metrics/PerceivedComplexity:
-  Max: 9
+  Max: 10
 
--- a/spec/controllers/room_settings_controller_spec.rb
+++ b/spec/controllers/room_settings_controller_spec.rb
@@ -50,6 +50,61 @@ RSpec.describe Api::V1::RoomSettingsController, type: :controller do
       expect(room.room_meeting_options.take.value).to eq('notOptionalAnymore')
     end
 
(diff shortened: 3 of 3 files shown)


KH-Amir-TN:
# REBASED ON 99628ede8921f9ef83361c86ff523ef89de4fb78

app/controllers/api/v1/room_settings_controller.rb
```diff
@@ -15,12 +15,36 @@ def show
...
-        return render_error status: :bad_request unless option&.update(value: room_setting_params[:settingValue].to_s)
+                    generate_code
+                  when 'false'
+                    # Removing access code.
+                    return render_error status: :forbidden unless config == 'optional' # Removing access code is only allowed when it's optional.
+
+                    ''
+                  else
+                    return render_error status: :bad_request # Only explicit stringified '"true"|"false"' values are allowed.
+                  end
```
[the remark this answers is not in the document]
farhatahmad: Maybe create a private method called `access_code_value`?

```
option = @room.get_setting(name: room_setting_params[:settingName])

value = access_code_value(room_setting_params[:settingValue])

return render_error status: :bad_request unless option&.update(value:)
```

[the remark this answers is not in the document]
farhatahmad: Also, the additional logic in the case statement is overkill. Just return forbidden if the config isn't optional, no need for all the specific cases

app/controllers/api/v1/room_settings_controller.rb
```diff
@@ -15,12 +15,20 @@ def show
...
 
-        return render_error status: :forbidden unless config == 'optional'
+        name = room_setting_params[:settingName]
+        value = room_setting_params[:settingValue].to_s
 
-        option = @room.get_setting(name: room_setting_params[:settingName])
+        is_access_code = %w[glViewerAccessCode glModeratorAccessCode].include? name
 
-        return render_error status: :bad_request unless option&.update(value: room_setting_params[:settingValue].to_s)
+        return render_error status: :forbidden if config == 'false' || (config == 'true' && (!is_access_code || value == 'false'))
```
KH-Amir-TN, line 25: This may be hard to fallow, it's a result of a mathematical simplification (Boolean Algebra) of the fallowing expression:
```
if (is_access_code && value != 'false' && config == 'false') || (is_access_code && value == 'false' && config != 'optional') || (!is_access_code && config != 'optional')
  return render_error status: :forbidden # Re-generting access code is only forbidden when it's disabled.
end
```

First simplification:
```
 if (is_access_code && value != 'false' && config == 'false') || (config != 'optional' && (!is_access_code || value == 'false'))
   return render_error status: :forbidden # Re-generting access code is only forbidden when it's disabled.
end
```
...

Simplified version:
```
if config == 'false' || (config == 'true' && (!is_access_code || value == 'false'))
  return render_error status: :forbidden # Re-generting access code is only forbidden when it's disabled.
end
```

farhatahmad: Couldn't you just do `if config != 'optional' || !(is_access_code && config == 'true' && value != 'false')

app/controllers/api/v1/room_settings_controller.rb
```diff
@@ -34,6 +42,21 @@ def find_room
...
       end
+
+      def infer_access_code(value:)
+        # Access code update logic maps 'false' -> removing a code & <OTHER> -> generating a code.
+        case value
+        when 'false'
+          '' # Removing access code.
+        else
+          generate_code # By default, authorized req will (Re)generate access code.
+        end
```
farhatahmad, line 53: no need for the `case` anymore

```
def infer_access_code(value:)
  return '' if value == 'false'
  
  generate_code
end

```

KH-Amir-TN: You're right,
Do you like further simplifying it into a one line with:
```
def infer_access_code(value:)
  value == 'false' ? '' : generate_code
end
```

farhatahmad: Yeah that works

app/controllers/api/v1/room_settings_controller.rb
KH-Amir-TN: I believe `if config != 'optional' || !(is_access_code && config == 'true' && value != 'false')` is the same as:
`if config != 'optional' || !is_access_code || config != 'true' || value == 'false'`.
Unfortunately, this is not the case.
This fails for cases like the option isn't an access code and it's optional, an optional forced enabled access code,...
Plus it's not order independent,

KH-Amir-TN: I may be wrong, will further discuss it in our meeting today

farhatahmad: How about this? This way we make it clear the 2 conditions to get through are either optional config or force enabled access code

`if !(config == optional || (config == true && is_access_code && value == true))`

KH-Amir-TN: Yup, that's 🔥 
It's a lot more readable,
I'm just make this minor changes because we're mapping only `value=='false'` explicitly,
`unless config == 'optional' || (config == 'true' && is_access_code && value != 'false')`

farhatahmad (approved):

PR #3770, opened 2022-08-16
Fix Branding Image

scouillard:
## Description

Finish the Branding Image feature
Create shared_component/Logo component, set max-height and max-width to Logo

### DIFF
--- a/app/controllers/api/v1/admin/site_settings_controller.rb
+++ b/app/controllers/api/v1/admin/site_settings_controller.rb
@@ -26,7 +26,6 @@ module Api
 
           update = if params[:name] == 'BrandingImage'
                      site_setting.image.attach params[:site_setting][:value]
-                     site_setting.update(value: site_setting.image.blob.filename.to_s)
                    else
                      site_setting.update(value: params[:site_setting][:value].to_s)
                    end
--- a/app/controllers/api/v1/admin/site_settings_controller.rb
+++ b/app/controllers/api/v1/admin/site_settings_controller.rb
@@ -26,6 +26,7 @@ module Api
 
           update = if params[:name] == 'BrandingImage'
                      site_setting.image.attach params[:site_setting][:value]
+                     site_setting.update(value: site_setting.image.blob.filename.to_s)
                    else
                      site_setting.update(value: params[:site_setting][:value].to_s)
                    end
@@ -33,6 +34,24 @@ module Api
           return render_error status: :bad_request unless update
 
           render_data status: :ok
+        end
+
+        def destroy
+          site_setting = SiteSetting.joins(:setting)
+                                    .find_by(
+                                      provider: current_provider,
+                                      setting: { name: params[:name] }
+                                    )
+
+          return render_error status: :not_found unless site_setting
+
+          site_setting.image.purge if params[:name] == 'BrandingImage'
+
+          if site_setting.update(value: '')
+            render_data status: :ok
+          else
+            render_error status: :bad_request
+          end
         end
       end
     end
--- a/app/controllers/api/v1/admin/site_settings_controller.rb
+++ b/app/controllers/api/v1/admin/site_settings_controller.rb
@@ -35,24 +35,6 @@ module Api
 
           render_data status: :ok
         end
-
-        def destroy
-          site_setting = SiteSetting.joins(:setting)
-                                    .find_by(
-                                      provider: current_provider,
-                                      setting: { name: params[:name] }
-                                    )
-
-          return render_error status: :not_found unless site_setting
-
-          site_setting.image.purge if params[:name] == 'BrandingImage'
-
-          if site_setting.update(value: '')
-            render_data status: :ok
-          else
-            render_error status: :bad_request
-          end
-        end
       end
     end
   end
--- a/app/javascript/components/admin/site_settings/appearance/BrandingImage.jsx
+++ b/app/javascript/components/admin/site_settings/appearance/BrandingImage.jsx
@@ -3,6 +3,7 @@ import { Card } from 'react-bootstrap';
 import { CloudUploadIcon } from '@heroicons/react/outline';
 import useUpdateSiteSetting from '../../../../hooks/mutations/admin/site_settings/useUpdateSiteSetting';
 import FilesDragAndDrop from '../../../shared_components/utilities/FilesDragAndDrop';
+import FormControl from '../../../shared_components/forms/FormControl';
 
 export default function BrandingImage() {
   const updateSiteSetting = useUpdateSiteSetting('BrandingImage');
@@ -26,7 +27,7 @@ export default function BrandingImage() {
                 className="d-none"
                 type="file"
                 onChange={(e) => updateSiteSetting.mutate(e.target.files[0])}
-                accept="image/*"
+                accept=".png,.svg,.jpg"
               />
               <Card.Title className="text-brand">
                 Click to Upload <span className="fs-5 fw-normal text-muted"> or drag and drop </span>
--- a/app/javascript/components/admin/site_settings/appearance/BrandingImage.jsx
+++ b/app/javascript/components/admin/site_settings/appearance/BrandingImage.jsx
@@ -8,11 +8,14 @@ export default function BrandingImage() {
   const updateSiteSetting = useUpdateSiteSetting('BrandingImage');
 
   return (
-    <div className="mt-4">
-      <h6> Branding Image </h6>
-
-      <FilesDragAndDrop numOfFiles={1} onDrop={(files) => updateSiteSetting.mutate(files[0])} formats={['.jpg', '.png']}>
-        <Card className="border-0 shadow-sm mt-3 text-center">
+    <div className="mb-3">
+      <h5> Branding Image </h5>
+      <FilesDragAndDrop
+        numOfFiles={1}
+        onDrop={(files) => updateSiteSetting.mutate(files[0])}
+        formats={['.jpg', '.png', '.svg']}
+      >
+        <Card className="border border-2 border-whitesmoke mt-3 text-center">
           <label htmlFor="file" className="presentation-upload">
             <Card.Body className="py-5 text-secondary cursor-pointer">
               <div className="icon-circle rounded-circle d-block mx-auto mb-3">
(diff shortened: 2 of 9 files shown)


scouillard:
Trying to access /assets/images from a PORO file (setting_getter) so we can set a default Branding Image.

app/controllers/api/v1/admin/site_settings_controller.rb
```diff
@@ -31,6 +32,22 @@ def update
...
+        def destroy
+          site_setting = SiteSetting.joins(:setting)
+                                    .find_by(
+                                      provider: current_provider,
+                                      setting: { name: params[:name] }
+                                    )
+
+          return render_error status: :not_found unless site_setting
+
+          site_setting.image.purge if params[:name] == 'BrandingImage'
```
KH-Amir-TN, line 45: We can check if the update operation succeeded,

app/javascript/components/admin/site_settings/appearance/DeleteBrandingImageModal.jsx
```diff
@@ -0,0 +1,31 @@
...
+  );
+}
+
+DeleteBrandingImageModal.propTypes = {
+  handleClose: PropTypes.func,
+};
+
+DeleteBrandingImageModal.defaultProps = {
+  handleClose: () => {},
+};
```
KH-Amir-TN, line 31: Why this is not consistent with its siblings `DeleteRoomForm, DeleteUserForm, DeleteRoleForm,...`?

app/javascript/components/admin/site_settings/appearance/BrandingImage.jsx
```diff
@@ -1,18 +1,51 @@
...
 
-      <FilesDragAndDrop numOfFiles={1} onDrop={(files) => updateSiteSetting.mutate(files[0])} formats={['.jpg', '.png']}>
-        <Card className="border-0 shadow-sm mt-3 text-center">
+  return (
+    <div className="mb-3">
+      <h5> Branding Image </h5>
+      <FilesDragAndDrop
+        numOfFiles={1}
+        onDrop={(files) => updateSiteSetting.mutate(files[0])}
+        formats={['.jpg', '.png']}
```
KH-Amir-TN, line 46: We're mentioning that we allow SVG files, but this does not seem to be the case.

app/javascript/components/admin/site_settings/appearance/BrandingImage.jsx
```diff
@@ -1,18 +1,51 @@
...
+import DeleteBrandingImageModal from './DeleteBrandingImageModal';
 
 export default function BrandingImage() {
   const updateSiteSetting = useUpdateSiteSetting('BrandingImage');
+  const { data: siteSettings } = useSiteSettings();
 
-  return (
-    <div className="mt-4">
-      <h6> Branding Image </h6>
+  if (siteSettings.BrandingImage.length > 1) {
```
KH-Amir-TN, line 15: Why `siteSettings.BrandingImage.length > 1`?
An empty string is a falsy value in JS,

app/javascript/components/admin/site_settings/appearance/BrandingImage.jsx
```diff
@@ -1,18 +1,51 @@
...
+    return (
+      <div className="mb-3">
```
KH-Amir-TN, line 17: I believe that this could be extracted into its own component,

app/javascript/components/admin/site_settings/appearance/BrandingImage.jsx
```diff
@@ -1,18 +1,51 @@
...
+          </Card.Body>
+        </Card>
+      </div>
+    );
+  }
 
-      <FilesDragAndDrop numOfFiles={1} onDrop={(files) => updateSiteSetting.mutate(files[0])} formats={['.jpg', '.png']}>
-        <Card className="border-0 shadow-sm mt-3 text-center">
+  return (
+    <div className="mb-3">
```
KH-Amir-TN, line 41: This also may be extracted into a component?

app/javascript/hooks/mutations/admin/site_settings/useDeleteSiteSetting.jsx
```diff
@@ -0,0 +1,20 @@
...
+
+export default function useDeleteSiteSetting(name) {
+  const queryClient = useQueryClient();
+
+  return useMutation(
+    (data) => axios.delete(`/admin/site_settings/${name}.json`, data),
+    {
+      onSuccess: () => {
+        queryClient.invalidateQueries('getSiteSettings');
+        toast.success('Branding Image deleted');
```
KH-Amir-TN, line 13: Is this always gonna be a `Branding Image deleted`?
Shouldn't this be something generic `Site setting deleted`?

app/javascript/components/admin/site_settings/appearance/BrandingImage.jsx
```diff
@@ -1,18 +1,51 @@
...
 import useUpdateSiteSetting from '../../../../hooks/mutations/admin/site_settings/useUpdateSiteSetting';
+import useSiteSettings from '../../../../hooks/queries/admin/site_settings/useSiteSettings';
 import FilesDragAndDrop from '../../../shared_components/utilities/FilesDragAndDrop';
+import Logo from '../../../shared_components/Logo';
+import Modal from '../../../shared_components/modals/Modal';
+import DeleteBrandingImageModal from './DeleteBrandingImageModal';
 
 export default function BrandingImage() {
   const updateSiteSetting = useUpdateSiteSetting('BrandingImage');
+  const { data: siteSettings } = useSiteSettings();
```
KH-Amir-TN, line 13: With this setup, the `SiteSettings` fetches the settings, then `BrandingImage` re-fetches those settings ending with the `Logo` component fetching the `BrandingImage` setting,
We should simplify this in the future,

app/controllers/api/v1/admin/site_settings_controller.rb
```diff
@@ -23,6 +23,7 @@ def update
 
           update = if params[:name] == 'BrandingImage'
                      site_setting.image.attach params[:site_setting][:value]
+                     site_setting.update(value: site_setting.image.blob.filename.to_s)
```
KH-Amir-TN, line 26: Um, I understand what you're trying to do.
It feels a bit hackery IMO, for now it's okay but I think that the `site_settings` fetching logic may require having its own service managing such cases,

app/javascript/components/admin/site_settings/appearance/DeleteBrandingImageModal.jsx
scouillard: Because I think there is a plan to refactor most of the forms to a onClick hook.
Forms are not required for a simple request.

app/controllers/api/v1/admin/site_settings_controller.rb
scouillard: Yeah agreed.
I wanted to add the feature with the current mechanics.

app/javascript/components/admin/site_settings/appearance/BrandingImage.jsx
scouillard: It could but I think it's fine as it is

app/javascript/components/admin/site_settings/appearance/BrandingImage.jsx
scouillard: Yeah I thought about making this a component since we are using something similar in Room/Presentation.
But this will have to be for another PR.

app/javascript/components/admin/site_settings/appearance/BrandingImage.jsx
scouillard: Good catch

app/controllers/api/v1/admin/site_settings_controller.rb
scouillard: Isn't there a way to make this global? In application controller?
So if an update, destroy, create on a resource fails it returns a :bad_request?
Our methods are getting bulky with all the status returns.

app/javascript/components/admin/site_settings/appearance/BrandingImage.jsx
scouillard: Yeah, it's weird that siteSetting.BrandingImage and siteSettings.BrandingImage doesn't return the same thing.
I will remove the filename as it's not needed.

app/controllers/api/v1/admin/site_settings_controller.rb
KH-Amir-TN: Of course it's feasible,
But, I don't think it's a good idea, it raises its own issues and complexity.
In the future there's a good chance that we'll be returning some information about `:bad_request` cases,

app/javascript/components/admin/site_settings/appearance/BrandingImage.jsx
KH-Amir-TN: Of course, that's what I meant

app/javascript/components/admin/site_settings/appearance/BrandingImage.jsx
KH-Amir-TN: Yup, that is why we need to come back to this so that a service will handle that properly and be the source of truth ensuring consistency,

scouillard: After some talks with @farhatahmad , we will be dropping the destroy BrandingImage completely
That solves all our problems for now.

app/javascript/hooks/mutations/admin/site_settings/useUpdateSiteSetting.jsx
```diff
@@ -5,10 +5,23 @@ import axios from '../../../../helpers/Axios';
...
+  const imageValidation = (image) => {
+    const FILE_SIZE = 3_000_000;
+    const SUPPORTED_FORMATS = 'image/jpeg|image/png|image/svg';
+
+    if (image.size > FILE_SIZE) {
+      throw new Error('File is too large');
+    } else if (!image.type.match(SUPPORTED_FORMATS)) {
+      throw new Error('File type is not supported');
+    }
+  };
```
scouillard, line 18: This is needed as the FilesDragAndDrop component checks for the file type when _draggin' and droppin'_ the file, but not if the user clicks on the input and selects from it's files.

This should be added to the Presentation hook.

Also, this will replace the (overly complicated) create/update AvatarForm as we will get rid of the <Form> component, helpers, etc AND get rid of an extra button.

KH-Amir-TN: This is indeed a good idea, we could extend it to be a helper that we can leverage throughout the app.
My only concern is how we're displaying all errors on the `onError` callback,

scouillard: I see what you mean.
Should we hardcode something like
`if not 'file too large' or 'file type is not supported' then send 'generic error message'`

farhatahmad (dismissed):

app/controllers/api/v1/admin/site_settings_controller.rb
```diff
@@ -26,6 +26,7 @@ def update
...
```
farhatahmad, line 29: Do we still need to store the file name?

scouillard: Nope, forgot to remove it from backend. Good catch

farhatahmad (approved):

PR #3774, opened 2022-08-17
Renamed users column to participants to avoid naming collisions

farhatahmad:

### DIFF
--- a/app/models/recording.rb
+++ b/app/models/recording.rb
@@ -2,13 +2,14 @@
 
 class Recording < ApplicationRecord
   belongs_to :room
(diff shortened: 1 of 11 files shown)


app/models/recording.rb
```diff
@@ -2,13 +2,14 @@
 
 class Recording < ApplicationRecord
   belongs_to :room
+  has_one :user, through: :room
```
farhatahmad, line 5: Because this association was added, rails was getting confused between `users` and `user` so I renamed `users` to `participants`

scouillard (approved):

PR #3776, opened 2022-08-17
Room settings: Upgraded `RoomSettingsGetter` to hide access codes values

KH-Amir-TN:

### DIFF
--- a/app/services/room_settings_getter.rb
+++ b/app/services/room_settings_getter.rb
@@ -7,10 +7,11 @@ class RoomSettingsGetter
   # Hash(`<option_name> => {'true' => <Postive>, 'false' => <Negative>})`
   SPECIAL_OPTIONS = { 'guestPolicy' => { 'true' => 'ASK_MODERATOR', 'false' => 'ALWAYS_ACCEPT' } }.freeze
 
-  def initialize(room_id:, provider:, only_enabled: false, only_bbb_options: false)
+  def initialize(room_id:, provider:, show_codes: false, only_enabled: false, only_bbb_options: false)
     @room_id = room_id
     @only_bbb_options = only_bbb_options
     @only_enabled = only_enabled
+    @show_codes = show_codes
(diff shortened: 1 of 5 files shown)


app/services/room_settings_getter.rb
```diff
@@ -54,13 +55,17 @@ def infer_specials(room_settings:)
...
 
   def infer_codes(room_settings:, access_codes:)
-    access_codes.each do |key, code|
-      case room_settings[key]
-      when 'false'
-        room_settings[key] = '' # Forced disbaled access code will have an empty value.
-      when 'true'
-        room_settings[key] = code # Forced enabled access code will conserve its   original value.
-      end
+    filtered_access_codes = access_codes.slice(*room_settings.keys) # Filtering the available room access codes to minimize iterations.
```
KH-Amir-TN, line 58: Access codes when forced disabled and when `only_enabled: true` will be filtered out.
To minimize iteration we ensure that we iterate over the original access codes that do exist on the room setting after filtration. 
Besides it helps simplifies the logic below,

app/services/room_settings_getter.rb
```diff
@@ -7,10 +7,11 @@ class RoomSettingsGetter
   # Hash(`<option_name> => {'true' => <Postive>, 'false' => <Negative>})`
   SPECIAL_OPTIONS = { 'guestPolicy' => { 'true' => 'ASK_MODERATOR', 'false' => 'ALWAYS_ACCEPT' } }.freeze
 
-  def initialize(room_id:, provider:, only_enabled: false, only_bbb_options: false)
+  def initialize(room_id:, provider:, show_codes: false, only_enabled: false, only_bbb_options: false)
```
KH-Amir-TN, line 10: For security reasons by default access code values will be hidden,

farhatahmad:
Very nicely done 🔥

farhatahmad (approved):

PR #3779, opened 2022-08-17
Fix Spinner UI glitch in Admin tabs

scouillard:
## Description

Fix the Spinner UI glitch in Site Settings and Room Configs

## Testing Steps

- Before pulling the request, move between admin tabs, spam refresh and move between admin tabs.
- Do the same thing after pulling the request.

### DIFF
--- a/app/javascript/components/admin/room_configuration/RoomConfig.jsx
+++ b/app/javascript/components/admin/room_configuration/RoomConfig.jsx
@@ -29,61 +29,63 @@ export default function RoomConfig() {
                   <div className="p-4 border-bottom">
                     <h2> Room Configuration </h2>
                   </div>
-                  {
-                    (isLoading && <Row><Spinner /></Row>) || (
-                      <div className="p-4">
-                        <RoomConfigRow
-                          title="Mute user when they join"
-                          subtitle="Automatically mutes the user when they join the BigBlueButtonMeeting"
-                          mutation={() => useUpdateRoomConfig('muteOnStart')}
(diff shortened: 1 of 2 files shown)


app/javascript/components/admin/room_configuration/RoomConfig.jsx
```diff
@@ -29,61 +28,55 @@ export default function RoomConfig() {
...
-                          subtitle="Allows room owners to have a random alphanumeric code that can be shared with users.
-                          The code **if generated** will not be required and when used on any room meeting will join the user as a moderator."
-                          mutation={() => useUpdateRoomConfig('glModeratorAccessCode')}
-                          value={roomConfigs.glModeratorAccessCode}
-                        />
-                      </div>
-                    )
-                  }
+                  <div className="p-4">
+                    <RoomConfigRow
```
KH-Amir-TN, line 32: I see what you're trying to do,
I think that we should figure out why the spinner is glitching and not just omitting its use.
Because this will render the UI but without data, room configs will have no values until the data becomes available. 

PS: For you now you'll see that the configs aren't reflecting their values in the database and that is because of another issue that your PR helped me to spot.
I'll address it in another PR,

scouillard: I put the isLoading / Spinner back in a different way - the glitchy visual is gone but it still lags.
You are right, the isLoading is important.
Maybe we could add the isLoading as a more global so it's not so clumbersome in every component.. just some thoughts.

farhatahmad (approved):

PR #3786, opened 2022-08-18
Frontend access restrictions for regular users

scouillard:
## Description

Basic frontend access restrictions for the regular user.

### DIFF
--- a/app/javascript/components/admin/AdminNavSideBar.jsx
+++ b/app/javascript/components/admin/AdminNavSideBar.jsx
@@ -35,20 +35,20 @@ export default function AdminNavSideBar() {
         </Nav.Item>
       )}
       {(currentUser.permissions.ManageSiteSettings === 'true') && (
-        <Nav.Item>
-          <Nav.Link className="cursor-pointer text-muted" as={Link} to="/adminpanel/site-settings" eventKey="site-settings">
-            <CogIcon className="hi-s me-3" />
-            Site Settings
-          </Nav.Link>
-        </Nav.Item>
-      )}
-      {(currentUser.permissions.ManageSiteSettings === 'true') && (
-        <Nav.Item>
-          <Nav.Link className="cursor-pointer text-muted" as={Link} to="/adminpanel/room-configuration" eventKey="room-configuration">
-            <AdjustmentsIcon className="hi-s me-3" />
-            Room Configuration
-          </Nav.Link>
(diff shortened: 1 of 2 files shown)


scouillard:
Work in progress.
Starting to think that maybe we should split the "user" permissions (create room, shared list) from the "admin" permissions (Manage users, rooms, recordings, etc).
Or maybe return the permissions as an array,
permissions = [ { ManageUser: { value: true, type: 'admin' } }, { CreateRoom: { value: true, type: 'user' } } ]

app/javascript/components/admin/AdminNavSideBar.jsx
```diff
@@ -32,18 +34,22 @@ export default function AdminNavSideBar() {
...
+        </Nav.Item>
+      )}
+      {(currentUser.permissions.ManageSiteSettings === 'true') && (
+        <Nav.Item>
+          <Nav.Link className="cursor-pointer text-muted" as={Link} to="/adminpanel/room-configuration" eventKey="room-configuration">
+            <AdjustmentsIcon className="hi-s me-3" />
+            Room Configuration
+          </Nav.Link>
+        </Nav.Item>
+      )}
```
farhatahmad, line 52: You can just do this in 1 block - no need to check twice

app/javascript/components/shared_components/Header.jsx
```diff
@@ -13,6 +13,17 @@ export default function Header() {
...
 
+  const adminAccess = () => {
+    const { permissions } = currentUser;
+    delete permissions.SharedList;
+    delete permissions.CreateRoom;
+
+    if (Object.values(permissions).includes('true')) {
+      return <NavDropdown.Item as={Link} to="/adminpanel">Admin Panel</NavDropdown.Item>;
+    }
+    return false;
```
farhatahmad, line 24: This is trying to be too smart - what happens if we add another permission unrelated to admin? 

When it comes to security, always manually check for trues, rather than remove the unneeded permissions and then checking blindly for true

app/javascript/components/admin/AdminNavSideBar.jsx
scouillard: Is there a specific permission for Room Configs? Is it SiteSettings?

farhatahmad: Yeah

PR #3796, opened 2022-08-23
Admin UX changes

farhatahmad:
## Description

- Renamed `adminpanel` route to `admin` 
- Create a `RolePill` component that displays color

### DIFF
--- a/app/javascript/components/admin/roles/RolePill.jsx
+++ b/app/javascript/components/admin/roles/RolePill.jsx
@@ -1,8 +1,9 @@
 import React from 'react';
-import {Badge} from "react-bootstrap";
+import { Badge } from 'react-bootstrap';
+import PropTypes from 'prop-types';
 
 export default function RolePill({ role }) {
-  const { name, color } = role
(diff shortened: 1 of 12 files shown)


app/javascript/components/admin/roles/RolePill.jsx
```diff
@@ -0,0 +1,12 @@
...
+
+export default function RolePill({ role }) {
+  const { name, color } = role
+
+  return (
+    <Badge pill ref={(el) => el && el.style.setProperty('background-color', color, 'important')}>
+      {name}
+    </Badge>
+  );
+}
```
KH-Amir-TN, line 12: It may be good to include a prop type entry for the role,

KH-Amir-TN (approved):

app/javascript/components/admin/roles/RolePill.jsx
farhatahmad: Good catch - done

KH-Amir-TN (approved):

PR #3798, opened 2022-08-23
CanRecord Permission

hadicheaito1:
## Description

- CanRecord permission 
- Front end:
  - Removes the permission from roles page depending on room config record value
  - Remove the room setting based on room config and CanRecord permission
- Backend:
  - Handles the permission being set as well as accounting for room settings and room config
## Testing Steps

- Pull the code and test the different scenarios such as when room config is optional, and CanRecord permission is set to false, then you shouldn't be able to record and the record room setting shouldn't be displayed

## TODO in other PR's
- Make row component for the roles on roles admin panel page + CSS stuff
- Making the update RolePermission route better?
- Handling the ESLint complaint
- Roles page needs to be re-structured completely

### DIFF
--- a/app/javascript/components/rooms/room/room_settings/RoomSettings.jsx
+++ b/app/javascript/components/rooms/room/room_settings/RoomSettings.jsx
@@ -11,8 +11,10 @@ import DeleteRoomForm from '../forms/DeleteRoomForm';
 import useRoomConfigs from '../../../../hooks/queries/admin/room_configuration/useRoomConfigs';
 import AccessCodeRow from './AccessCodeRow';
 import useUpdateRoomSetting from '../../../../hooks/mutations/room_settings/useUpdateRoomSetting';
+import { useAuth } from '../../../../contexts/auth/AuthProvider';
 
 export default function RoomSettings() {
+  const currentUser = useAuth();
   const { friendlyId } = useParams();
   const roomSetting = useRoomSettings(friendlyId);
   const roomsConfigs = useRoomConfigs();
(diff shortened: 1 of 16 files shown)


app/javascript/components/rooms/room/room_settings/RoomSettings.jsx
```diff
@@ -10,15 +10,15 @@ import AccessCodes from './AccessCodes';
...
 import DeleteRoomForm from '../forms/DeleteRoomForm';
 import useRoomConfigs from '../../../../hooks/queries/admin/room_configuration/useRoomConfigs';
+import useRolePermissions from '../../../../hooks/queries/admin/role_permissions/useRolePermissions';
 
 export default function RoomSettings() {
   const { friendlyId } = useParams();
   const roomSetting = useRoomSettings(friendlyId);
   const roomsConfigs = useRoomConfigs();
-
+  const rolePermissions = useRolePermissions();
```
farhatahmad, line 19: Instead of loading this, you can just do `currentUser.permissions.CanRecord == 'true'`

app/services/room_settings_getter.rb
```diff
@@ -23,6 +24,12 @@ def initialize(room_id:, provider:, settings: [], show_codes: false, only_enable
   end
 
   def call
+    # checking if CanRecord permission is set to true when RoomConfig record is optional
+    if @rooms_configs['record'].nil? && (RolePermission.joins(:permission).find_by(role_id: @current_user.role_id,
+                                                                                   permission: { name: 'CanRecord' })&.value == 'false')
+      @rooms_configs['record'] = 'false'
+    end
```
farhatahmad, line 31: I would move this logic down below the call to `infer_codes` just so that it doesn't get overwritten. I would also move it to a private method to make it a bit cleaner to read. 

```
def infer_delete(room_settings:)
  return if RolePermission.joins(:permission).find_by(role_id: @current_user.role_id, permission: { name: 'CanRecord' }) == 'true'

  room_settings['record'] = false
end
```

farhatahmad: Something like this (needs to be tested) ^

hadicheaito1: Yup, tests will be part of next PR

farhatahmad (approved):

PR #3808, opened 2022-08-29
Replaced ids with uuids

farhatahmad:
## Description

Sequential ids are slightly problematic
1- They're guessable (ie add 1 to get to the next user)
2- They provide extra unneeded information (ie how many users)

I replaced them with `uuids` in the database

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -30,7 +30,7 @@ module Api
         return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
         if user.save
-          session[:user_id] = user.id
+          session[:user_id] ||= user.id
           token = user.generate_activation_token!
           render_data data: { token: }, status: :created # TODO: enable activation email sending.
         else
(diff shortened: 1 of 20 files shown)


app/controllers/api/v1/users_controller.rb
```diff
@@ -30,7 +30,7 @@ def create
         return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
         if user.save
-          session[:user_id] ||= user.id
+          session[:user_id] ||= user.reload.id
```
scouillard, line 33: This doesn't seem to work.
Upon creating an account, it does not log you in - it buffers until the page crashes.

KH-Amir-TN (approved):

app/controllers/api/v1/users_controller.rb
farhatahmad: Hmm it was crashing before I added and the reload fixed it - let me retest

farhatahmad: Just had to do with a stale session - fixed it by removing the `||=`

scouillard:
We need the ||= because we are using the users#create endpoint as an admin to create a new user.

We don't want to replace the session[:user_id] with the new user if the new user is made via the admin interface.

KH-Amir-TN (approved):

PR #3812, opened 2022-08-30
Extended permission checker to take array of names

farhatahmad:
## Description

`PermissionChecker` now optional takes an `Array` of permission names (or just 1 name) and returns true if any of the permissions return true 

## Testing Steps

1. Set `ManageRoles` to true for your current user 
2. Ensure the `Manage Roles` tab in the admin panel loads correctly
3. Set `ManageUsers` to true for your current user 
4. Ensure the `Edit User` page loads with correct dropdown for Roles

### DIFF
--- a/app/javascript/components/users/user/forms/UpdateUserForm.jsx
+++ b/app/javascript/components/users/user/forms/UpdateUserForm.jsx
@@ -35,7 +35,7 @@ export default function UpdateUserForm({ user }) {
   const fields = updateUserFormFields;
   const currentUser = useAuth();
 
-  const isAdmin = currentUser.permissions.ManageUsers === 'true'
+  const isAdmin = currentUser.permissions.ManageUsers === 'true';
   const { data: roles, isLoading } = useRoles('', isAdmin);
 
   if (isLoading) return <Spinner />;
--- a/app/javascript/components/users/user/forms/UpdateUserForm.jsx
+++ b/app/javascript/components/users/user/forms/UpdateUserForm.jsx
@@ -34,7 +34,9 @@ export default function UpdateUserForm({ user }) {
   const updateUser = useUpdateUser(user?.id);
   const fields = updateUserFormFields;
   const currentUser = useAuth();
-  const { data: roles, isLoading } = useRoles();
+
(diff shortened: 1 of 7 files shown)


app/javascript/components/users/user/forms/UpdateUserForm.jsx
```diff
@@ -34,7 +34,9 @@ export default function UpdateUserForm({ user }) {
   const updateUser = useUpdateUser(user?.id);
   const fields = updateUserFormFields;
   const currentUser = useAuth();
-  const { data: roles, isLoading } = useRoles();
+
+  const isAdmin = currentUser.permissions.ManageUsers === 'true'
+  const { data: roles, isLoading } = useRoles('', isAdmin);
```
farhatahmad, line 39: Only fire the query if the user is an Admin

app/services/permissions_checker.rb
```diff
@@ -1,25 +1,31 @@
...
-    when 'ManageRecordings'
-      return authorize_manage_recordings
+    return true if RolePermission.joins(:permission).exists?(
+      role_id: @current_user.role_id,
+      permission: { name: @permission_names },
+      value: 'true'
+    )
+
+    # convert to array if only checking for 1 permission (for simplicity)
+    Array(@permission_names).each do |permission|
```
farhatahmad, line 20: `Array('string')` returns `['string']`
`Array(['string'])` returns `['string']`

app/javascript/components/users/user/forms/UpdateUserForm.jsx
```diff
@@ -34,7 +34,9 @@ export default function UpdateUserForm({ user }) {
   const updateUser = useUpdateUser(user?.id);
   const fields = updateUserFormFields;
   const currentUser = useAuth();
-  const { data: roles, isLoading } = useRoles();
+
+  const isAdmin = currentUser.permissions.ManageUsers === 'true';
+  const { data: roles, isLoading } = useRoles('', isAdmin);
```
farhatahmad, line 39: Only fire the query if the user is an Admin

scouillard (approved):

PR #3821, opened 2022-09-01
Adding all permissions to edit role page

hadicheaito1:
## Description

- Adding all the permissions to the edit roles page
- `ManageRoles` permission does not show if the current user is looking at their own role's edit page
- Fixed bug that would show the wrong name in the input box

image

### DIFF
--- a/app/javascript/components/admin/roles/forms/EditRoleForm.jsx
+++ b/app/javascript/components/admin/roles/forms/EditRoleForm.jsx
@@ -13,6 +13,7 @@ import useUpdateRolePermission from '../../../../hooks/mutations/admin/role_perm
 import useRoomConfigs from '../../../../hooks/queries/admin/room_configuration/useRoomConfigs';
 import useRolePermissions from '../../../../hooks/queries/admin/role_permissions/useRolePermissions';
 import RolePermissionRow from '../RolePermissionRow';
+import { useAuth } from '../../../../contexts/auth/AuthProvider';
 
 export default function EditRoleForm({ role }) {
   const methods = useForm(editRoleFormConfig);
@@ -23,6 +24,7 @@ export default function EditRoleForm({ role }) {
   fields.name.placeHolder = defaultValues.name;
   const roomConfigs = useRoomConfigs();
   const rolePermissions = useRolePermissions(role.id);
+  const currentUser = useAuth();
 
   useEffect(
     () => {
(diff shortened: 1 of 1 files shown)


app/javascript/components/admin/roles/forms/EditRoleForm.jsx
```diff
@@ -41,13 +43,64 @@ export default function EditRoleForm({ role }) {
         </Form>
         {(roomConfigs.isLoading || rolePermissions.isLoading || roomConfigs.data.record === 'optional') && (
           <Stack>
+            <RolePermissionRow
+              permissionName="CreateRoom"
+              description="Can create rooms"
+              roleId={role.id}
+              defaultValue={rolePermissions.data.CreateRoom === 'true'}
```
scouillard, line 50: Maybe we could remove the .data from the admin/role_permissions controller in this PR?

scouillard (dismissed):

app/javascript/components/admin/roles/forms/EditRoleForm.jsx
hadicheaito1: The `.data` is the actual data that is returned from the `GET` request done in `useRolePermissions` (in this case it's the hash).
I see this is the way we are doing it in other places, the alternative would be to do this:
```
const { data: rolePermissions, isLoading } = useRolePermissions(role.id);
rolePermissions.CreateRoom === 'true' ...

```

This is one of those things where we need to discuss which format we should stick to

farhatahmad: If we replace ` const rolePermissions = useRolePermissions(role.id);` with  ` const {data: rolePermissions, isLoading: rolePermissionsIsLoading = useRolePermissions(role.id);` wouldn't that solve our problem?

app/javascript/components/admin/roles/forms/EditRoleForm.jsx
```diff
@@ -41,13 +43,64 @@ export default function EditRoleForm({ role }) {
...
+              updateMutation={updateRolePermission}
+            />
+            <RolePermissionRow
+              permissionName="ManageSiteSettings"
+              description="Allow users with this role to manage site settings"
+              roleId={role.id}
+              defaultValue={rolePermissions.data.ManageSiteSettings === 'true'}
+              updateMutation={updateRolePermission}
+            />
+            {(currentUser.role.id !== role.id) && (
```
farhatahmad, line 88: Add a comment here explaining

app/javascript/components/admin/roles/forms/EditRoleForm.jsx
hadicheaito1: Yeah it should work, just we have 2 different ways of doing things like this, that's why I was saying we should agree on one approach. I guess we can do the way mentioned now on and later on change the places where the .data is used

scouillard: Right right we are renaming data when destructuring the object

hadicheaito1: Agreed

farhatahmad (approved):

PR #3822, opened 2022-09-01
Made AnyoneCanStart setting work properly

farhatahmad:
## Testing Steps

1- Enable the setting on the room
2- Try to join before starting the meeting
3- Ensure the room starts and the user is automatically joined


spec/controllers/meetings_controller_spec.rb
```diff
@@ -4,29 +4,28 @@
 
 RSpec.describe Api::V1::MeetingsController, type: :controller do
```
farhatahmad, line 5: Mostly refactoring

hadicheaito1 (approved):

PR #3825, opened 2022-09-01
Increase password complexity

farhatahmad:
## Description

Increases password complexity to match the following rules: 
- Atleast 1 uppercase
- Atleast 1 lowercase
- Atleast 1 symbol
- Atleast 1 number

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -25,7 +25,11 @@ class User < ApplicationRecord
   validates :email, presence: true, uniqueness: { case_sensitive: false, scope: :provider }
   validates :provider, presence: true
 
-  validates :password, presence: true, confirmation: true, on: :create, unless: :external_id?
+  validates :password,
+            presence: true,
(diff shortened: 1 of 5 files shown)


scouillard (approved):

app/models/user.rb
```diff
@@ -25,7 +25,11 @@ class User < ApplicationRecord
   validates :email, presence: true, uniqueness: { case_sensitive: false, scope: :provider }
   validates :provider, presence: true
 
-  validates :password, presence: true, confirmation: true, on: :create, unless: :external_id?
+  validates :password,
+            presence: true,
+            format: /\A(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*(\W|_)).{8,}\z/,
```
hadicheaito1, line 30: Should we add a comment that explains what the format needs to be so it's easier to read :
`Atleast 1 uppercase, Atleast 1 lowercase, Atleast 1 symbol, Atleast 1 number`

hadicheaito1 (approved):

app/models/user.rb
farhatahmad: Will do it in  follow up pr

PR #3836, opened 2022-09-07
Implemented require authentication room setting

farhatahmad:
## Description

Implemented the `Require Authentication` as a room setting. At the moment, the user will be redirected to the home page if they are not signed in with a toast message

TODO:
- Implement `return_to` functionality so that a user returns to the signed in room after signing in

## Testing Steps

- Enable the setting 
- Copy join url and paste into incognito window
- Ensure you are redirected to home page with an error
- Sign into incognito window
- Repeat and ensure you are not redirected and are able to join

### DIFF
--- a/app/assets/stylesheets/rooms.scss
+++ b/app/assets/stylesheets/rooms.scss
@@ -17,7 +17,6 @@
 
 
 .room-settings-row {
-  max-width: 400px;
   line-height: 28px;
 }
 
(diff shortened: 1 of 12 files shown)


scouillard:
I get a toast error when changing the `Require users to be signed in before joining` field in admin/room config.
Not sure from the logs what the problem is.

hadicheaito1:
> I get a toast error when changing the `Require users to be signed in before joining` field in admin/room config. Not sure from the logs what the problem is.
> 
> image

Same

farhatahmad:
My bad - forgot to mention you have to drop your DB because of the changed migration

scouillard:
Trying to join the Meeting if you are already logged in doesn't work.
image

farhatahmad:
@scouillard good catch - fixed now

scouillard (approved):

PR #3839, opened 2022-09-07
CreateRoom permission for Sign-in + protecting backend

hadicheaito1:
## Description

- Back-end endpoints are protected if user does not have permission
- User is redirected to `MainPage` on sign in if they don't have proper permission
## Testing Steps

- Turn `CreateRoom` permission to false for User role 
- Log in as regular user role
- Should be redirected to `/main-page` instead of `rooms`
## TODO
- Do the same for sign in
- UI

### DIFF
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -6,7 +6,7 @@ module Api
       skip_before_action :ensure_authenticated, only: %i[public_show]
 
       before_action :find_room, only: %i[show update destroy recordings recordings_processing purge_presentation public_show]
-      before_action do
+      before_action except: %i[public_show] do
         ensure_authorized('CreateRoom')
       end
       before_action only: %i[create] do
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -6,11 +6,12 @@ module Api
       skip_before_action :ensure_authenticated, only: %i[public_show]
 
       before_action :find_room, only: %i[show update destroy recordings recordings_processing purge_presentation public_show]
-
+      before_action do
+        ensure_authorized('CreateRoom')
+      end
       before_action only: %i[create] do
         ensure_authorized('ManageUsers', user_id: room_params[:user_id])
       end
-
       before_action only: %i[show destroy] do
         ensure_authorized('ManageRooms', friendly_id: params[:friendly_id])
       end
--- a/app/javascript/hooks/mutations/sessions/useCreateSession.jsx
+++ b/app/javascript/hooks/mutations/sessions/useCreateSession.jsx
@@ -6,13 +6,20 @@ import axios from '../../../helpers/Axios';
 export default function useCreateSession(token) {
   const queryClient = useQueryClient();
   const navigate = useNavigate();
-
+  let createRoomPermission = true;
   return useMutation(
-    (session) => axios.post('/sessions.json', { session, token }),
+    (session) => axios.post('/sessions.json', { session, token }).then((response) => {
+      createRoomPermission = response.data.data.permissions.CreateRoom;
+    }),
     {
       onSuccess: () => {
         queryClient.invalidateQueries('useSessions');
-        navigate('/rooms');
+        // if the current user does NOT have the CreateRoom permission, then do not re-direct to rooms page
+        if (createRoomPermission === 'false') {
+          navigate('/main-page');
+        } else {
+          navigate('/rooms');
+        }
       },
       onError: () => {
         toast.error('Incorrect username or password. \n Please try again');
--- a/app/javascript/hooks/mutations/sessions/useCreateSession.jsx
+++ b/app/javascript/hooks/mutations/sessions/useCreateSession.jsx
@@ -6,17 +6,15 @@ import axios from '../../../helpers/Axios';
 export default function useCreateSession(token) {
   const queryClient = useQueryClient();
   const navigate = useNavigate();
-  let createRoomPermission = true;
+
   return useMutation(
-    (session) => axios.post('/sessions.json', { session, token }).then((response) => {
-      createRoomPermission = response.data.data.permissions.CreateRoom;
-    }),
+    (session) => axios.post('/sessions.json', { session, token }),
     {
-      onSuccess: () => {
+      onSuccess: (response) => {
         queryClient.invalidateQueries('useSessions');
         // if the current user does NOT have the CreateRoom permission, then do not re-direct to rooms page
-        if (createRoomPermission === 'false') {
-          navigate('/main-page');
+        if (response.data.data.permissions.CreateRoom === 'false') {
+          navigate('/home');
         } else {
           navigate('/rooms');
         }
--- a/app/javascript/main.jsx
+++ b/app/javascript/main.jsx
@@ -25,6 +25,7 @@ import Roles from './components/admin/roles/Roles';
 import ResetPassword from './components/users/password_management/ResetPassword';
 import EditUser from './components/admin/manage_users/EditUser';
 import EditRole from './components/admin/roles/EditRole';
+import MainPage from './components/rooms/room/MainPage';
 
 const queryClient = new QueryClient();
 
@@ -52,6 +53,7 @@ const root = (
             <Route path="/rooms" element={<Rooms />} />
             <Route path="/rooms/:friendlyId" element={<Room />} />
             <Route path="/rooms/:friendlyId/join" element={<RoomJoin />} />
+            <Route path="/main-page" element={<MainPage />} />
             <Route path="*" element={<h1 className="text-center">404</h1>} />
           </Route>
         </Routes>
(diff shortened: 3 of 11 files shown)


hadicheaito1:
Suggestions for this new page/route name? Something instead of `/main-page` ?

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -6,11 +6,12 @@ class RoomsController < ApiController
       skip_before_action :ensure_authenticated, only: %i[public_show]
 
       before_action :find_room, only: %i[show update destroy recordings recordings_processing purge_presentation public_show]
-
+      before_action except: %i[public_show] do
```
hadicheaito1, line 9: I added the `except` after looking at the rspec test.
I don't fully understand why the `session[:user_id] = nil` in the rspec test for this. Could someone explain the usage of `public_show`

scouillard:
Shouldn't the "createRoom" permission just deal with the "create room" button pretty much? 
A user without the permission would still be able to belong to SharedRooms, and access /rooms, right?

app/controllers/api/v1/rooms_controller.rb
scouillard: `session[:user_id] = nil` is to log off the current_user. So in specs it is usually to make sure some one not logged in does not have the access to something.

hadicheaito1:
> Shouldn't the "createRoom" permission just deal with the "create room" button visibility essentially? A user without the permission would still be able to belong to SharedRooms, and access /rooms, right?

I wish :sweat_smile: 
From what I understand, if `CreateRoom` permission is set to false, then it disallows the user from creating rooms, accessing any room (as well as SharedRooms and all that), as well as recordings (since recordings belong to rooms).

The permission name is misleading imo

app/controllers/api/v1/rooms_controller.rb
hadicheaito1: Yeah i'm aware that it signs out the user. I'm more looking for why do that for this specific test and why do they still have access to `public_show` even thought they are not logged in

scouillard: Right. I'm also unsure of what is going on with `public_show` (and `publicRoom` from #3836)

scouillard:
Wow, okay.
Maybe we should redirect those users to our landing page at '/' and maybe remove the sign in sign up button if the user is signed in?
We also have to figure out something for '/' because right now it crashes if a logged in user tries to access it.

hadicheaito1:
> Wow, okay. Maybe we should redirect those users to our landing page at '/' and maybe remove the sign in sign up button if the user is signed in? I mean it seems like there is not much those users can do but have access to their profile via the navbar. We also have to figure out something for '/' because right now it crashes if a logged in user tries to access it.

Based on v2, the users without this permission are able to access their own profile and are able to paste a join url to a meeting. This might change for v3.

Yeah it's crashing because of the forbidden status being returned. Soon we are going to have to make a page for this case so that something is rendered and it doesn't just break the app

scouillard:
I think that is how I would proceed for the '/' homepage.
Case 1 - User is not logged
Regular homepage, full height without navbar.
Case 2 - User is logged in && has createRoom permission
'/' redirects to '/rooms'
Case 3 - User is logged in && does NOT have createRoom permission
Regular homepage with navbar.

scouillard:
> > Wow, okay. Maybe we should redirect those users to our landing page at '/' and maybe remove the sign in sign up button if the user is signed in? I mean it seems like there is not much those users can do but have access to their profile via the navbar. We also have to figure out something for '/' because right now it crashes if a logged in user tries to access it.
> 
> Based on v2, the users without this permission are able to access their own profile and are able to paste a join url to a meeting. This might change for v3.
> 
> Yeah it's crashing because of the forbidden status being returned. Soon we are going to have to make a page for this case so that something is rendered and it doesn't just break the app

I'm sorry I might be dragging out of this PR scope - what is returning forbidden?

spec/controllers/rooms_controller_spec.rb
```diff
@@ -3,9 +3,9 @@
 require 'rails_helper'
 
 RSpec.describe Api::V1::RoomsController, type: :controller do
-  let(:user) { create(:user) }
-  let(:user_with_manage_rooms_permission) { create(:user, :with_manage_rooms_permission) }
-  let(:user_with_manage_users_permission) { create(:user, :with_manage_users_permission) }
+  let(:user) { create(:user, :with_create_room_permission) }
+  let(:user_with_manage_rooms_permission_and_create_room_permission) { create(:user, :with_manage_rooms_permission, :with_create_room_permission) }
+  let(:user_with_manage_users_permission_and_create_room_permission) { create(:user, :with_manage_users_permission, :with_create_room_permission) }
```
scouillard, line 8: Those are duplicated

hadicheaito1:
> > > Wow, okay. Maybe we should redirect those users to our landing page at '/' and maybe remove the sign in sign up button if the user is signed in? I mean it seems like there is not much those users can do but have access to their profile via the navbar. We also have to figure out something for '/' because right now it crashes if a logged in user tries to access it.
> > 
> > 
> > Based on v2, the users without this permission are able to access their own profile and are able to paste a join url to a meeting. This might change for v3.
> > Yeah it's crashing because of the forbidden status being returned. Soon we are going to have to make a page for this case so that something is rendered and it doesn't just break the app
> 
> I'm sorry I might be dragging out of this PR scope - what is returning forbidden? The crash on '/' if logged in is in V3 current state.

My bad I misread your comment. Ignore the forbidden status stuff

spec/controllers/rooms_controller_spec.rb
hadicheaito1: One is for users and one is for rooms

spec/controllers/recordings_controller_spec.rb
```diff
@@ -3,8 +3,10 @@
 require 'rails_helper'
 
 RSpec.describe Api::V1::RecordingsController, type: :controller do
-  let(:user) { create(:user) }
-  let(:user_with_manage_recordings_permission) { create(:user, :with_manage_recordings_permission) }
+  let(:user) { create(:user, :with_create_room_permission) }
+  let(:user_with_manage_recordings_permission_and_create_room_permission) do
+    create(:user, :with_manage_recordings_permission, :with_create_room_permission)
+  end
```
scouillard, line 9: The createRoom permission is so restrictive, maybe we should just enable it by default with every other permissions?
Else we will have to do this for all of our specs pretty much, no?
Maybe we could have a negative logic trait `:without_create_room_permission` instead and use it when we need to test the createRoom permission.

spec/controllers/recordings_controller_spec.rb
```diff
@@ -3,8 +3,10 @@
 require 'rails_helper'
 
 RSpec.describe Api::V1::RecordingsController, type: :controller do
-  let(:user) { create(:user) }
-  let(:user_with_manage_recordings_permission) { create(:user, :with_manage_recordings_permission) }
+  let(:user) { create(:user, :with_create_room_permission) }
```
scouillard, line 6: I dont think the :with_create_room_permission trait belongs here

spec/controllers/recordings_controller_spec.rb
scouillard: Something like:
```
  let(:user_with_manage_recordings_permission) do
    create(:user, :with_manage_recordings_permission, :with_create_room_permission)
  end
```

And if needed:
```
  let(:user_with_manage_recordings_permission) do
    create(:user, :with_manage_recordings_permission, :without_create_room_permission)
  end
```

spec/controllers/rooms_controller_spec.rb
scouillard: Lol right I'm sorry

spec/controllers/recordings_controller_spec.rb
scouillard: Actually, maybe there is a way to treat a user without the createRoom permission almost like an unauthorised user instead?

app/javascript/hooks/mutations/sessions/useCreateSession.jsx
```diff
@@ -6,13 +6,20 @@ import axios from '../../../helpers/Axios';
...
     {
       onSuccess: () => {
         queryClient.invalidateQueries('useSessions');
-        navigate('/rooms');
+        // if the current user does NOT have the CreateRoom permission, then do not re-direct to rooms page
+        if (createRoomPermission === 'false') {
+          navigate('/main-page');
+        } else {
+          navigate('/rooms');
+        }
```
farhatahmad, line 22: This is kind of weird. Can we not do all of this logic in the `onSuccess` if you just add an argument to it? `onSuccess: (data) ...`

app/javascript/main.jsx
```diff
@@ -52,6 +53,7 @@ const root = (
             <Route path="/rooms" element={<Rooms />} />
             <Route path="/rooms/:friendlyId" element={<Room />} />
             <Route path="/rooms/:friendlyId/join" element={<RoomJoin />} />
+            <Route path="/main-page" element={<MainPage />} />
```
farhatahmad, line 56: I would call it `/home`

spec/controllers/recordings_controller_spec.rb
farhatahmad: Yeah, by default users should have `create room` enabled. We can do it in an `after_create` in the factory

farhatahmad (approved):

PR #3847, opened 2022-09-09
Basic implementation of locales

scouillard:
## Description

Basic setup of the i18n library

## Usage
1. In every component that has texts/string, import the react-i18n library
`import { useTranslation } from 'react-i18next';`

2. Call the useTranslation hook
`  const { t } = useTranslation();`

3. Generate the locale string
`<h1 className="mt-4"> { t('welcome_bbb')} </h1>`

4. Add the strings in public/locales/<language>/translation.json
```
{
  "welcome_bbb": "Welcome to BigBlueButton."
}
```

## Testing Steps

- Go to the user profile and toggle the language between 'English' and 'French'
- Navigate to '/' and watch the "Welcome to BBB" message change from english to french.

## TODO

- Figure a smart way to scope/namespace the locales/strings 
- Complete reading of https://github.com/i18next/i18next-http-backend#introduction
- Add the [Language Detector Library](https://github.com/i18next/i18next-browser-languageDetector#introduction) so that the browser detects the language on the user even if it is not logged in. We can create a `i18next` cookie on session creation for example.

image


app/javascript/App.jsx
```diff
@@ -11,10 +12,15 @@ export default function App() {
   const currentUser = useAuth();
   const pageHeight = currentUser?.signed_in ? 'regular-height' : 'no-header-height';
 
+  // //i18n
+  const { i18n } = useTranslation();
+  useEffect(() => {
+    i18n.changeLanguage(currentUser?.language);
+  }, [currentUser]);
```
farhatahmad, line 19: Shouldn't this be `currentUser?.language'? That way, it only changes if the user's language changes?

app/javascript/main.jsx
```diff
@@ -31,31 +32,33 @@ const queryClient = new QueryClient();
...
-            <Route path="/admin/roles" element={<Roles />} />
-            <Route path="/admin/roles/edit/:roleId" element={<EditRole />} />
-            <Route path="/rooms" element={<Rooms />} />
-            <Route path="/rooms/:friendlyId" element={<Room />} />
-            <Route path="/rooms/:friendlyId/join" element={<RoomJoin />} />
-            <Route path="*" element={<h1 className="text-center">404</h1>} />
-          </Route>
-        </Routes>
-      </Router>
+      <React.Suspense fallback="Loading...">
```
farhatahmad, line 35: What's this change for?

scouillard: React Suspense is to be used with React "lazy-loading", which is not something we are using right now. 
Removed React Suspense from the Router tree.

farhatahmad (approved):

PR #3851, opened 2022-09-12
CreateRoom Home page front-end

hadicheaito1:
## Description

- Front-end for new Home page

image

### DIFF
--- a/app/javascript/components/home/Home.jsx
+++ b/app/javascript/components/home/Home.jsx
@@ -1,8 +1,27 @@
 import React from 'react';
+import { Card, Stack } from 'react-bootstrap';
+import Button from 'react-bootstrap/Button';
+import Logo from '../shared_components/Logo';
 
 // This page is shown is the current_user does NOT have CreateRoom permission
 export default function Home() {
   return (
-    <>This page is for any user who doesnt have CreateRoom permission</>
+    <div className="mt-5">
+      <div className="text-center mb-4">
+        <Logo size="large" />
+      </div>
+      <Card className="col-md-8 mx-auto p-5 border-0 shadow-sm text-center">
+        <h1 className="mt-4"> Welcome to BigBlueButton. </h1>
+        <div className="mt-4 px-xxl-5">
+          <div className="text-start">
+            <h6> Please enter the URL of your meeting. </h6>
+            <Stack direction="horizontal">
+              <input className="form-control" id="joinUrl" />
+              <Button variant="brand" className="ms-2">Join</Button>
+            </Stack>
+          </div>
+        </div>
+      </Card>
+    </div>
   );
 }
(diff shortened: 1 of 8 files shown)


app/javascript/components/home/Home.jsx
```diff
@@ -1,8 +1,27 @@
...
+      <div className="text-center mb-4">
+        <Logo size="large" />
+      </div>
+      <Card className="col-md-8 mx-auto p-5 border-0 shadow-sm text-center">
+        <h1 className="mt-4"> Welcome to BigBlueButton. </h1>
+        <div className="mt-4 px-xxl-5">
+          <div className="text-start">
+            <h6> Please enter the URL of your meeting. </h6>
+            <Stack direction="horizontal">
+              <input className="form-control" id="joinUrl" />
```
KH-Amir-TN, line 19: If it's possible, it's better to use the form template.
If it's not necessary we could use `react-bootstrap`, we're using native bootstrap when adding classes manually to native html elements.
It's not a big issue, but from best practices perspective it's good that we use `react-bootstrap` components like `Form.Control`.

KH-Amir-TN (approved):

app/javascript/components/home/HomePage.jsx
```diff
@@ -1,14 +1,30 @@
...
+  // redirect user to correct page based on signed in status and CreateRoom permission
+  useEffect(
+    () => {
+      if (currentUser.signed_in && currentUser.permissions.CreateRoom === 'true') {
+        navigate('/rooms');
+      } else if (currentUser.signed_in && currentUser.permissions.CreateRoom === 'false') {
+        navigate('/home');
+      }
+    },
+    [],
```
scouillard, line 26: I would put `currentUser.signed_in` in the brackets (I think it's called an anchor?)

KH-Amir-TN: Good catch @scouillard !
This will introduce problems for users who had their permissions changed and haven't reloaded the tab.
We should set the dependency array to something like `[currentUser.signed_in, currentUser.permissions.CreateRoom]`.
However, be aware of  possible `null|undefined` cases (TypeError: Cannot read properties of `undefined|null` ...).

app/javascript/components/shared_components/Header.jsx
```diff
@@ -45,7 +45,7 @@ export default function Header() {
             }
             <NavDropdown.Item as={Link} to="/">Need help?</NavDropdown.Item>
             <NavDropdown.Divider />
-            <NavDropdown.Item as={Link} to="/" onClick={deleteSession.mutate}>Sign Out</NavDropdown.Item>
```
hadicheaito1, line 48: This had to change due to a bug when singed in user would sign out.
The link would redirect to the `HomePage` before the mutation for signing out the user was even done, this was problematic since it would go into the `useEffect` in `HomePage` and still think that the user is signed in, then redirecting them to the wrong page.

We instead are going to redirect in the mutation and await the query to finish before redirecting. This way the `currentUser` in `HomePage` will have the right information

PR #3852, opened 2022-09-12
Add Access Code checker

scouillard:
## Description

Update the `create_meeting_options` method in Room to check if the `glViewerAccessCode` and `glModeratorAccessCode `configs are enabled. If so, the upcoming created Rooms will have a generated access code.

## Testing Steps

- Enable or disable the access codes options in Admin/Room Configuration
- Create a new room
- Go to Room/Room Settings
- If enabled, the access code(s) should have been generated

### DIFF
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -16,8 +16,6 @@ module Api
         ensure_authorized('ManageRooms', friendly_id: params[:friendly_id])
       end
 
-      after_action :access_code_checker, only: :create
-
       # GET /api/v1/rooms.json
       def index
         # Return the rooms that belong to current user
@@ -58,6 +56,7 @@ module Api
         return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
         if @new_room.save
+          access_code_checker
           logger.info "room(friendly_id):#{@new_room.friendly_id} created for user(id):#{@new_room.user_id}"
           render_data status: :created
         else
@@ -108,7 +107,7 @@ module Api
         params.require(:room).permit(:name, :user_id, :presentation)
       end
 
-      # Generates an access code on room create if the access code room configuration is enabled
+      # Generates an access code on rooms#create if the access code room configuration is enabled
       def access_code_checker
         access_code_settings = RoomSettingsGetter.new(
           room_id: @new_room.id,
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -51,13 +51,13 @@ module Api
       def create
         # TODO: amir - ensure accessibility for authenticated requests only.
         # The created room will be the current user's unless a user_id param is provided with the request.
-        @new_room = Room.create(name: room_params[:name], user_id: room_params[:user_id])
+        room = Room.create(name: room_params[:name], user_id: room_params[:user_id])
 
         return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
-        if @new_room.save
-          generate_access_code
-          logger.info "room(friendly_id):#{@new_room.friendly_id} created for user(id):#{@new_room.user_id}"
+        if room.save
+          room.create_meeting_options(current_provider)
+          logger.info "room(friendly_id):#{room.friendly_id} created for user(id):#{room.user_id}"
           render_data status: :created
         else
           render_error status: :bad_request
@@ -105,23 +105,6 @@ module Api
 
       def room_params
         params.require(:room).permit(:name, :user_id, :presentation)
-      end
-
-      # Generates an access code on rooms#create if the access code room configuration is enabled
-      def generate_access_code
-        access_code_settings = RoomSettingsGetter.new(
-          room_id: @new_room.id,
-          provider: current_provider,
-          current_user:,
-          show_codes: false,
-          settings: %w[glViewerAccessCode glModeratorAccessCode]
-        ).call
-
-        access_code_settings.each do |access_code_setting|
-          RoomMeetingOption.joins(:meeting_option)
-                           .where(room: @new_room, meeting_option: { name: access_code_setting })
-                           .update(value: SecureRandom.alphanumeric(6).downcase)
-        end
       end
     end
   end
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -56,7 +56,6 @@ module Api
         return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
         if room.save
-          room.create_meeting_options(current_provider)
           logger.info "room(friendly_id):#{room.friendly_id} created for user(id):#{room.user_id}"
           render_data status: :created
         else
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -110,11 +110,13 @@ module Api
 
       # Generates an access code on room create if the access code room configuration is enabled
       def access_code_checker
-        access_code_settings = MeetingOption.joins(:rooms_configurations)
-                                            .where(rooms_configurations: { provider: current_provider, value: 'true' })
-                                            .pluck(:name, :value)
-                                            .to_h
-                                            .slice('glViewerAccessCode', 'glModeratorAccessCode')
+        access_code_settings = RoomSettingsGetter.new(
+          room_id: @new_room.id,
+          provider: current_provider,
+          current_user:,
+          show_codes: false,
+          settings: %w[glViewerAccessCode glModeratorAccessCode]
+        ).call
 
         access_code_settings.each do |access_code_setting|
           room_meeting_option = RoomMeetingOption.joins(:meeting_option).where(room: @new_room, meeting_option: { name: access_code_setting })
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -16,6 +16,8 @@ module Api
         ensure_authorized('ManageRooms', friendly_id: params[:friendly_id])
       end
 
+      after_action :access_code_checker, only: :create
+
       # GET /api/v1/rooms.json
       def index
         # Return the rooms that belong to current user
@@ -51,13 +53,12 @@ module Api
       def create
         # TODO: amir - ensure accessibility for authenticated requests only.
         # The created room will be the current user's unless a user_id param is provided with the request.
-        # user_id = room_params[:user_id]
-        room = Room.create(name: room_params[:name], user_id: room_params[:user_id])
+        @new_room = Room.create(name: room_params[:name], user_id: room_params[:user_id])
 
         return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
-        if room.save
-          logger.info "room(friendly_id):#{room.friendly_id} created for user(id):#{room.user_id}"
+        if @new_room.save
+          logger.info "room(friendly_id):#{@new_room.friendly_id} created for user(id):#{@new_room.user_id}"
           render_data status: :created
         else
           render_error status: :bad_request
@@ -105,6 +106,22 @@ module Api
 
       def room_params
         params.require(:room).permit(:name, :user_id, :presentation)
+      end
+
+      # Generates an access code on room create if the access code room configuration is enabled
+      def access_code_checker
+        room_settings = MeetingOption.joins(:rooms_configurations)
+                                     .where(rooms_configurations: { provider: current_provider })
+                                     .pluck(:name, :value)
+                                     .to_h
+        access_code_settings = []
+        access_code_settings << 'glModeratorAccessCode' if room_settings['glModeratorAccessCode'] == 'true'
+        access_code_settings << 'glViewerAccessCode' if room_settings['glViewerAccessCode'] == 'true'
+
+        access_code_settings.each do |access_code_setting|
+          room_meeting_option = RoomMeetingOption.joins(:meeting_option).where(room: @new_room, meeting_option: { name: access_code_setting })
+          room_meeting_option.update(value: SecureRandom.alphanumeric(6).downcase)
+        end
       end
     end
   end
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -56,7 +56,7 @@ module Api
         return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
         if @new_room.save
-          access_code_checker
+          generate_access_code
           logger.info "room(friendly_id):#{@new_room.friendly_id} created for user(id):#{@new_room.user_id}"
           render_data status: :created
         else
@@ -108,7 +108,7 @@ module Api
       end
 
       # Generates an access code on rooms#create if the access code room configuration is enabled
-      def access_code_checker
+      def generate_access_code
         access_code_settings = RoomSettingsGetter.new(
           room_id: @new_room.id,
           provider: current_provider,
@@ -118,8 +118,9 @@ module Api
         ).call
 
         access_code_settings.each do |access_code_setting|
-          room_meeting_option = RoomMeetingOption.joins(:meeting_option).where(room: @new_room, meeting_option: { name: access_code_setting })
-          room_meeting_option.update(value: SecureRandom.alphanumeric(6).downcase)
+          RoomMeetingOption.joins(:meeting_option)
+                           .where(room: @new_room, meeting_option: { name: access_code_setting })
+                           .update(value: SecureRandom.alphanumeric(6).downcase)
         end
       end
     end
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -110,13 +110,11 @@ module Api
 
       # Generates an access code on room create if the access code room configuration is enabled
       def access_code_checker
-        room_settings = MeetingOption.joins(:rooms_configurations)
-                                     .where(rooms_configurations: { provider: current_provider })
-                                     .pluck(:name, :value)
-                                     .to_h
-        access_code_settings = []
-        access_code_settings << 'glModeratorAccessCode' if room_settings['glModeratorAccessCode'] == 'true'
-        access_code_settings << 'glViewerAccessCode' if room_settings['glViewerAccessCode'] == 'true'
+        access_code_settings = MeetingOption.joins(:rooms_configurations)
+                                            .where(rooms_configurations: { provider: current_provider, value: 'true' })
+                                            .pluck(:name, :value)
+                                            .to_h
+                                            .slice('glViewerAccessCode', 'glModeratorAccessCode')
 
         access_code_settings.each do |access_code_setting|
           room_meeting_option = RoomMeetingOption.joins(:meeting_option).where(room: @new_room, meeting_option: { name: access_code_setting })
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -16,7 +16,6 @@ class Room < ApplicationRecord
   validates :meeting_id, presence: true, uniqueness: true
 
   before_validation :set_friendly_id, :set_meeting_id, on: :create
-  after_create :create_meeting_options
 
   attr_accessor :shared, :active, :participants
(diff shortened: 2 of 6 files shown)


scouillard:
Tried to use the RoomSettingsGetter service, but it does not work with AccessCode because they are different than the other settings - they return an empty string or the access code instead of a boolean.

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -16,6 +16,8 @@ class RoomsController < ApiController
         ensure_authorized('ManageRooms', friendly_id: params[:friendly_id])
       end
 
+      after_action :access_code_checker, only: :create
```
farhatahmad, line 19: I'm not sure I like the `after_action` / `after_create`. I'd prefer if we move this logic directly into the `create` action instead

farhatahmad: Since you're depending on the action initializing a `@new_room` for you

scouillard: Yeah, well this would solve my current issue with Rspec. I'm having a hard time accessing the after_action in specs.

scouillard:
Not super proud of the specs.
It is working but I would like to find a better solution for this.

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -106,6 +106,22 @@ def find_room
       def room_params
         params.require(:room).permit(:name, :user_id, :presentation)
       end
+
+      # Generates an access code on rooms#create if the access code room configuration is enabled
+      def access_code_checker
```
farhatahmad, line 111: I would call this `generate_access_code` or something like that

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -106,6 +106,22 @@ def find_room
...
+          room_id: @new_room.id,
+          provider: current_provider,
+          current_user:,
+          show_codes: false,
+          settings: %w[glViewerAccessCode glModeratorAccessCode]
+        ).call
+
+        access_code_settings.each do |access_code_setting|
+          room_meeting_option = RoomMeetingOption.joins(:meeting_option).where(room: @new_room, meeting_option: { name: access_code_setting })
+          room_meeting_option.update(value: SecureRandom.alphanumeric(6).downcase)
```
farhatahmad, line 122: You can do this in one line 
```
RoomMeetingOption
  .joins(:meeting_option)
  .where(room: @new_room, meeting_option: { name: access_code_setting })
  .update(value: SecureRandom.alphanumeric(6).downcase)
```

spec/controllers/rooms_controller_spec.rb
```diff
@@ -173,6 +173,26 @@
...
+        allow(fake_room_settings_getter).to receive(:call).and_return(
+          { 'glViewerAccessCode' => true, 'glModeratorAccessCode' => false }
+        )
+        # can we avoid hardcoding the friendly id?
+        allow(SecureRandom).to receive(:alphanumeric).and_return('8ac31fcc-b550-48f3-a1cd-eab7ec413f35', 'abc123')
+      end
+
+      it 'creates a room with pre-generated access codes' do
+        post :create, params: room_params
+        expect(Room.last.room_meeting_options[0].value).to eq('abc123') # there is a better way to access the room resource
```
farhatahmad, line 192: Just to be 100% safe, you should look up the Room based on the room_params (by user_id and room name)

spec/controllers/rooms_controller_spec.rb
```diff
@@ -173,6 +173,26 @@
...
+      let!(:meeting_option) { create(:meeting_option, name: 'glViewerAccessCode') }
+      let!(:room_meeting_option) { create(:room_meeting_option, meeting_option:) }
+
+      before do
+        allow(RoomSettingsGetter).to receive(:new).and_return(fake_room_settings_getter)
+        allow(fake_room_settings_getter).to receive(:call).and_return(
+          { 'glViewerAccessCode' => true, 'glModeratorAccessCode' => false }
+        )
+        # can we avoid hardcoding the friendly id?
+        allow(SecureRandom).to receive(:alphanumeric).and_return('8ac31fcc-b550-48f3-a1cd-eab7ec413f35', 'abc123')
```
farhatahmad, line 187: You don't need this check - you can just check that it's not blank below

spec/controllers/rooms_controller_spec.rb
```diff
@@ -173,6 +173,26 @@
       expect(response).to have_http_status(:forbidden)
     end
 
+    context 'the access code meeting option is enabled' do
+      let(:fake_room_settings_getter) { instance_double(RoomSettingsGetter) }
+      let!(:meeting_option) { create(:meeting_option, name: 'glViewerAccessCode') }
+      let!(:room_meeting_option) { create(:room_meeting_option, meeting_option:) }
```
farhatahmad, line 179: Why are these needed if you're stubbing the RoomSettingsGetter?

spec/controllers/rooms_controller_spec.rb
scouillard: Great idea - thank you

spec/controllers/rooms_controller_spec.rb
scouillard: The RoomMeetingOption factory is doing `value { [true, false, '12345'].sample }`
I will see if I can play with it without breaking the other specs.

spec/controllers/rooms_controller_spec.rb
scouillard: The RoomSettingsGetter will return the settings values, but AFAIK, it doesn't actually create the RoomMeetingOption and MeetingOption? I will double check.

spec/controllers/rooms_controller_spec.rb
farhatahmad: I'm not sure what you mean. The value in the db is not relevant since you're stubbing the RoomSettingsGetter response anyways.

scouillard: Maybe there is something I didn't get, but the RoomSettingsGetter returns a true or false depending if the access_code option is enabled or not. 

If I want to get the actual value of the access_code, the string value, I have to get it elsewhere.

In the specs, and in the room#generate_access_code, I will find the value by matching the RoomMeetingOption/MeetingOption with the room id and the name of the meeting option.

spec/controllers/rooms_controller_spec.rb
farhatahmad: You shouldn't need to - the `generate_access_code` method should create the RoomMeetingOption for you

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -51,13 +51,13 @@ def public_show
...
-        # user_id = room_params[:user_id]
-        room = Room.create(name: room_params[:name], user_id: room_params[:user_id])
+        @new_room = Room.create(name: room_params[:name], user_id: room_params[:user_id])
 
         return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
-        if room.save
-          logger.info "room(friendly_id):#{room.friendly_id} created for user(id):#{room.user_id}"
+        if @new_room.save
+          generate_access_code
```
hadicheaito1, line 59: Wouldn't it be better to just create a `room` and pass in that room to the `generate_access_code` function instead of creating a global var `new_room` ?

scouillard: Yes, passing the room as an argument would be better.
I can't recall why I did it that way.
This whole PR is getting a big revamp soon - the logic is being put inside the Room model instead of in the controller.

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -51,13 +51,13 @@ def public_show
...
         # The created room will be the current user's unless a user_id param is provided with the request.
-        # user_id = room_params[:user_id]
-        room = Room.create(name: room_params[:name], user_id: room_params[:user_id])
+        @new_room = Room.create(name: room_params[:name], user_id: room_params[:user_id])
 
         return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
-        if room.save
-          logger.info "room(friendly_id):#{room.friendly_id} created for user(id):#{room.user_id}"
+        if @new_room.save
```
KH-Amir-TN, line 58: The room should be already created,
I think a test like `persisted?` is more adequate.

KH-Amir-TN: If it's feasible using `create!`  or changing the logic to first instantiate a room model `new` and then `save` it  may help

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -106,6 +106,23 @@ def find_room
       def room_params
         params.require(:room).permit(:name, :user_id, :presentation)
       end
+
+      # Generates an access code on rooms#create if the access code room configuration is enabled
+      def generate_access_code
```
KH-Amir-TN, line 111: I don't think that is indeed generating access codes if they are forced enabled on rooms creation,
I think it's auto generating access codes for created rooms all time,

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -106,6 +106,23 @@ def find_room
...
+        access_code_settings = RoomSettingsGetter.new(
```
KH-Amir-TN, line 112: Besides, I believe this as is isn't doing anything except making an extra query

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -106,6 +106,23 @@ def find_room
...
+          room_id: @new_room.id,
+          provider: current_provider,
+          current_user:,
+          show_codes: false,
+          settings: %w[glViewerAccessCode glModeratorAccessCode]
+        ).call
+
+        access_code_settings.each do |access_code_setting|
```
KH-Amir-TN, line 120: I believe that the `access_code_settings` will always be `%w[glViewerAccessCode  glModeratorAccessCode]`,

KH-Amir-TN: Besides, all points mentioned above.
Using `Hash#each_key` seems to be more adequate since not using the values,

KH-Amir-TN (changes requested):
I think this is a model concern and can be further simplified and optimized if made as part of the `Room#create_meeting_options`.
We can generate access codes on the run as we create the access codes room meeting options, if the access code room config is forced enabled then we generate and set it on creation otherwise we default it.

scouillard:
> I think this is a model concern and can be further simplified and optimized if made as part of the `Room#create_meeting_options`. We can generate access codes on the run as we create the access codes room meeting options, if the access code room config is forced enabled then we generate and set it on creation otherwise we default it.

Absolutely,

I learnt about create_meeting_options yesterday.
I am almost done revamping the PR, moving the logic from the controller to the model - I'm sorry I should have mentioned it.
I am learning from your PR reviews so it is not in vain.

app/controllers/api/v1/rooms_controller.rb
KH-Amir-TN: That's odd, I tested either.
Maybe, I missed something.

app/models/room.rb
```diff
@@ -37,6 +36,21 @@ def get_setting(name:)
                         .find_by(meeting_option: { name: })
   end
 
+  # Autocreate all meeting options using the default values
+  def create_meeting_options(provider)
+    rooms_configs = MeetingOption.joins(:rooms_configurations)
```
KH-Amir-TN, line 41: I think if we extend `MeetingOption#get_config_value` to fetch single/multiple room configs by name and provider will reduce this boilerplate code and we can leverage it throughout the application.

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -51,12 +51,12 @@ def public_show
       def create
         # TODO: amir - ensure accessibility for authenticated requests only.
         # The created room will be the current user's unless a user_id param is provided with the request.
-        # user_id = room_params[:user_id]
         room = Room.create(name: room_params[:name], user_id: room_params[:user_id])
 
         return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
         if room.save
+          room.create_meeting_options(current_provider)
```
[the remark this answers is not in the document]
scouillard: I couldn't find a solution to access the Provider inside the model

spec/factories/meeting_option_factory.rb
```diff
@@ -2,7 +2,7 @@
 
 FactoryBot.define do
   factory :meeting_option do
-    name { Faker::Vehicle.unique.make }
+    name { Faker::Address.unique.street_address }
```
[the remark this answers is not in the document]
scouillard: Faker was throwing an error.
There is not enough `Vehicule.make`, which is a problem because of the unique tag.
I double checked and it doesn't seem like we are creating a crazy amount of meeting_options.

app/models/room.rb
scouillard: I'll work on this today

app/models/room.rb
```diff
@@ -37,6 +36,21 @@ def get_setting(name:)
                         .find_by(meeting_option: { name: })
   end
 
+  # Autocreate all meeting options using the default values
+  def create_meeting_options(provider)
```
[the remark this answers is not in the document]
scouillard: I didn't think about using the room owner provider, good call.

spec/factories/meeting_option_factory.rb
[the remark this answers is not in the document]
scouillard: @KH-Amir-TN has suggested to clear the cache.
It seems like the uniques stored in cache does not get cleared in between spec groups.
`Vehicle.unique.make` is also used in `room_settings_getter_spec.rb`

app/models/room.rb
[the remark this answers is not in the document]
scouillard: Thank you Amir for the great solution!

app/models/room.rb
scouillard: Done! `get_config_value` returns a hash with the configs values.

spec/factories/meeting_option_factory.rb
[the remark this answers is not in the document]
scouillard: I think using Lorem.word is more adequate than Vehicle.make and should solve are our unique scarcity issues.

spec/controllers/rooms_controller_spec.rb
```diff
@@ -173,6 +173,11 @@
       expect(response).to have_http_status(:forbidden)
     end
 
+    it 'calls create_meeting_options on room' do
```
KH-Amir-TN, line 176: I think we can omit this if we're testing the callback in the room model spec.

app/controllers/api/v1/room_settings_controller.rb
```diff
@@ -15,15 +15,17 @@ def show
...
-        config = MeetingOption.get_config_value(name: room_setting_params[:settingName], provider: current_provider)&.value
-        return render_error status: :bad_request unless config
-
         name = room_setting_params[:settingName]
         value = room_setting_params[:settingValue].to_s
 
+        config = MeetingOption.get_config_value(name:, provider: current_provider)
+        config_value = config[name]
+
+        return render_error status: :bad_request unless config_value
```
KH-Amir-TN, line 24: If we're adding this then I think we may need to add a test for this in the `room_settings_controller_spec`.

spec/controllers/room_settings_controller_spec.rb
```diff
@@ -41,6 +41,7 @@
 
   describe '#update' do
     it 'uses MeetingOption::get_config_value and updates the setting if its config is "optional"' do
+      expect(MeetingOption).to receive(:get_config_value).and_call_original
```
KH-Amir-TN, line 44: Is this required?

spec/models/meeting_option_spec.rb
```diff
@@ -9,7 +9,7 @@
     it { is_expected.to validate_uniqueness_of(:name) }
   end
 
-  describe 'classs methods' do
+  describe 'class methods' do
```
KH-Amir-TN, line 12: My bad xD

spec/models/meeting_option_spec.rb
```diff
@@ -40,15 +40,15 @@
           create(:rooms_configuration, meeting_option:, provider: 'greenlight', value: 'optional')
 
           rooms_config = described_class.get_config_value(name: 'setting', provider: 'greenlight')
-          expect(rooms_config.value).to eq('optional')
+          expect(rooms_config['setting']).to eq('optional')
```
KH-Amir-TN, line 43: Um, nice!
We may even test this multi configs fetching capability?

spec/models/room_spec.rb
```diff
@@ -54,12 +54,41 @@
     end
   end
 
-  describe 'after_create' do
```
KH-Amir-TN, line 57: Why we're changing the structure?

spec/models/meeting_option_spec.rb
KH-Amir-TN, line 43: Um, nice!
We may even test this multi configs fetching capability?

spec/models/room_spec.rb
```diff
@@ -54,12 +54,41 @@
...
-        expect { create(:room) }.to change(RoomMeetingOption, :count).from(0).to(5)
+    before do
+      create_list(:meeting_option, 5)
+    end
+
+    it 'creates a RoomMeetingOption for each MeetingOption' do
+      expect { create(:room) }.to change(RoomMeetingOption, :count).from(0).to(6)
+    end
+
+    it 'does not generate an access code if the room config is not enabled' do
```
KH-Amir-TN, line 68: Access code feels like a test group, we can group those examples under a `context` maybe?

spec/controllers/room_settings_controller_spec.rb
scouillard: It is required because `get_config_value` is called in the `after_create create_meeting_options` callback in Room.

spec/models/meeting_option_spec.rb
scouillard: Isn't it being tested already since we are returning multi configs in `create_meeting_options` and a single config in `room_settings#update` ?

spec/models/room_spec.rb
scouillard: Ah good catch, it's because we removed the `after_create callback` and then we put it back.

app/controllers/api/v1/room_settings_controller.rb
scouillard: I'm not sure what you want to test

spec/models/room_spec.rb
scouillard: I put the after_create context, and upon review I dont think an extra context is necessary

app/controllers/api/v1/room_settings_controller.rb
KH-Amir-TN: The case where the action will return a `bad_request` response to the client.
Like with the unfound 404 tests.

spec/controllers/room_settings_controller_spec.rb
KH-Amir-TN: Um, I see what you mean.
However, I think we can further clean it.
If you intentionally wanna ensure that the method gets called twice you can assert that more specifically rather being so generic (that brakes a bit the resiliency of the tests).
Example instead of:
` expect(MeetingOption).to receive(:get_config_value).and_call_original`
This is more specific:
` expect(MeetingOption).to receive(:get_config_value).with(name: %w[glViewerAccessCode glModeratorAccessCode], provider: 'greenlight'| room.user.provider).and_call_original`
That ensures that if called that should be coming from the `Room#create_meeting_options`.
You can also stub the `Room#create_meeting_options` itself to isolate your test if it's implementation details are not relevant to this test. 
There's also a lot of what you can leverage from the Rspec mocking API that can better answer your need for this matter,

spec/models/meeting_option_spec.rb
KH-Amir-TN: Are we testing that in the `create_meeting_options` tests?
However, that's an integration test I think we may like to initially target that in its unit test catching all cases,
In example, in the `create_meeting_options` test you were only creating a `glViewerAccessCode` meeting option record only so to ensure that the multi configs fetching is working is only because we know the implementation of the actual method,

app/controllers/api/v1/room_settings_controller.rb
scouillard: Seems to be unrelated to this PR

KH-Amir-TN (approved):

KH-Amir-TN (approved):

farhatahmad (approved):

PR #3857, opened 2022-09-14
Password complexity: Added password complexity checks to front end.

KH-Amir-TN:
+ Added password complexity checks to validations schemas on the front end.
+ Changed the Greenlight documentation URL on the Footer.

## Testing Steps

> 1. Pull the code.

image
image

### DIFF
--- a/app/javascript/helpers/forms/ChangePwdFormHelpers.jsx
+++ b/app/javascript/helpers/forms/ChangePwdFormHelpers.jsx
@@ -4,12 +4,22 @@ import { yupResolver } from '@hookform/resolvers/yup';
 const validationSchema = yup.object({
   // TODO: amir - Revisit validations.
   old_password: yup.string().required('Please enter your current password.'),
-  new_password: yup.string().required('Please enter your new password.').min(8, 'Password must have at least 8 characters.'),
-  password_confirmation: yup.string().oneOf([yup.ref('new_password')], 'Your passwords do not match.'),
+  new_password: yup.string()
+    .matches(
+      /^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*[`@%~!#£$\\^&*()\][+={}/|:;"'<>\-,.?_ ]).{8,}$/,
+      'Password must have at least:',
+    )
+    .min(8, '- Eight characters.')
+    .test('oneLower', '- One lowercase letter.', (pwd) => pwd.match(/[a-z]/))
+    .test('oneUpper', '- One uppercase letter.', (pwd) => pwd.match(/[A-Z]/))
+    .test('oneDigit', '- One digit.', (pwd) => pwd.match(/\d/))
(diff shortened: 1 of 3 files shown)


app/javascript/helpers/forms/ChangePwdFormHelpers.jsx
```diff
@@ -4,12 +4,22 @@ import { yupResolver } from '@hookform/resolvers/yup';
 const validationSchema = yup.object({
   // TODO: amir - Revisit validations.
   old_password: yup.string().required('Please enter your current password.'),
-  new_password: yup.string().required('Please enter your new password.').min(8, 'Password must have at least 8 characters.'),
-  password_confirmation: yup.string().oneOf([yup.ref('new_password')], 'Your passwords do not match.'),
+  new_password: yup.string()
```
KH-Amir-TN, line 7: Here's the strategy:
The first rule that contains the full regex pattern will act in as title and all of the rest of validators will be custom where each will target a portion of the pattern and have a specific error addressing it.
This will enable us to:
1. Avoid having a note or a paragraph in place like with v2.
2. The errors will be specific so reacting to certain errors can be specific in the future.
3. Each will have an error feedback component and therefore the UI will be working as expected (Errors will be nested and can be dealt with like with any other elements).
4. The errors are reactive meaning they will change regarding what the user is writing.

app/javascript/helpers/forms/SignupFormHelpers.jsx
```diff
@@ -5,8 +5,17 @@ const validationSchema = yup.object({
...
+  password: yup.string()
+    .matches(
+      /^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*(\W|_)).{8,}$/,
+      'Password must have at least:',
+    )
+    .min(8, '- Eight charachters.')
+    .test('oneLower', '- One lowercase letter.', (pwd) => pwd.match(/^.*[a-z].*$/))
+    .test('oneUpper', '- One uppercase letter.', (pwd) => pwd.match(/^.*[A-Z].*$/))
+    .test('oneDigit', '- One digit.', (pwd) => pwd.match(/^.*\d.*$/))
+    .test('oneSpecial', '- One none alphanumeric charachter.', (pwd) => pwd.match(/^.*(\W|_).*$/)),
```
farhatahmad, line 17: I would say "one symbol" not "one alphanumeric character"

farhatahmad: I would also remove the `.` at the end of each line

KH-Amir-TN: The regex pattern is `\W|_` that accepts any no alphanumeric charachter so 'é' will be treated as a symbol.

KH-Amir-TN: Should we change the pattern (that's what we avoided on v2) or we keep it as it but we indicate a symbol which is a subset of the accepted combinations?

farhatahmad: I'd prefer we specify symbol rather than alphanumeric character

farhatahmad (approved):

PR #3861, opened 2022-09-15
Solved various security concerns through adding a session_token and session_expiry

farhatahmad:
## Description

Replace `user_id` with `session_token` in the CookieStore

## Testing Steps

This is a little hard to test, the easiest would be through a client that can send API requests (ex Postman)

1- Sign in 
2- Copy session contents to API client
3- Send request to authenticated route (ex `http://localhost:3000/api/v1/rooms.json`)
4- Ensure you get a valid response
5- Sign out with the user
6- Retry request
7- Ensure request fails

### DIFF
--- a/app/controllers/api/v1/sessions_controller.rb
+++ b/app/controllers/api/v1/sessions_controller.rb
@@ -56,7 +56,6 @@ module Api
           }
         end
 
-        session[:user_id] = user.id
         session[:session_token] = user.session_token
         session[:session_expiry] = user.session_expiry
       end
--- a/app/controllers/api/v1/sessions_controller.rb
+++ b/app/controllers/api/v1/sessions_controller.rb
@@ -47,7 +47,6 @@ module Api
         if session_params[:extend_session]
           cookies.encrypted[:_extended_session] = {
             value: {
-              user_id: user.id,
               session_token: user.session_token,
               session_expiry: user.session_expiry
             },
--- a/app/controllers/api/v1/sessions_controller.rb
+++ b/app/controllers/api/v1/sessions_controller.rb
@@ -47,8 +47,7 @@ module Api
         if session_params[:extend_session]
           cookies.encrypted[:_extended_session] = {
             value: {
-              session_token: user.session_token,
-              session_expiry: user.session_expiry
+              session_token: user.session_token
             },
             expires: 7.days,
             httponly: true,
@@ -57,7 +56,6 @@ module Api
         end
 
         session[:session_token] = user.session_token
-        session[:session_expiry] = user.session_expiry
       end
 
       def sign_out
--- a/app/controllers/api/v1/sessions_controller.rb
+++ b/app/controllers/api/v1/sessions_controller.rb
@@ -41,20 +41,29 @@ module Api
       end
 
       def sign_in(user)
+        user.generate_session_token!(extended_session: session_params[:extend_session])
+
         # Creates an extended_session cookie if extend_session is selected in sign in form.
         if session_params[:extend_session]
           cookies.encrypted[:_extended_session] = {
             value: {
-              user_id: user.id
+              user_id: user.id,
+              session_token: user.session_token,
+              session_expiry: user.session_expiry
             },
             expires: 7.days,
-            httponly: true
+            httponly: true,
+            secure: true
           }
         end
+
         session[:user_id] = user.id
+        session[:session_token] = user.session_token
+        session[:session_expiry] = user.session_expiry
       end
 
       def sign_out
+        current_user.generate_session_token!
         reset_session
         cookies.delete :_extended_session
       end
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -32,7 +32,6 @@ module Api
         if user.save
           user.generate_session_token!
           unless current_user # if this is NOT an admin creating a user
-            session[:user_id] = user.id
             session[:session_token] = user.session_token
             session[:session_expiry] = user.session_expiry
           end
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -31,6 +31,9 @@ module Api
 
         if user.save
           session[:user_id] ||= user.id
+          session[:session_token] ||= user.session_token
+          session[:session_expiry] ||= user.session_expiry
+
           token = user.generate_activation_token!
           UserMailer.with(user:, expires_in: User::ACTIVATION_TOKEN_VALIDITY_PERIOD.from_now,
                           activation_url: activate_account_url(token)).activate_account_email.deliver_later
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -31,10 +31,8 @@ module Api
 
         if user.save
           user.generate_session_token!
-          unless current_user # if this is NOT an admin creating a user
-            session[:session_token] = user.session_token
-            session[:session_expiry] = user.session_expiry
-          end
+          session[:session_token] = user.session_token unless current_user # if this is NOT an admin creating a user
+
           token = user.generate_activation_token!
           UserMailer.with(user:, expires_in: User::ACTIVATION_TOKEN_VALIDITY_PERIOD.from_now,
                           activation_url: activate_account_url(token)).activate_account_email.deliver_later
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -30,10 +30,12 @@ module Api
         return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
         if user.save
-          session[:user_id] ||= user.id
-          session[:session_token] ||= user.session_token
-          session[:session_expiry] ||= user.session_expiry
-
+          user.generate_session_token!
+          unless current_user # if this is NOT an admin creating a user
+            session[:user_id] = user.id
+            session[:session_token] = user.session_token
+            session[:session_expiry] = user.session_expiry
+          end
           token = user.generate_activation_token!
           UserMailer.with(user:, expires_in: User::ACTIVATION_TOKEN_VALIDITY_PERIOD.from_now,
                           activation_url: activate_account_url(token)).activate_account_email.deliver_later
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,10 +7,7 @@ class ApplicationController < ActionController::Base
   # Returns the current signed in User (if any)
   def current_user
     # Overwrites the session cookie if an extended_session cookie exists
-    if cookies.encrypted[:_extended_session].present?
-      session[:session_token] ||= cookies.encrypted[:_extended_session]['session_token']
-      session[:session_expiry] ||= cookies.encrypted[:_extended_session]['session_expiry']
-    end
+    session[:session_token] ||= cookies.encrypted[:_extended_session]['session_token'] if cookies.encrypted[:_extended_session].present?
 
     user = User.find_by(session_token: session[:session_token])
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,10 +7,19 @@ class ApplicationController < ActionController::Base
   # Returns the current signed in User (if any)
   def current_user
     # Overwrites the session cookie if an extended_session cookie exists
-    session[:user_id] ||= cookies.encrypted[:_extended_session]['user_id'] if cookies.encrypted[:_extended_session].present?
+    if cookies.encrypted[:_extended_session].present?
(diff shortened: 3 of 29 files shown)


app/controllers/api/v1/sessions_controller.rb
```diff
@@ -41,20 +41,29 @@ def session_params
...
             value: {
-              user_id: user.id
+              user_id: user.id,
+              session_token: user.session_token,
+              session_expiry: user.session_expiry
             },
             expires: 7.days,
-            httponly: true
+            httponly: true,
+            secure: true
```
KH-Amir-TN, line 56: Using the `secure` attribute means that GL3 in a production environment cannot be used unless through HTTPS.

KH-Amir-TN: I understand your intent,
Both of the session cookies are encrypted and signed, if the secret key base isn't compromised users will have no access to plain data but rather a cipher.

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -41,20 +41,29 @@ def session_params
...
           }
         end
+
         session[:user_id] = user.id
+        session[:session_token] = user.session_token
+        session[:session_expiry] = user.session_expiry
       end
 
       def sign_out
+        current_user.generate_session_token!
```
KH-Amir-TN, line 66: Shouldn't we reset the token data?
Isn't the  signin will generate new tokens?

app/models/user.rb
```diff
@@ -36,6 +36,8 @@ class User < ApplicationRecord
   validates :reset_digest, uniqueness: true, if: :reset_digest?
   validates :activation_digest, uniqueness: true, if: :activation_digest?
```
KH-Amir-TN, line 38: Shouldn't we add in model validation for uniqueness?
The database exceptions aren't the same as the `ActiveRecord` ones.

app/models/user.rb
```diff
@@ -75,6 +77,17 @@ def self.reset_token_expired?(sent_at)
...
+  def generate_session_token!(extended_session: false)
+    digest = User.generate_digest(SecureRandom.alphanumeric(40))
+    expiry = extended_session ? 7.days.from_now : 6.hours.from_now
+
+    update! session_token: digest, session_expiry: expiry
+  rescue ActiveRecord::RecordInvalid
+    raise unless errors.attribute_names.include? :session_token
+
+    retry
+  end
```
KH-Amir-TN, line 89: Without the model level validation, I think this will not give the expected outcome.

app/controllers/api/v1/sessions_controller.rb
KH-Amir-TN, line 56: Using the secure attribute could mean that GL3 in a production environment will only be accessed through HTTPS.

app/controllers/api/v1/users_controller.rb
```diff
@@ -31,6 +31,9 @@ def create
 
         if user.save
           session[:user_id] ||= user.id
+          session[:session_token] ||= user.session_token
```
KH-Amir-TN, line 34: We're storing the token or the hash on the session itself?
I think from security perspective the token shouldn't be logged or stored, it will be transmitted securely to the client which will provide in upon requesting.

app/controllers/application_controller.rb
```diff
@@ -8,7 +8,10 @@ class ApplicationController < ActionController::Base
   def current_user
     # Overwrites the session cookie if an extended_session cookie exists
     session[:user_id] ||= cookies.encrypted[:_extended_session]['user_id'] if cookies.encrypted[:_extended_session].present?
-    @current_user ||= User.find_by(id: session[:user_id])
+
+    user = User.find_by(id: session[:user_id])
+    user = nil unless valid_session?(user)
+    @current_user ||= user
```
KH-Amir-TN, line 14: There's a small possibility where a request that should have its session ended will circumvent the check,
The `@current_user` is memoizing the first none `nil` value so **L13** will have only an effect on fresh requests.

spec/controllers/admin/server_recordings_controller_spec.rb
```diff
@@ -9,7 +9,8 @@
 
   before do
     request.headers['ACCEPT'] = 'application/json'
-    session[:user_id] = user_with_manage_recordings_permission.id
+    # session[:user_id] = user_with_manage_recordings_permission.id
```
KH-Amir-TN, line 12: Is this comment required?

app/controllers/application_controller.rb
KH-Amir-TN: Besides, catching an invalid session should invalidate (reset) it to reduce error surface.

app/controllers/api/v1/users_controller.rb
```diff
@@ -31,6 +31,9 @@ def create
...
+          session[:session_expiry] ||= user.session_expiry
```
KH-Amir-TN, line 35: This block is dedicated for signing in, so it should be able to use the same logic for signin,
The keeping of already existed session data is error prone,
Example, because of the lack of reset to invalid sessions a user that got logged out because of stale session data may have an error when creating an account.

app/models/user.rb
```diff
@@ -36,6 +36,8 @@ class User < ApplicationRecord
   validates :reset_digest, uniqueness: true, if: :reset_digest?
   validates :activation_digest, uniqueness: true, if: :activation_digest?
 
+  after_create :generate_session_token!
```
KH-Amir-TN, line 39: I think it's better to have one source of truth or a unique route to manage session tokens to reduce error surface and simplify the logic.
Managing session tokens feels as a session/controller and not a model concern.

app/controllers/api/v1/sessions_controller.rb
KH-Amir-TN: https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#restrict_access_to_cookies

app/controllers/application_controller.rb
```diff
@@ -8,7 +8,10 @@ class ApplicationController < ActionController::Base
   def current_user
     # Overwrites the session cookie if an extended_session cookie exists
     session[:user_id] ||= cookies.encrypted[:_extended_session]['user_id'] if cookies.encrypted[:_extended_session].present?
-    @current_user ||= User.find_by(id: session[:user_id])
+
+    user = User.find_by(id: session[:user_id])
```
farhatahmad, line 12: User.where(session_token; session[:session_token]).where(DateTime.now > session[:session_expiry]).first

app/controllers/api/v1/sessions_controller.rb
farhatahmad: This is intended

app/models/user.rb
```diff
@@ -75,6 +76,17 @@ def self.reset_token_expired?(sent_at)
     Time.current > (sent_at.in(RESET_TOKEN_VALIDITY_PERIOD))
   end
 
+  def generate_session_token!(extended_session: false)
+    digest = User.generate_digest(SecureRandom.alphanumeric(40))
```
KH-Amir-TN, line 80: Since we'll need only the digest we can leverage a performance boost by using `random_bytes` or `gen_random`?

app/controllers/external_controller.rb
```diff
@@ -13,7 +13,7 @@ def create_user
       u.language = extract_language_code user_info['locale']
     end
 
-    session[:user_id] = user.id
+    session[:session_token] = user.session_token
```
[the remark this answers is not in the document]
KH-Amir-TN: I think for `openid_connect` this is the endpoint for both signup/signin.
So I believe this will fail to authenticate users because it's not generating/renewing session tokens when they authenticate?

app/models/user.rb
KH-Amir-TN: FYI here's a benchmark, for a `500_000` trials for each case on my local setup `gen_random` is approximetly 5.7 times faster:
image

app/controllers/api/v1/sessions_controller.rb
farhatahmad: It will be by default when we set up `force_ssl=true`

app/controllers/application_controller.rb
```diff
@@ -7,8 +7,20 @@ class ApplicationController < ActionController::Base
...
+      session[:session_token] ||= cookies.encrypted[:_extended_session]['session_token']
+      session[:session_expiry] ||= cookies.encrypted[:_extended_session]['session_expiry']
+    end
+
+    user = User.find_by(session_token: session[:session_token])
+
+    if user && invalid_session?(user)
+      reset_session
+      cookies.delete :_extended_session
+      user = nil
```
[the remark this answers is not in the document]
farhatahmad: Hmm yeah let me try that

app/controllers/application_controller.rb
```diff
@@ -27,6 +39,14 @@ def current_provider
 
   private
 
+  # Checks if the user's session_token matches the session and that it is not expired
+  def invalid_session?(user)
```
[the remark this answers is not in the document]
farhatahmad: a `nil` user isn't invalid. It just means that they are unauthenticated

app/controllers/external_controller.rb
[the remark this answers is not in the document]
farhatahmad: Yeah this just needs a generate token before saving to session

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -41,20 +41,27 @@ def session_params
...
           }
         end
-        session[:user_id] = user.id
+
+        session[:session_token] = user.session_token
+        session[:session_expiry] = user.session_expiry
       end
 
       def sign_out
+        current_user.generate_session_token!
```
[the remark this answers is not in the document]
farhatahmad: I'm not sure what you mean here

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -41,20 +41,27 @@ def session_params
       end
 
       def sign_in(user)
+        user.generate_session_token!(extended_session: session_params[:extend_session])
+
         # Creates an extended_session cookie if extend_session is selected in sign in form.
         if session_params[:extend_session]
           cookies.encrypted[:_extended_session] = {
```
scouillard, line 48: I suppose that we are intending to eventually remove the [:_extended_session] cookie since we can manually set up the expiry in User `generate_session_token!(extended_session: false)`?

app/controllers/api/v1/sessions_controller.rb
KH-Amir-TN: Gotcha,

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -41,20 +41,27 @@ def session_params
...
             value: {
-              user_id: user.id
+              session_token: user.session_token,
+              session_expiry: user.session_expiry
             },
             expires: 7.days,
```
KH-Amir-TN, line 53: I think it would be better to extract the expiration periods into a centralized entity like a constant.
The expiration period looks to be used here and in the `User#generate_session_token!` which introduces if not carefully synchronized a good room for error,
However, that does not seem to have a security concern.

app/controllers/api/v1/sessions_controller.rb
farhatahmad: Nope - we need the cookie to store the value on the client side, or else there's no way for us to tell which user is signed in

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -41,20 +41,27 @@ def session_params
...
       def sign_in(user)
+        user.generate_session_token!(extended_session: session_params[:extend_session])
+
         # Creates an extended_session cookie if extend_session is selected in sign in form.
         if session_params[:extend_session]
           cookies.encrypted[:_extended_session] = {
             value: {
-              user_id: user.id
+              session_token: user.session_token,
+              session_expiry: user.session_expiry
```
KH-Amir-TN, line 51: It seems that we're only writing the `:session_expiry` to  but not reading it from the cookie(s),
I think with this design we'd only need `session_digest` only?

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -41,20 +41,27 @@ def session_params
...
             expires: 7.days,
-            httponly: true
+            httponly: true,
+            secure: true
           }
         end
-        session[:user_id] = user.id
+
+        session[:session_token] = user.session_token
+        session[:session_expiry] = user.session_expiry
```
KH-Amir-TN, line 60: Setting the `session_expiry` (which may be not required as mentionned above) will cause the signin logic to differ/diverge from external accounts signin (which does set only the `session_digest`).
However, since the `session_expiry` could be omitted that does not have a critical implication except for enlarging cookies size and its implication (4KB limit + slower requests loading [it's not much for this project scope]).

scouillard (approved):

scouillard (approved):

PR #3903, opened 2022-09-20
Allow get_config_value to return multiple values

scouillard:
## Description

Allow the `get_config_value` method for MeetingOption to return a hash with multiple values.
This is needed as we require to return the values of both `glViewerAccessCode` and `glModeratorAccessCode` in #3852

### DIFF
--- a/app/controllers/api/v1/room_settings_controller.rb
+++ b/app/controllers/api/v1/room_settings_controller.rb
@@ -15,15 +15,17 @@ module Api
 
       # PATCH /api/v1/room_settings/:friendly_id
       def update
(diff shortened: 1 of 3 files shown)


spec/controllers/room_settings_controller_spec.rb
```diff
@@ -41,6 +41,7 @@
 
   describe '#update' do
     it 'uses MeetingOption::get_config_value and updates the setting if its config is "optional"' do
+      expect(MeetingOption).to receive(:get_config_value).and_call_original
```
scouillard, line 44: This is needed because `get_config_value` is called in the `after_create` method `create_meeting_option` in Room.

KH-Amir-TN (approved):

farhatahmad (approved):

PR #3910, opened 2022-09-21
Pass avatar through to BigBlueButton

farhatahmad:
## Description

Passes the avatarUrl into BigBlueButton (see below)
image

## Testing Steps

To test this locally, you NEED to use ngrok 

1- Update your profile picture
2- Start a room and join it
3- Ensure your profile picture in BigBlueButton matches the one in Greenlight

### DIFF
--- a/app/controllers/api/v1/meetings_controller.rb
+++ b/app/controllers/api/v1/meetings_controller.rb
@@ -27,7 +27,12 @@ module Api
           return render_error status: :bad_request unless e.key == 'idNotUnique'
         end
 
-        render_data data: BigBlueButtonApi.new.join_meeting(room: @room, name: current_user.name, role: 'Moderator'), status: :created
+        render_data data: BigBlueButtonApi.new.join_meeting(
+          room: @room,
+          name: current_user.name,
+          avatar_url: current_user.avatar.attached? ? url_for(current_user.avatar) : nil,
+          role: 'Moderator'
+        ), status: :created
       end
 
       # POST /api/v1/meetings/:friendly_id/status.json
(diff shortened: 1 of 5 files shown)


app/controllers/api/v1/meetings_controller.rb
```diff
@@ -27,7 +27,12 @@ def start
           return render_error status: :bad_request unless e.key == 'idNotUnique'
         end
 
-        render_data data: BigBlueButtonApi.new.join_meeting(room: @room, name: current_user.name, role: 'Moderator'), status: :created
+        render_data data: BigBlueButtonApi.new.join_meeting(
+          room: @room,
+          name: current_user.name,
```
KH-Amir-TN, line 32: Can we extend the `join_meeting` interface (method signature) so that like with the room argument takes in a user argument and extract anything it needs from the record itself.
That will remove the need to have some logic done in/pre  the method call like `current_user.avatar.attached? ? url_for(current_user.avatar) : nil` and gives more flexibility in the future and stability to the interface,
Something like `join_meeting(room:, user: role:)` maybe?

spec/controllers/meetings_controller_spec.rb
```diff
@@ -39,7 +39,7 @@
       ).and_call_original
 
       expect_any_instance_of(MeetingStarter).to receive(:call)
-      expect_any_instance_of(BigBlueButtonApi).to receive(:join_meeting).with(room:, name: user.name, role: 'Moderator')
+      expect_any_instance_of(BigBlueButtonApi).to receive(:join_meeting).with(room:, name: user.name, avatar_url: nil, role: 'Moderator')
```
KH-Amir-TN, line 42: Changing the interface can help stabilize the interface and remove the excessive need to update mocks and stubs when updating the method implementation,

KH-Amir-TN (approved):

app/controllers/api/v1/meetings_controller.rb
farhatahmad: Tried that at first and couldn't get it in a clean way. It's tricky because the user is optional, meaning that you have to have these conditionals either way

farhatahmad: I'm going to refactor some of this in the near future to clean up complexity and duplication

PR #3917, opened 2022-09-21
Last Session rooms card and page

hadicheaito1:
## Description

- Replacing created at with last session on rooms card and rooms page under name

image
image
image

### DIFF
--- a/app/javascript/components/rooms/RoomCard.jsx
+++ b/app/javascript/components/rooms/RoomCard.jsx
@@ -40,9 +40,9 @@ export default function RoomCard({ room }) {
             <span className="text-muted">{ t('room.shared_by', { room }) } </span>
           )}
           { room.last_session ? (
-            <span className="text-muted"> Last Session: { room.last_session } </span>
+            <span className="text-muted"> { t('room.last_session', { room }) } </span>
           ) : (
-            <span className="text-muted mt-2"> No previous sessions created </span>
+            <span className="text-muted mt-2"> { t('room.no_last_session') } </span>
           )}
         </Stack>
       </Card.Body>
--- a/app/javascript/components/rooms/RoomCard.jsx
+++ b/app/javascript/components/rooms/RoomCard.jsx
@@ -66,12 +66,18 @@ export default function RoomCard({ room }) {
   );
 }
 
+RoomCard.defaulProps = {
+  room: PropTypes.shape({
+    last_session: '',
+  }),
+};
+
 RoomCard.propTypes = {
   room: PropTypes.shape({
     id: PropTypes.string.isRequired,
     friendly_id: PropTypes.string.isRequired,
     name: PropTypes.string.isRequired,
-    last_session: PropTypes.string.isRequired,
+    last_session: PropTypes.string,
     shared_owner: PropTypes.string,
     active: PropTypes.bool,
     participants: PropTypes.number,
--- a/app/javascript/components/rooms/RoomCard.jsx
+++ b/app/javascript/components/rooms/RoomCard.jsx
@@ -36,10 +36,13 @@ export default function RoomCard({ room }) {
 
         <Stack className="my-4">
           <Card.Title className="mb-0"> { room.name } </Card.Title>
-          { room.shared_owner ? (
+          { room.shared_owner && (
             <span className="text-muted">{ t('room.shared_by', { room }) } </span>
+          )}
+          { room.last_session ? (
+            <span className="text-muted"> Last Session: { room.last_session } </span>
           ) : (
-            <span className="text-muted"> { room.created_at } </span>
+            <span className="text-muted mt-2"> No previous sessions created </span>
           )}
         </Stack>
       </Card.Body>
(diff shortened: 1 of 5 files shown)


app/javascript/components/rooms/RoomCard.jsx
```diff
@@ -32,10 +32,13 @@ export default function RoomCard({ room }) {
 
         <Stack className="my-4">
           <Card.Title className="mb-0"> { room.name } </Card.Title>
-          { room.shared_owner ? (
+          { room.shared_owner && (
             <span className="text-muted">{ t('room.shared_by', { room }) } </span>
+          )}
+          { room.last_session ? (
+            <span className="text-muted"> Last Session: { room.last_session } </span>
```
KH-Amir-TN, line 39: If it's feasible, can we add in localization here?

app/javascript/components/rooms/RoomCard.jsx
```diff
@@ -32,10 +32,13 @@ export default function RoomCard({ room }) {
...
           ) : (
-            <span className="text-muted"> { room.created_at } </span>
+            <span className="text-muted mt-2"> No previous sessions created </span>
```
KH-Amir-TN, line 41: And here?

app/javascript/components/rooms/room/Room.jsx
```diff
@@ -33,7 +33,11 @@ export default function Room() {
       <Row className="my-5">
         <Col>
           <h3>{room.name}</h3>
-          <p className="text-muted"> { room.created_at }</p>
+          { room.last_session ? (
```
KH-Amir-TN, line 36: Same here,

app/serializers/current_room_serializer.rb
```diff
@@ -24,4 +25,8 @@ def owner_name
   def owner_avatar
     user_avatar(object.user)
   end
+
+  def last_session
+    object.last_session.strftime('%A %B %e, %Y %l:%M%P')
```
KH-Amir-TN, line 30: I think this is fine for now, but we need to refer back to this in the future.
Since, that will make it hard for us to localize the datetime and time zone it,
I think a better approach is to serve a timestamp and leave the formatting responsibility to the frontend app.

app/serializers/room_serializer.rb
```diff
@@ -1,10 +1,15 @@
 # frozen_string_literal: true
 
 class RoomSerializer < ApplicationSerializer
-  attributes :id, :name, :friendly_id, :created_at
+  attributes :id, :name, :friendly_id, :last_session
   attribute :shared_owner, if: -> { object.shared }
+  attribute :last_session, if: -> { object.last_session }
```
KH-Amir-TN, line 6: Do we need this conditional rendering?

app/javascript/components/rooms/room/Room.jsx
```diff
@@ -33,7 +33,11 @@ export default function Room() {
...
+            <span className="text-muted"> Last Session: { room.last_session } </span>
+          ) : (
+            <span className="text-muted mt-2"> No previous sessions created </span>
```
KH-Amir-TN, line 39: And here,

app/serializers/room_serializer.rb
hadicheaito1: Yes we need it for the case that the room has `last_session` as `nil` still.

I don't need to add it to line 4 though since I put line 6, therefore I will remove it from line 4 :+1:

app/javascript/components/rooms/RoomCard.jsx
hadicheaito1: Done

hadicheaito1:
Let me know if you guys think we should remove the 'Last Session:' part on the room card.
The reason I added it is because it may not be as obvious what that date is by just looking at the card, but i think it would look better without it

app/javascript/components/rooms/RoomCard.jsx
```diff
@@ -60,7 +63,7 @@ RoomCard.propTypes = {
     id: PropTypes.string.isRequired,
     friendly_id: PropTypes.string.isRequired,
     name: PropTypes.string.isRequired,
-    created_at: PropTypes.string.isRequired,
+    last_session: PropTypes.string.isRequired,
```
KH-Amir-TN, line 66: With the conditional rendering on the serializer, I suspect that this will throw prop types warnings when there's no `last_session`,

KH-Amir-TN (dismissed):

app/serializers/room_serializer.rb
farhatahmad: Why don't we still pass it through and then handle the nil case on the front-end?

farhatahmad: Well actually, you're already handling that case, so does this even make a difference?

farhatahmad:
I think it's important that we put the label so that people know what that date is, but we can shorten the date to make it look a little better

app/javascript/components/rooms/RoomCard.jsx
hadicheaito1: Good catch, done

hadicheaito1:
> I think it's important that we put the label so that people know what that date is, but we can shorten the date to make it look a little better

Yeah agreed. I removed the day of the week to make it shorter

KH-Amir-TN (approved):
# 🚀 🚀 🚀

farhatahmad (approved):

PR #3922, opened 2022-09-22
Fixed active record queries to avoid loading all avatars

farhatahmad:

Query before
image

Query after 
image


scouillard (approved):

scouillard:
`pagy_array` returns an array of all the avatar, and `pagy` will return an array of the page's avatar?

farhatahmad:
Not really - the query above `pagy_array` was returning an array of all users, which `pagy_array` then "fake" paginates into sets of 5. By removing the `to_a`, it now returns an active record collection with can be "real" paginated using `pagy`

PR #3925, opened 2022-09-22
Fix Share Access

scouillard:
## Description

- Fix the Share Access edge case error (backend solution).
- Change the SharableUsers and SharedUsers mutation hooks to use the new template.
- Add the keepPreviousData param to the SharedUsers and ShareableUsers query hooks.

**Problem:** 
If there is only one sharableUser left, the form data will return a single string instead of an array.`"user_id"`
It is a problem because our controller is expecting an array.
It is kind of an edge case because if you select/share with only one user, it will still return an array with a single string. `["user_id"]`.
The bug will happen only if there is _a single user left that is shareable_.
For example, if your local app has only two users (yourself and another one), the shared access feature will not work at all (as yourself is never included in shareable users).

**Solution:**
I went thru React Hook Form docs to see if we could format the form response, which would be the ideal solution, without success.
~The proposed solution is to add an _if checktype_.~
Put the params response inside the ruby Array() method in backend so the server always deals with an array.

### DIFF
--- a/app/javascript/components/rooms/room/shared_access/forms/SharedAccessForm.jsx
+++ b/app/javascript/components/rooms/room/shared_access/forms/SharedAccessForm.jsx
@@ -19,8 +19,7 @@ export default function SharedAccessForm({ handleClose }) {
   const { friendlyId } = useParams();
   const { onSubmit } = useShareAccess({ friendlyId, closeModal: handleClose });
   const [input, setInput] = useState();
-  const [shareableUsers, setShareableUsers] = useState();
-  useShareableUsers(friendlyId, input, setShareableUsers);
+  const { data: shareableUsers } = useShareableUsers(friendlyId, input);
 
   return (
     <div id="shared-access-form">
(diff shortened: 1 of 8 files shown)


scouillard:
Not really happy with this solution.
Plan on spending some time (friday) to explore different options (maybe ditch React Hook Form?) and update the useSharedAccess hook as per the new template.

app/javascript/components/rooms/room/shared_access/forms/SharedAccessForm.jsx
```diff
@@ -42,8 +42,8 @@ export default function SharedAccessForm({ handleClose }) {
                       <Stack direction="horizontal" className="py-2">
                         <Form.Check
                           type="checkbox"
+                          key={user.id}
```
KH-Amir-TN, line 45: What's the use of `key` here?

KH-Amir-TN (dismissed):

hadicheaito1 (dismissed):

app/javascript/components/rooms/room/shared_access/forms/SharedAccessForm.jsx
scouillard: I read that it is recommended to use the key attributes inside arrays
https://reactjs.org/docs/lists-and-keys.html#keys

KH-Amir-TN: Yes, it should be done on the parent component which you have done on #L39 `<tr key={user.id} className="align-middle">`,
The key on the `Form.Check` I believe isn't requireid,

farhatahmad (approved):

PR #3929, opened 2022-09-23
Emails: Added email format validation to `User` model.

KH-Amir-TN:
## Testing Steps

> 1. Pull the code.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -78,6 +78,19 @@ RSpec.describe User, type: :model do
         end
       end
     end
+
+    context 'email format' do
+      it 'accepts valid email format' do
+        user = build(:user, email: 'user-1.dep-1@users.org-1.tld')
+        expect(user).to be_valid
+      end
+
+      it 'refuses invalid email formats' do
+        user = build(:user, email: 'INVALID')
+        expect(user).to be_invalid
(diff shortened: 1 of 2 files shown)


spec/models/user_spec.rb
```diff
@@ -90,6 +90,19 @@
...
+    context 'email format' do
+      it 'accepts valid email format' do
+        user = build(:user, email: 'user@users.com')
+        expect(user).to be_valid
+      end
+
+      it 'refuses invalid email formats' do
+        user = build(:user, email: 'INVALID')
+        expect(user).to be_invalid
+        expect(user.errors.attribute_names).to match_array([:email])
```
KH-Amir-TN, line 103: The matching instead of inclusion is intentional to ensure that only the email attribute is invalid,

Gemfile
```diff
@@ -67,3 +67,5 @@ group :test do
   gem 'shoulda-matchers', '~> 5.0'
   gem 'webdrivers'
 end
+
+gem 'validate_email', '~> 0.1.6'
```
farhatahmad, line 71: Why the gem? I think it would be clearer if we put the regex on the model instead - that way, we can make any changes to it if ever necessary

farhatahmad (approved):

PR #3935, opened 2022-09-26
Add feedback in Room meeting is running

scouillard:
## Description

Add feedback in Room if the meeting is running.
Create an "online pill" as feedback.
Thinking about adding this pill to the Room card in Rooms as well.
image

## Testing Steps

Start/End a meeting and see the change in Room.

### DIFF
--- a/app/javascript/components/rooms/room/Room.jsx
+++ b/app/javascript/components/rooms/room/Room.jsx
@@ -20,12 +20,13 @@ function copyInvite() {
 export default function Room() {
   const { t } = useTranslation();
   const { friendlyId } = useParams();
-  const { isLoading, data: room } = useRoom(friendlyId);
+  const { isLoading: isLoadingRoom, data: room } = useRoom(friendlyId);
   const startMeeting = useStartMeeting(friendlyId);
-  const { isLoadingRunning, data: isRunning } = useMeetingRunning(friendlyId);
+  const { isLoading: isLoadingRunning, data: isRunning } = useMeetingRunning(friendlyId);
 
   if (isLoadingRunning) return <Spinner />; // Todo: amir - Revisit this.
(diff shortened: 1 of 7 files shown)


app/javascript/components/rooms/room/Room.jsx
```diff
@@ -19,7 +22,9 @@ export default function Room() {
   const { friendlyId } = useParams();
   const { isLoading, data: room } = useRoom(friendlyId);
   const startMeeting = useStartMeeting(friendlyId);
+  const { isLoadingRunning, data: isRunning } = useMeetingRunning(friendlyId);
```
KH-Amir-TN, line 25: I suspect that `isLoadingRunning` here is `undefined`?
To rename it I suggest the use of the following syntax `const {isLoading: isLoadingRunning, data: isRunning} = useMeetingRunning(...);`

KH-Amir-TN: Or maybe avoid destructing the returned value and use namespaces `const meetingRunningQuery = useMeetingRunning(...);`

app/javascript/components/rooms/room/Room.jsx
```diff
@@ -31,13 +36,25 @@ export default function Room() {
...
       </Row>
       <Row className="my-5">
-        <Col>
-          <h3>{room.name}</h3>
+        <Col className="col-xxl-8">
+          <Stack direction="horizontal" gap={2}>
+            <h1>{room.name}</h1>
+            { isRunning
+              && (
+                <Badge className="rounded-pill online-pill ms-2 text-success">
```
KH-Amir-TN, line 44: Can we extract this into a shared component?
That will help us re-using it,

scouillard: It is changed into a component in the following PR #3938

farhatahmad (approved):

PR #3937, opened 2022-09-26
Added avatar validations

farhatahmad:
## Description

Added the following restrictions to avatars:
-  max 300px x 300px
- must be one of  [:png, :jpg, :jpeg, :svg]
- smaller than 3mb


scouillard:
`active_storage_validation` seems like a pretty nice gem.

scouillard:
Should we add some Frontend validations?
The AvatarEditor library should take care of the max size of the avatar.
We can pass any type of file to the `useCreateAvatar` hook as it will be converted to a png by default by the toBlob() method (line 13)

scouillard (approved):

KH-Amir-TN (approved):

farhatahmad:
@scouillard Yeah it may be better to add some more friendly messaging - I'll merge as is and create a note for it

PR #3939, opened 2022-09-27
AuthZ (FE): Added `AuthenticatedAndActiveOnly` & `UnauthenticatedOnly` protected routes.

KH-Amir-TN:
## Description

Added AuthenticatedAndActiveOnly & UnauthenticatedOnly protected routes. 
+ Minor refactoring to navigation logic on signin and signup creating a single source of navigation (dispatcher pattern).

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/javascript/contexts/auth/AuthProvider.jsx
+++ b/app/javascript/contexts/auth/AuthProvider.jsx
@@ -26,6 +26,7 @@ export default function AuthProvider({ children }) {
     permissions: currentUser?.permissions,
     role: currentUser?.role,
     active: currentUser?.active,
+    signed_out: currentUser?.signed_out,
   };
 
   const memoizedCurrentUser = useMemo(() => user, [user]);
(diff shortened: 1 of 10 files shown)


scouillard (approved):

KH-Amir-TN:
> I find the syntax slightly odd here - can we not create a component that does whatever checks is needed and then returns `Route` instead? That way `Route` and `ProtectedRoute` are interchangeable?

Unfortunately, that seems to be unfeasible with the current version of react router.
Will provide detailed explanation during today's meeting,

KH-Amir-TN:
This is only an increment,
The complete UX will be achieved once all subsequent PRs gets merged,

app/javascript/contexts/auth/AuthProvider.jsx
```diff
@@ -26,6 +26,7 @@ export default function AuthProvider({ children }) {
     permissions: currentUser?.permissions,
     role: currentUser?.role,
     active: currentUser?.active,
+    signed_out: currentUser?.signed_out,
```
KH-Amir-TN, line 29: We can clean this by introducing a small global state that includes the `currentUser` along with our small client state,
A client state is the state that the client application uses to store data that it needs to retrieve and has no relation with the server data (that includes as an example from where you were navigating, what component triggered the logic, your pointer coordination, scrolling coordination, ....)
So we won't do that hackery of `queryClient.setQueryData(...)`.

PR #3940, opened 2022-09-27
UI/UX: Added `ErrorBoundary` component.

KH-Amir-TN:
## Description

https://reactjs.org/docs/error-boundaries.html
Usage:
```
{// Top level, boundary will will display TopErrorPage in place to avoid bubbling the error and having the whole tree unmouted}
<ErrorBoundary fallback={TopErrorPage}>
  <ErrorBoundary>
      {// By default, with no fallback UI any encountered error will unmount the subtree}
     <NavBar />
   </ErrorBoundary>
  <Container>
    <ErrorBoundary fallback={MainErrorPage}>
      {// When encountering an error the error boundary will display MainErrorPage in place}
        <Outlet />
    </ErrorBoundary>
  </Container>
    <ErrorBoundary>
      <Footer/>
  </ ErrorBoundary>
<ErrorBoundary />
````
A deep level usage:
````
export default function AdminPanel(){
return (
   <Container>
  <SideBar />
   <Content>
      {// E.g we want to if possible catch an error bubbling inisde the admin panel tab content area before it ruins the entire UI}
     <ErrorBoundary fallback={PanelErrorPage}>
         <Table />
         {...}
     </ErrorBoundary>
   </Content>
</Container>
);
}
````
Props:
- fallback: A react component that will be rendered in place of the subtree in case of an encountered error, the component will have two props `error` and `errorInfo`, the first is the error object and the latter is a react errorInfo object that includes information about the error from react's perspective e.g the component stack trace.
## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/javascript/components/shared_components/ErrorBoundary.jsx
+++ b/app/javascript/components/shared_components/ErrorBoundary.jsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default class ErrorBoundary extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { error: null, errorInfo: null };
+  }
+
+  componentDidCatch(error, errorInfo) {
+    // Catch errors in any components below in the tree and re-render with error message
+    this.setState({
+      error,
+      errorInfo,
+    });
+  }
+
+  render() {
+    const { error, errorInfo } = this.state;
+    const { fallback: Fallback, children } = this.props;
+
+    if (errorInfo) {
+      // Rendering the fallback UI while providing the error object and info.
+      return <Fallback error={error} errorInfo={errorInfo} />;
+    }
+    // In normal cases, render the subtree.
+    return children;
+  }
+}
+
+ErrorBoundary.defaultProps = {
+  fallback: () => null,
+};
+
+ErrorBoundary.propTypes = {
+  children: PropTypes.node.isRequired,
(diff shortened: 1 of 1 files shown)


scouillard:
Where are we planning to use this?

scouillard (approved):

hadicheaito1 (approved):
I see the use of this. But where do we plan on using this as Samuel asked? Everywhere? And what does the usage look like when using this component in another file?

KH-Amir-TN:
> Where are we planning to use this? We can see a live example [here](https://codepen.io/gaearon/pen/wqvxGa?editors=0010)

Wherever we require catching a raising exception during rendering, stop its bubbling and make side effects preserving the rest of the UI intact (mounted),
Example, we may like to display some UI inside our adminpanel content area if something goes wrong there instead of affecting the entire UI by counting on a higher error boundary in the tree,
I'm quoting this from the official docs: _"For example, Facebook Messenger wraps content of the sidebar, the info panel, the conversation log, and the message input into separate error boundaries. If some component in one of these UI areas crashes, the rest of them remain interactive."_
So, for our case I think we'd start by one top level error boundary and maybe another one that try catching errors in the main container subtree trying to keep hold of the layout in case of an issue.
And as we go will see the need to add other boundaries,

farhatahmad:
Why are we using a class component? Shouldn't we use a function component for consistency?

PR #3942, opened 2022-09-27
Fixed admin queries

farhatahmad:
## Description

There was some inconsistency on how components were fetching their information. Now, each component will only pull the siteSettings that it needs, rather than loading ALL site settings

### DIFF
--- a/app/javascript/components/admin/site_settings/SiteSettings.jsx
+++ b/app/javascript/components/admin/site_settings/SiteSettings.jsx
@@ -8,12 +8,8 @@ import Appearance from './appearance/Appearance';
 import Administration from './administration/Administration';
 import Settings from './settings/Settings';
 import Registration from './registration/Registration';
-import useSiteSettings from '../../../hooks/queries/admin/site_settings/useSiteSettings';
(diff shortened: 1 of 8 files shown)


scouillard (approved):

app/javascript/components/admin/site_settings/SiteSettings.jsx
```diff
@@ -31,29 +27,20 @@ export default function SiteSettings() {
                   <div className="ps-4 pe-4 pt-4">
                     <h3> Customize Greenlight </h3>
                   </div>
-                  {
-                    isLoading
```
hadicheaito1, line 35: Why is it safe to remove the isLoading now?

KH-Amir-TN (approved):
# 🚀 🚀 🚀

app/javascript/components/admin/site_settings/SiteSettings.jsx
farhatahmad: I moved the whole query into `Administration` - the check for is loading is there

PR #3943, opened 2022-09-27
Online Feedback - create online column in Room

scouillard:
## Description

Add online boolean column to Room.
When a meeting is started, `room.online = true`.
When a meeting is closed, `room.online = false` via the callback endpoint in ExternalController.

You can pass a Room or a bunch of Rooms to RunningMeetingChecker and it will confirm if a meeting is truly online by running `get_meeting_info` on the provided rooms.

## TODO

- Add RunningMeetingChecker specs
- Run benchmark on ServerRooms? `active_meetings` vs running the RunningMeeting service?

image
image

### DIFF
--- a/app/controllers/api/v1/admin/server_rooms_controller.rb
+++ b/app/controllers/api/v1/admin/server_rooms_controller.rb
@@ -14,9 +14,19 @@ module Api
 
           pagy, rooms = pagy(rooms)
 
-          rooms = RunningMeetingChecker.new(rooms:).call
+          active_rooms = BigBlueButtonApi.new.active_meetings
+          active_rooms_hash = {}
 
-          render_data data: rooms, meta: pagy_metadata(pagy), serializer: ServerRoomSerializer, status: :ok
+          active_rooms.each do |active_room|
+            active_rooms_hash[active_room[:meetingID]] = active_room[:participantCount]
+          end
+
+          @rooms.each do |room|
+            room.active = active_rooms_hash.key?(room.meeting_id)
+            room.participants = active_rooms_hash[room.meeting_id]
+          end
+
+          render_data data: @rooms, meta: pagy_metadata(pagy), serializer: ServerRoomSerializer, status: :ok
         end
       end
     end
--- a/app/controllers/api/v1/admin/server_rooms_controller.rb
+++ b/app/controllers/api/v1/admin/server_rooms_controller.rb
@@ -21,12 +21,12 @@ module Api
             active_rooms_hash[active_room[:meetingID]] = active_room[:participantCount]
           end
 
-          @rooms.each do |room|
+          rooms.each do |room|
             room.active = active_rooms_hash.key?(room.meeting_id)
             room.participants = active_rooms_hash[room.meeting_id]
           end
 
-          render_data data: @rooms, meta: pagy_metadata(pagy), serializer: ServerRoomSerializer, status: :ok
+          render_data data: rooms, meta: pagy_metadata(pagy), serializer: ServerRoomSerializer, status: :ok
         end
       end
     end
--- a/app/controllers/api/v1/meetings_controller.rb
+++ b/app/controllers/api/v1/meetings_controller.rb
@@ -23,7 +23,6 @@ module Api
             current_user:,
             provider: current_provider
           ).call
-          @room.update(online: true)
         rescue BigBlueButton::BigBlueButtonException => e
           return render_error status: :bad_request unless e.key == 'idNotUnique'
         end
@@ -67,7 +66,6 @@ module Api
               current_user:,
               provider: current_provider
             ).call
-            @room.update(online: true)
           rescue BigBlueButton::BigBlueButtonException => e
             return render_error status: :bad_request unless e.key == 'idNotUnique'
           end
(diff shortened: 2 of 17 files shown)


app/services/running_meeting_checker.rb
```diff
@@ -1,22 +1,37 @@
...
+      # If provided a single room, the service will use bbb get_meeting_info request
+      # get_meeting_info returns an error if the meeting is NOT active
+      begin
+        active_room = BigBlueButtonApi.new.get_meeting_info(meeting_id: @rooms.meeting_id)
+        @rooms.active = true
+        @rooms.participants = active_room[:participantCount]
+      rescue BigBlueButton::BigBlueButtonException
+        @rooms.active = false
+        @rooms.participants = nil
+      end
```
farhatahmad, line 33: Does `get_meeting_info` throw an exception if the meeting isn't running?

scouillard: It returns message: `A meeting with that ID does not exist, messageKey: notFound`
It doesn't seem to be an error as its return code 200.

scouillard: image

app/services/running_meeting_checker.rb
```diff
@@ -1,22 +1,37 @@
...
   def initialize(rooms:)
     @rooms = rooms
   end
 
   def call
-    active_rooms = BigBlueButtonApi.new.active_meetings
-    active_rooms_hash = {}
+    # If provided an array of rooms, the service will use bbb active_meetings request
+    if @rooms.is_a?(Array)
+      active_rooms = BigBlueButtonApi.new.active_meetings
```
farhatahmad, line 12: BigBlueButtonApi.new.get_meeting_info(meeting_id: @rooms.pluck(:meeting_id))

app/controllers/api/v1/admin/server_rooms_controller.rb
```diff
@@ -14,7 +14,17 @@ def index
...
+          active_rooms_hash = {}
+
+          active_rooms.each do |active_room|
+            active_rooms_hash[active_room[:meetingID]] = active_room[:participantCount]
+          end
+
+          rooms.each do |room|
+            room.active = active_rooms_hash.key?(room.meeting_id)
+            room.participants = active_rooms_hash[room.meeting_id]
+          end
```
scouillard, line 27: If `active_meeting `just runs a bunch of `get_meeting_info`, we might as well use our RunningMeeting service here?
I will benchmark tonight or tomorrow.

scouillard:
Lot's of file have changed but most of the stuff is in `services/RunningMeetingChecker` and `RoomsController`.
The other files are mostly the change from the virtual "active" to the column "online" and bunch of styling.

app/controllers/api/v1/admin/server_rooms_controller.rb
```diff
@@ -14,7 +14,7 @@ def index
 
           pagy, rooms = pagy(rooms)
 
-          rooms = RunningMeetingChecker.new(rooms:).call
+          RunningMeetingChecker.new(rooms:).call
```
scouillard, line 17: Double check that `get_meetings` is running a bunch of `get_meeting_info` as well so we might as well use our service in ServerRooms for consistency.

app/controllers/api/v1/admin/server_rooms_controller.rb
scouillard, line 27: I put the old ServerRooms logic so the specs doesn't break.
I think `active_rooms` is just running a bunch of `get_meeting_info`, so we might as well use our service in ServerRooms as well?

db/migrate/20220928193827_add_online_to_room.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddOnlineToRoom < ActiveRecord::Migration[7.0]
+  def change
+    add_column :rooms, :online, :boolean
```
hadicheaito1, line 5: What was the deciding factor for creating a column for `online` vs the approach you had before?

scouillard: It's to avoid doing an expensive `active_meetings` query on every Rooms refresh. We store the meeting status in the database and then we double check only the online meetings status with `get_meeting_info`

app/controllers/api/v1/meetings_controller.rb
```diff
@@ -66,6 +67,7 @@ def status
               current_user:,
               provider: current_provider
             ).call
+            @room.update(online: true)
```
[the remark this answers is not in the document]
scouillard: Should have done that from the start thank you

app/services/meeting_starter.rb
```diff
@@ -21,7 +21,7 @@ def call
     begin
       meeting = BigBlueButtonApi.new.start_meeting room: @room, options: options, presentation_url: @presentation_url
 
-      @room.update!(last_session: DateTime.strptime(meeting[:createTime].to_s, '%Q'))
+      @room.update!(last_session: DateTime.strptime(meeting[:createTime].to_s, '%Q', online: true))
```
farhatahmad, line 24: This isn't in the right place, it should be after the `)`

scouillard: Embarassing

farhatahmad (approved):

PR #3949, opened 2022-09-29
Adding roomLimit permission

hadicheaito1:
## Description

- Adding RoomLimit permission
- Setting default to 100 for admin, user, and guest


db/data/20220721145016_populate_permissions.rb
```diff
@@ -10,7 +10,8 @@ def up
       { name: 'ManageSiteSettings' },
       { name: 'ManageRoles' },
       { name: 'SharedList' },
-      { name: 'CanRecord' }
+      { name: 'CanRecord' },
+      { name: 'roomLimit' }
```
farhatahmad, line 14: Since we've pushed Alpha, we need to create new migrations - not push to old ones

scouillard:
Shouldn't this be a default? 255 seems kind of high already?

db/data/20220929151206_populate_room_limit_permission.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class PopulateRoomLimitPermission < ActiveRecord::Migration[7.0]
+  def up
+    Permission.create! [
+      { name: 'roomLimit' }
```
scouillard, line 6: Could we name it RoomLimit with a capital R for consistency

hadicheaito1: Done

db/data/20220929151319_populate_room_limit_role_permission.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class PopulateRoomLimitRolePermission < ActiveRecord::Migration[7.0]
+  def up
+    admin = Role.find_by(name: 'Administrator')
+    user = Role.find_by(name: 'User')
+    guest = Role.find_by(name: 'Guest')
+
```
farhatahmad, line 8: You can do this all in 1 migration, no need for 2

db/data/20220721145016_populate_permissions.rb
farhatahmad: Remove this :)

db/data/20220929151319_populate_room_limit_role_permission.rb
```diff
@@ -0,0 +1,22 @@
...
+    room_limit = Permission.find_by(name: 'RoomLimit')
+
+    RolePermission.create! [
+
+      { role: admin, permission: room_limit, value: 255 },
+      { role: user, permission: room_limit, value: 255 },
+      { role: guest, permission: room_limit, value: 255 }
```
farhatahmad, line 15: Maybe we set the default to something nicer? 100?

db/data/20220930022007_populate_room_limit_role_permission.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    admin = Role.find_by(name: 'Administrator')
+    user = Role.find_by(name: 'User')
+    guest = Role.find_by(name: 'Guest')
+
+    room_limit = Permission.find_by(name: 'RoomLimit')
+
+    RolePermission.create! [
+
+      { role: admin, permission: room_limit, value: 100 },
```
hadicheaito1, line 17: I think it's better to make the values strings to stay consistent, what do you think?

db/data/20220930022007_populate_room_limit_role_permission.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class PopulateRoomLimitRolePermission < ActiveRecord::Migration[7.0]
+  def up
+    Permission.create! [
```
KH-Amir-TN, line 5: I think you can add this directly to the `PopulatePermissions`.

db/data/20220930022007_populate_room_limit_role_permission.rb
KH-Amir-TN: They will be casted into a string when saving the record but I believe it's better to be consistent with the rest of role permissions creation.

KH-Amir-TN (approved):
I think at this step you can update and use both of `PopulatePermissions` and `PopulateRolePermissions` since that's what we gonna do at a later step when merging migrations.

scouillard (dismissed):

db/data/20220930022007_populate_room_limit_role_permission.rb
hadicheaito1: I was told by @farhatahmad to create new migration file for these since we released alpha and don't want people to have to nuke their db for the new changes

db/data/20220930022007_populate_room_limit_role_permission.rb
hadicheaito1: WIll do in next PR

db/data/20220930022007_populate_room_limit_role_permission.rb
KH-Amir-TN: Yeah, good point!

farhatahmad (approved):

PR #3950, opened 2022-09-29
Added ResyncOnLogin Site Setting

farhatahmad:
## Description

Added a new site setting that will only appear if using an external authentication provider. This site setting will determine what is the source of truth - Greenlight or the Authentication provider.

If Greenlight is the source of truth, the user's information (name, email, picture, .. ) will never change even if they change in the authentication provider. 

If the authentication provider is the source of truth, the information in greenlight (name, email, picture) will always match whatever is in the authentication provider

### DIFF
--- a/app/controllers/external_controller.rb
+++ b/app/controllers/external_controller.rb
@@ -5,15 +5,29 @@ class ExternalController < ApplicationController
 
   # GET 'auth/:provider/callback'
   def create_user
+    resync = SettingGetter.new(setting_name: 'ResyncOnLogin', provider: current_provider).call
+
     credentials = request.env['omniauth.auth']
     user_info = credentials['info']
-    user = User.find_or_create_by(email: user_info['email']) do |u|
-      u.external_id = credentials['uid']
-      u.name = user_info['name']
-      u.provider = current_provider
-      u.role = Role.find_by(name: 'User') # TODO: - Ahmad: Move to service
-      u.language = extract_language_code user_info['locale']
-    end
+
+    user = if resync
+             User.find_or_initialize_by(external_id: credentials['uid']).tap do |u|
+               u.email = user_info['email']
+               u.name = user_info['name']
+               u.provider = current_provider
+               u.role = Role.find_by(name: 'User') # TODO: - Ahmad: Move to service
+               u.language = extract_language_code user_info['locale']
+               u.save!
+             end
+           else
+             User.find_or_create_by(external_id: credentials['uid']) do |u|
+               u.email = user_info['email']
+               u.name = user_info['name']
+               u.provider = current_provider
+               u.role = Role.find_by(name: 'User') # TODO: - Ahmad: Move to service
+               u.language = extract_language_code user_info['locale']
+             end
+           end
 
     user.generate_session_token!
     session[:session_token] = user.session_token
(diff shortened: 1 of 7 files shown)


app/controllers/external_controller.rb
```diff
@@ -3,15 +3,29 @@
...
+             end
+           else
+             User.find_or_create_by(external_id: credentials['uid']) do |u|
+               u.email = user_info['email']
+               u.name = user_info['name']
+               u.provider = current_provider
+               u.role = Role.find_by(name: 'User') # TODO: - Ahmad: Move to service
+               u.language = extract_language_code user_info['locale']
+             end
+           end
```
farhatahmad, line 28: I'm sure there's a better way to do this - I just haven't found it yet

KH-Amir-TN: I've invested some time on this, don't know if that helps?
```
def create_user
    provider = current_provider # A small optimization to avoid triggering the current_provider logic twice.

    user_info = {
      name: credentials['info']['name'],
      email: credentials['info']['email'],
      provider:,
      role: Role.find_by(name: 'User'),
      language: extract_language_code(credentials['info']['locale'])
    }

    # Finding or creating the user external account.
    user = User.find_or_create_by!(external_id: credentials['uid']) { |u| u.assign_attributes(user_info) }
    # Getting the current_provider resync site setting.
    resync = SettingGetter.new(setting_name: 'ResyncOnLogin', provider:).call
    # Updating the user external account if it changed? with the external openid_connect provider data.
    # Only when resync site setting is enabled for the provider.
    if resync
      user.assign_attributes(user_info)
      user.save! if user.changed?
    end

    user.generate_session_token!
    session[:session_token] = user.session_token

    redirect_to '/rooms'
  end
 ```

Comment (KH-Amir-TN) [reply app/controllers/external_controller.rb:28]: For the external accounts, we'd need to infer their role based on their email address especially since it can change when resync is enabled?
I think that makes sense only for creation and for updates (when resync is on) it may introduce an attack vector (privilege escalation) for when the external provider is compromised?

**Review by KH-Amir-TN — APPROVED** [review]

File: app/controllers/external_controller.rb (hunk 1 above)
Comment (scouillard) [reply app/controllers/external_controller.rb:28]: If resync is true, it will use `.new` + `.save` instead of `.create` to catch validations? What is the use of this?

Comment (farhatahmad) [reply app/controllers/external_controller.rb]: Seems to work - I made some slight changes to it

File: app/controllers/external_controller.rb [hunk 2]
```diff
@@ -5,14 +5,23 @@ class ExternalController < ApplicationController
...
-      u.role = Role.find_by(name: 'User') # TODO: - Ahmad: Move to service
-      u.language = extract_language_code user_info['locale']
+    user_info = {
+      name: credentials['info']['name'],
+      email: credentials['info']['email'],
+      language: extract_language_code(credentials['info']['locale'])
+    }
+
+    user = User.find_or_create_by!(external_id: credentials['uid'], provider:) do |u|
+      user_info[:role] = Role.find_by(name: 'User')
```
farhatahmad, line 18: Only set the role if creating the user regardless of resync (will be changed to in @scouillard's PR)

KH-Amir-TN (approved):

PR #3955, opened 2022-09-30
Add Admin/ManageUsers sorting

scouillard:
## Description

Add "name" and "role" sorting on Admin/Manage Users table.
Set default sorting with "role_id" - the role with the lowest role_id, by default "administrator", will appear on top.

### DIFF
--- a/app/controllers/api/v1/admin/users_controller.rb
+++ b/app/controllers/api/v1/admin/users_controller.rb
@@ -9,8 +9,11 @@ module Api
         end
 
         def active_users
+          sort_config = config_sorting(allowed_columns: %w[name email role.name])
+
           # TODO: Change to get active users only
-          users = User.includes(:role).with_provider(current_provider).with_attached_avatar.search(params[:search])
+
+          users = User.includes(:role).with_provider(current_provider).with_attached_avatar.order(sort_config, role_id: :desc)&.search(params[:search])
 
           pagy, users = pagy(users)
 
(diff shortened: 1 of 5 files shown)


app/controllers/api/v1/admin/users_controller.rb
```diff
@@ -9,8 +9,13 @@ class UsersController < ApiController
...
 
         def active_users
+          sort_config = config_sorting(allowed_columns: %w[name roles.name])
+
           # TODO: Change to get active users only
-          users = User.includes(:role).with_provider(current_provider).with_attached_avatar.search(params[:search])
+          users = User.includes(:role)
+                      .with_provider(current_provider)
+                      .with_attached_avatar
+                      .order(sort_config, role_id: :desc)&.search(params[:search])
```
hadicheaito1, line 18: By default, should we make it alphabetical by user's name? Just an idea

hadicheaito1 (dismissed):

app/controllers/api/v1/admin/users_controller.rb
scouillard: Could be. I think it makes sense to put admins at the top of the list. But it might be less appropriate if a provider has a bunch of custom roles.

scouillard: I think I would let it as is unless I am told that most providers don't use the default role hierarchy of admin-users-guests

scouillard: Yep I like this

farhatahmad (approved):

PR #3956, opened 2022-10-03
RoomLimit - Front end + validations

hadicheaito1:
## Description

- Front end for the RoomLimit role permission input
- Front end validations done

image

## TODO in next PRs
- Back end check for if the new limit is in the range before updating
- Back end check for room limit before creating new rooms


hadicheaito1:
Note:
Currently the input box is being filled with a placeholder that has a lighter font colour than when the actual input value is set such as the role name input box on this page. I was not able to set the value of the `RoleLimit` since the `rolePermissions` would not be loaded in time for me to set the value in the `useEffect`. I wasn't sure what to do in this case and would like your suggestions since I think the two boxes should be consistent obviously.

app/javascript/helpers/forms/EditRoleFormHelpers.jsx
```diff
@@ -1,18 +1,28 @@
...
 });
 
-export const editRoleFormConfig = {
+const validationSchemaRoomLimit = yup.object({
+  roomLimit: yup.number().required('Please enter the room count limit.').min(0, 'Minimum atleast 0').max(100, 'Allowed maximum is 100'),
+});
+
+export const editRoleFormConfigRoleName = {
+  mode: 'onBlur',
+  defaultValues: {},
```
KH-Amir-TN, line 15: It's recommended to provide in `defaultValues` even if we have a `useEffect` hook that will set values,
The form will render before the hook will be called,

app/javascript/components/admin/roles/forms/EditRoleForm.jsx
```diff
@@ -16,30 +18,36 @@ import RolePermissionRow from '../RolePermissionRow';
...
 
   if (roomConfigs.isLoading || rolePermissionsIsLoading) return <Spinner />;
+  fieldsRoomLimit.roomLimit.placeHolder = rolePermissions.RoomLimit;
 
   return (
     <div>
       <Stack>
-        <Form methods={methods} onBlur={(e) => updateRoleAPI.mutate({ name: e.target.value })}>
-          <FormControl field={fields.name} type="text" />
+        <Form methods={methodsRoleName} onBlur={(e) => updateRoleAPI.mutate({ name: e.target.value })}>
```
KH-Amir-TN, line 45: I suspect that `react-hook-form` here isn't handling neither the submission nor the form data.
We're bypassing it,
A great example of the implication of that is that client side validations have no effect besides providing a feedback (it will not prevent the API call to be made).
Another thing that may cause an issue is the lack of preventing of the browser defaults for the `onblur`  (haven't checked the default behavior and its implication) on the form.

app/javascript/helpers/forms/EditRoleFormHelpers.jsx
KH-Amir-TN: Here that is causing the problem where you blur the input without typing any value you'll get this weird error message:
`roomLimit must be a `number` type, but the final value was: `NaN` (cast from the value `""`).`
image

With the issues noted above that undefined value will be passed to the API:
image

app/javascript/components/admin/roles/forms/EditRoleForm.jsx
KH-Amir-TN: After ensuring the good use of react hook form I recommend that you (if possible) separate both of these forms into their separate components and make this a layout component that renders both of these forms instead of this `EditRoleForm` that has two forms,

KH-Amir-TN (changes requested):

app/javascript/components/admin/roles/forms/EditRoleForm.jsx
hadicheaito1: Yes you are right, I was having issues with this. Honestly I find the best solution to just create an update button beside each of the inputs. I feel this would also align with the rest of the application

farhatahmad:
This looks kind of weird - I'd rather we move it below the other permissions and not make full width

hadicheaito1:
> This looks kind of weird - I'd rather we move it below the other permissions and not make full width

I made it like this:
image

After playing around with it, I found that when it error it messes up (something to do with the validations), so we might have to keep the input bigger to prevent this:
(Notice how it moves when validation error)
image


Also lmk if we should keep the Room Limit label on top of the input bar and not push it to the end

farhatahmad (approved):

PR #3961, opened 2022-10-03
Reintroduce csrf

farhatahmad:
## Description

`reset_session` invalidates the CSRF. This isn't the perfect solution but is better than what we currently have

### DIFF
--- a/app/javascript/helpers/Axios.jsx
+++ b/app/javascript/helpers/Axios.jsx
@@ -11,6 +11,7 @@ const axiosInstance = axios.create(
     headers: {
       'Content-Type': 'application/json',
       Accept: 'application/json',
+      'X-CSRF-TOKEN':  document.querySelector('[name=csrf-token]').getAttribute("content")
     },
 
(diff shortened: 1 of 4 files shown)


app/javascript/helpers/Axios.jsx
```diff
@@ -11,6 +11,7 @@ const axiosInstance = axios.create(
     headers: {
       'Content-Type': 'application/json',
       Accept: 'application/json',
+      'X-CSRF-TOKEN': document.querySelector('[name=csrf-token]').getAttribute('content'),
```
hadicheaito1, line 14: What's this doing exactly?

hadicheaito1 (approved):

app/javascript/helpers/Axios.jsx
farhatahmad: It gets the csrf token that Rails sets in th header and passes it to the backend so that Rails can check it on every non GET requests

KH-Amir-TN (approved):

PR #3962, opened 2022-10-03
Add DefaultRole SiteSetting

scouillard:
## Description

Add DefaultRole site setting.
Admin can now set a default role upon user creation.
Create "default_role" method in Application controller as "default_role" is needed in both local and external account creation.
Setting is accessible in Admin/SiteSettings/Registration.

## TODO
Upcoming PR: Make default roles (User, Admin, Guest) undeletable.

## Testing Steps

Create a new role.
Select the new role in SiteSettings/Registration.
Create a new user and verify it has the selected default role.

image

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -97,7 +97,7 @@ module Api
       private
 
       def user_params
-        params.require(:user).permit(:name, :email, :password, :avatar, :language, :role_id)
+        params.require(:user).permit(:name, :email, :password, :password_confirmation, :avatar, :language, :role_id)
       end
 
       def change_password_params
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -97,7 +97,7 @@ module Api
       private
 
       def user_params
-        params.require(:user).permit(:name, :email, :password, :password_confirmation, :avatar, :language, :role_id)
+        params.require(:user).permit(:name, :email, :password, :avatar, :language, :role_id)
       end
 
       def change_password_params
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -26,7 +26,7 @@ module Api
         # TODO: amir - ensure accessibility for unauthenticated requests only.
         params[:user][:language] = I18n.default_locale if params[:user][:language].blank?
 
-        user = UserCreator.new(user_params:, provider: current_provider).call
+        user = UserCreator.new(user_params:, provider: current_provider, default_role:).call
 
         # TODO: Add proper error logging for non-verified token hcaptcha
         return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -26,7 +26,7 @@ module Api
         # TODO: amir - ensure accessibility for unauthenticated requests only.
         params[:user][:language] = I18n.default_locale if params[:user][:language].blank?
 
-        user = UserCreator.new(user_params:, provider: current_provider, default_role:).call
+        user = UserCreator.new(user_params:, provider: current_provider, role: default_role).call
 
         # TODO: Add proper error logging for non-verified token hcaptcha
         return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -36,7 +36,7 @@ class ApplicationController < ActionController::Base
   # Returns the default role
   def default_role
     default_role_setting = SettingGetter.new(setting_name: 'DefaultRole', provider: current_provider).call
-    @default_role = Role.find_by(name: default_role_setting, provider: current_provider)
+    @default_role = Role.find_by(name: default_role_setting, provider: current_provider) || Role.find_by(name: 'User', provider: current_provider)
   end
 
   private
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -19,6 +19,7 @@ class ApplicationController < ActionController::Base
     @current_user ||= user
   end
 
+
   # Returns whether hcaptcha is enabled by checking if ENV variables are set
   def hcaptcha_enabled?
     (ENV['HCAPTCHA_SITE_KEY'].present? && ENV['HCAPTCHA_SECRET_KEY'].present?)
@@ -32,6 +33,13 @@ class ApplicationController < ActionController::Base
                             'greenlight'
                           end
   end
+
+  # Returns the default role
+  def default_role
+    default_role_setting = SettingGetter.new(setting_name: 'DefaultRole', provider: current_provider).call
+    @default_role = Role.find_by(name: default_role_setting)
+  end
+
 
   private
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -36,7 +36,7 @@ class ApplicationController < ActionController::Base
   # Returns the default role
   def default_role
     default_role_setting = SettingGetter.new(setting_name: 'DefaultRole', provider: current_provider).call
-    @default_role = Role.find_by(name: default_role_setting)
+    @default_role = Role.find_by(name: default_role_setting, provider: current_provider)
   end
 
   private
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -19,7 +19,6 @@ class ApplicationController < ActionController::Base
     @current_user ||= user
   end
 
-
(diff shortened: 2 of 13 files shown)


scouillard:
Will re-check specs after lunch

farhatahmad:
You need to factor in external accounts as well

app/controllers/api/v1/users_controller.rb
```diff
@@ -97,7 +97,7 @@ def change_password
       private
 
       def user_params
-        params.require(:user).permit(:name, :email, :password, :avatar, :language, :role_id)
+        params.require(:user).permit(:name, :email, :password, :password_confirmation, :avatar, :language, :role_id)
```
scouillard, line 100: Kind of unrelated to the PR. Couldn't create a User from admin panel without it.

farhatahmad: @KH-Amir-TN ?

db/data/20221003201116_populate_default_role_settings.rb
```diff
@@ -0,0 +1,18 @@
...
+      { name: 'DefaultRole' }
+    ]
+
+    SiteSetting.create! [
+      { setting: Setting.find_by(name: 'DefaultRole'), provider: 'greenlight', value: 'User' }
+    ]
+  end
+
+  def down
+    Setting.find_by(name: 'DefaultRole').delete
```
KH-Amir-TN, line 15: I suspect that this will fail,
A general rule of thumb, destroy the dependent resources before destroying the resource that uses them.

app/controllers/application_controller.rb
```diff
@@ -33,6 +33,12 @@ def current_provider
                           end
   end
 
+  # Returns the default role
+  def default_role
+    default_role_setting = SettingGetter.new(setting_name: 'DefaultRole', provider: current_provider).call
+    @default_role = Role.find_by(name: default_role_setting)
```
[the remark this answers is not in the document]
KH-Amir-TN: And if we can't find the "default_role" we may have to default to the provider "User" role which is stable and not meant to be removed, right?

spec/services/user_creator_spec.rb
```diff
@@ -4,22 +4,35 @@
 
 describe UserCreator, type: :service do
   describe '#call' do
-    let!(:users) { create(:role, name: 'User') }
+    let(:users) { create(:role, name: 'User') }
+    let(:teachers) { create(:role, name: 'Teachers') }
+    let(:fake_setting_getter) { instance_double(SettingGetter) }
```
KH-Amir-TN, line 9: What's the use of `fake_setting_getter` here?

spec/services/user_creator_spec.rb
```diff
@@ -4,22 +4,35 @@
...
 
     before do
       setting = create(:setting, name: 'RoleMapping')
       create(:site_setting, setting:, provider: 'greenlight', value: 'Decepticons=@decepticons.cybertron,Autobots=autobots.cybertron')
     end
 
-    it 'creates a user with the role matching a rule if role found' do
+    it 'creates a user with the default role' do
```
KH-Amir-TN, line 16: Isn't this test duplicated in favor of the last two examples?

KH-Amir-TN (commented):
Things that we may need to check:
1. The role name uniqueness per provider scope.
2. A user record cannot have a role from a different provider.
3. the default_role if not found we fallback to the current_provider "User" role.
4. "User" role cannot be deleted even when no record depends on it (I believe this is a WIP) 
5. Each provider should have its essential roles created? "User, Guest & administrator"

app/controllers/application_controller.rb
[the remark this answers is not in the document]
scouillard: > Besides, what is the use of the instance variabe @default_role?

For consistency as currrent_provider and current_user are returning an instance variable. I had the same question TBH. Does not need to return an instance variable.

app/services/user_creator.rb
```diff
@@ -1,18 +1,19 @@
 # frozen_string_literal: true
 
 class UserCreator
-  def initialize(user_params:, provider:)
+  def initialize(user_params:, provider:, role:)
```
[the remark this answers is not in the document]
scouillard: IMO, Role feels right as we can in fact pass any role to it.

spec/services/user_creator_spec.rb
scouillard: What last two examples? There is definitely some redundancy in the user_creator_specs. This test isn't really adding anything I agree.

db/data/20221003201116_populate_default_role_settings.rb
scouillard: What do you propose? Delete all Settings and all Roles?
Most of our recent migrations have `ActiveRecord::IrreversibleMigration`

app/controllers/application_controller.rb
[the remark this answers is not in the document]
scouillard: > 



> Things that we may need to check:
> 
> 1. The role name uniqueness per provider scope.
> 2. A user record cannot have a role from a different provider.
> 3. the default_role if not found we fallback to the current_provider "User" role.
> 4. "User" role cannot be deleted even when no record depends on it (I believe this is a WIP)
> 5. Each provider should have its essential roles created? "User, Guest & administrator"

1. Check. `populate_roles` migration in db/data
2. Check in current PR.
3. Should always be found since the default value of the DefaultRole SiteSetting is set to 'User' in `populate_default_role_setting` migration in db/data. Could add redundancy to ApplicationController#default_role.
4. Done in a previous PR. "Guest" "Administrator" and "User" can't be deleted.
5. Check. `populate_roles` migration in db/data

db/data/20221003201116_populate_default_role_settings.rb
scouillard: I swapped them

app/controllers/application_controller.rb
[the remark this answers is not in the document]
KH-Amir-TN: I think I may want to synchronize with you on the concerns addressed above and their scope,
I believe that we may not have solved all of the issues,

app/services/user_creator.rb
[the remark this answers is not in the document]
KH-Amir-TN: But in that context it will used as a default/fallback role?
A default role is a role, it just indicates that that role will be used by default.
In that context the role coming from `infer_role_from_email` can be named `role` when the other is a fallback role,
However, it's not an issue and I'm okay with both naming schema as long that it's clear (which is).

db/data/20221003201116_populate_default_role_settings.rb
KH-Amir-TN: I think the issue isn't solved,
You can try rolling back the data migration to confirm if your rollback logic is valid for that you can run `rails data:rollback` simply since it's the last migration,
I believe something like was intended:
```
  def down
    setting = Setting.find_by(name: 'DefaultRole')
    SiteSetting.where(setting:).destroy_all
    setting.destroy!
  end
```

scouillard: Got it. Nicely done. Thank you

farhatahmad (approved):

app/controllers/application_controller.rb
```diff
@@ -33,6 +33,12 @@ def current_provider
                           end
   end
 
+  # Returns the default role
+  def default_role
+    default_role_setting = SettingGetter.new(setting_name: 'DefaultRole', provider: current_provider).call
+    @default_role = Role.find_by(name: default_role_setting, provider: current_provider) || Role.find_by(name: 'User', provider: current_provider)
```
KH-Amir-TN, line 39: Sorry for the late review,
What's the use of the `@default_role` ?
Don't we like to raise an exception when we cannot find the User record?
What do you think guys?

app/controllers/external_controller.rb
```diff
@@ -15,7 +15,7 @@ def create_user
     }
 
     user = User.find_or_create_by!(external_id: credentials['uid'], provider:) do |u|
-      user_info[:role] = Role.find_by(name: 'User')
+      user_info[:role] = default_role
```
KH-Amir-TN, line 18: Good catch!

app/services/user_creator.rb
```diff
@@ -1,18 +1,19 @@
...
 
   def call
-    role = infer_role_from_email(@user_params[:email])
+    email_role = infer_role_from_email(@user_params[:email])
 
     User.new({
       provider: @provider,
-      role:
+      role: email_role || @role
     }.merge(@user_params))
```
KH-Amir-TN, line 17: I think we may want to overwrite the `user_params` with the provider and role instead of the opposite that we're doing,
 There's a room for error from the service perspective where the user_params can overwrite the role and provider,

app/controllers/application_controller.rb
scouillard: I'm returning an instance variable for consistency as all other methods in ApplicationControllers are returning an instance variable.

KH-Amir-TN: Being consistent is good,
However, these variables has no function except they're causing an extra usage of memory that we don't need.

PR #3964, opened 2022-10-04
Small styling changes

farhatahmad:
## Description

Added the following changes
1- Light border around room card when hovering over it (see screenshot)
2- Added Placeholders for Feature Tabs to prevent flickering
3- Fixed Homepage buttons overflowing


farhatahmad:
It's a very light border - just giving some feedback to the user that this is what they're clicking on

hadicheaito1 (commented):
I think it looks good. Just rubocop needs to be fixed (it looks unrelated to your changes?) @farhatahmad

KH-Amir-TN (commented):
I believe placeholders will be our primary way to provide loading feedbacks to users,
However, I got some notes regarding the use of placeholders to solve the flickering of the tab.
We can discuss that during today's meeting,

PR #3970, opened 2022-10-05
Fix Join Meeting layout

scouillard:
## Description

Add Avatar to Join.
Remove include_owner logic in Room#show, Room#show_public.
Fix Join layout.

## Testing Steps

Try to use the Join feature and make sure nothing broke.

image

image

### DIFF
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -271,6 +271,22 @@ input.search-bar {
   }
 }
 
+.btn-link {
+  background-color: transparent;
+  border: 0;
+  color: #0d6efd;
+  padding-left: 0;
+
+  &:hover {
+    color: #0a58ca !important;
+    text-decoration: none !important;
+  }
+
+  &:focus {
+    box-shadow: none;
+  }
+}
+
 //Bootstrap overwrite
 .form-switch .form-check-input:focus {
   border-color: rgba(0, 0, 0, 0.25);
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -275,7 +275,6 @@ input.search-bar {
   background-color: transparent;
   border: 0;
   color: #0d6efd;
-  padding-left: 0;
 
   &:hover {
     color: #0a58ca !important;
--- a/app/javascript/components/rooms/room/RoomJoin.jsx
+++ b/app/javascript/components/rooms/room/RoomJoin.jsx
@@ -117,6 +117,8 @@ export default function RoomJoin() {
     </div>
   );
 
+  env.OPENID_CONNECT = true;
+
   return (
     <div className="vertical-center">
       <div className="text-center pb-4">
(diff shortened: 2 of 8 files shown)


app/javascript/components/rooms/room/RoomJoin.jsx
```diff
@@ -152,6 +155,9 @@ export default function RoomJoin() {
           )}
         </Card.Footer>
       </Card>
+      <div className="text-center text-muted mt-3"> { t('authentication.already_have_account') }
+        <Link to="/signin" className="text-link"> { t('authentication.sign_in') } </Link>
```
farhatahmad, line 159: This doesn't work if external accounts is connected - we should think of a better way to handle that case

app/assets/stylesheets/application.bootstrap.scss
```diff
@@ -271,6 +271,21 @@ input.search-bar {
   }
 }
 
+.btn-link {
+  background-color: transparent;
+  border: 0;
+  color: #0d6efd;
+
+  &:hover {
+    color: #0a58ca !important;
```
farhatahmad, line 280: I'm not sure we should harcode this. If your PrimaryColor is red, this would look very weird

app/assets/stylesheets/application.bootstrap.scss
```diff
@@ -271,6 +271,21 @@ input.search-bar {
   }
 }
 
+.btn-link {
+  background-color: transparent;
+  border: 0;
+  color: #0d6efd;
```
farhatahmad, line 277: I think this should match the PrimaryColor

app/assets/stylesheets/application.bootstrap.scss
scouillard: image

app/assets/stylesheets/application.bootstrap.scss
scouillard: It's to match the Powered by Greenlight link. IMO It has to be blue so it indicates it is a link.

app/assets/stylesheets/application.bootstrap.scss
scouillard: the "Sign In" is a button since the OPENID connect is a form. I created the "link" variant so it looks like an <a> link.

app/assets/stylesheets/application.bootstrap.scss
farhatahmad: Yeah but we're branding everything else, so we might as well brand the links as well

farhatahmad (approved):

app/assets/stylesheets/application.bootstrap.scss
farhatahmad: We can discuss this tomorrow

scouillard: I'd keep the links blue. But yeah I know what you mean. I have been branding the inputs for example but I think we should let it the default blue as well.

scouillard: Yeah we could brand it too it looks all right.

PR #3972, opened 2022-10-06
RoomLimit Backend Check

hadicheaito1:
## Description

- Check when the room is created if the limit has been reached or passed. IF so, then return error and don't create new room
- Check works in both cases where you are creating your own room or if someone is creating a room for another user/role
## Testing Steps

- Change room limit for signed in role
- Try to create more rooms than the limit
- Set limit for user role 
- try to create a room for user role from manage users admin panel
- also log in to user role and try to pass the limit


hadicheaito1:
Waiting on #3956

spec/controllers/rooms_controller_spec.rb
```diff
@@ -204,6 +204,24 @@
...
+      end
+    end
+
+    context 'user has reached the room limit set for their role' do
+      it 'room is not created since room limit has been reached' do
+        user = create(:user, :with_roomLimit_3_permission)
+        sign_in_user(user)
+        create_list(:room, 3, user:)
+        room_params = {
+          user_id: user.id,
```
KH-Amir-TN, line 218: I think this line isn't required,

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -58,6 +58,8 @@ def public_show
 
       # POST /api/v1/rooms.json
       def create
+        return render_error status: :bad_request, errors: 'RoomLimitError' if room_limit_reached(User.find(room_params[:user_id]))
```
KH-Amir-TN, line 61: Can we furhter clean this `room_limit_reached(User.find(room_params[:user_id`?
We can make the method take in a `user_id` or load the user record first and maybe check if it exists before proceeding (Obscure the error in that case),

app/javascript/hooks/mutations/admin/manage_users/useCreateServerRoom.jsx
```diff
@@ -14,7 +14,13 @@ export default function useCreateServerRoom({ userId, onSettled }) {
         queryClient.invalidateQueries('getServerRooms');
         toast.success(t('toast.success.room.room_created'));
       },
-      onError: () => { toast.error(t('toast.error.problem_completing_action')); },
+      onError: (err) => {
+        if (err.response.data.errors === 'RoomLimitError') {
```
KH-Amir-TN, line 18: Well done!

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -114,6 +116,14 @@ def find_room
       def room_params
         params.require(:room).permit(:name, :user_id, :presentation)
       end
+
+      def room_limit_reached(user)
+        if RolePermission.joins(:permission).find_by(role_id: user.role_id, permission: { name: 'RoomLimit' }).value.to_i <= user.rooms.count.to_i
```
KH-Amir-TN, line 121: This is starting to feel that it should be extracted in a `RolePermission` or a `Permission` class method,

KH-Amir-TN (approved):
This is really clean,
However, if we want some extra control we may like to make the check on the model layer,

app/controllers/api/v1/rooms_controller.rb
scouillard: Did you think about adding this logic to services/PermissionsChecker.rb instead?

hadicheaito1: I feel it would complicate things a bit to check for this specific case in the `PermissionsChecker`. `PermissionsChecker` i feel should stick to only returning whether or not the user has the permission.

hadicheaito1: Decided to go with Amir's suggestion of adding the function to the `RolePermission` class

scouillard: I can help you with tomorrow if you want. I agree it's more complicated. I guess it's the price to pay for consistensy. On the top of my mind, I think it would work pretty much like every other permission.

app/models/role_permission.rb
```diff
@@ -5,4 +5,15 @@ class RolePermission < ApplicationRecord
   belongs_to :role
 
   validates :value, presence: true
+
+  def self.room_limit_reached(user_id)
+    return render_error status: :not_found unless User.find(user_id)
```
hadicheaito1, line 10: Only issue i find with this check per Amir's request, is that we are doing the User.find query twice now. Is this an issue? Is there a better way?

hadicheaito1: Also can/will move this function to be part of PermissionChecker service

app/controllers/api/v1/rooms_controller.rb
KH-Amir-TN: I agree with what Samuel suggested,
This is indeed requiring what the `call` logic is providing,
It's a permission,
That may remove the need to do that minor refactoring to provide a method that does the joining and fetches a permission record cleanly (if the same logic isn't used in several areas of the app),

app/models/role_permission.rb
KH-Amir-TN: I think you lost me on this,
I meant to maybe make a class method that reassembles to `MeetingOption.get_setting_value` that removes the need to write that query manually each time we fetch the value of a permission for a given role,

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -12,6 +12,9 @@ class RoomsController < ApiController
       before_action only: %i[create] do
         ensure_authorized('ManageUsers', user_id: room_params[:user_id])
       end
+      before_action only: %i[create] do
```
KH-Amir-TN, line 15: This is fine for now,
But if we're gonna have a lot of this `AND` cases for the same actions as a pattern we may like to make a small refactoring to clean this up,

app/services/permissions_checker.rb
```diff
@@ -56,4 +58,14 @@ def authorize_manage_recordings
 
     @current_user.recordings.find_by(record_id: @record_id).present?
   end
+
+  def authorize_room_limit
+    user = User.find(@user_id)
```
KH-Amir-TN, line 63: We may like to be consistent with the rest of the `authorize_*` methods by returning `false` directly if the `@user_id` is `blank?`.

app/services/permissions_checker.rb
```diff
@@ -56,4 +58,14 @@ def authorize_manage_recordings
...
+    # return true if room limit has not been reached
+    if RolePermission.joins(:permission).find_by(role_id: user.role_id, permission: { name: 'RoomLimit' }).value.to_i <= user.rooms.count.to_i
+      return false
```
KH-Amir-TN, line 66: This is not that crucial at this point,
But we can make a small refactoring to enable extracting the permissions hash and access the permission value here without making any extra queries to the database,
So, one query fetches the hash of `Hash(:permission_name => :value)` once and then the rest of the logic access it at the application level,

farhatahmad:
To solve your issue with the permission checker and the error stuff, I would move away from the `ensure_authorized` pattern and just add a check to the top of the `create` method .

Basically
```
def create
  return render_error, error: "RoomLimitError" if PermissionChecker.....
end
```

spec/factories/user_factory.rb
```diff
@@ -57,6 +57,18 @@
...
+      after(:create) do |user|
+        create(:role_permission, role: user.role, permission: create(:permission, name: 'RoomLimit'), value: '100')
+      end
+    end
+
+    trait :with_roomLimit_3_permission do
+      after(:create) do |user|
+        create(:role_permission, role: user.role, permission: create(:permission, name: 'RoomLimit'), value: '3')
+      end
+    end
```
farhatahmad, line 70: I wouldn't put this in a trait, I would just create the role permission directly in the rooms_controller spec

spec/factories/user_factory.rb
```diff
@@ -57,6 +57,18 @@
       end
     end
 
+    trait :with_roomLimit_100_permission do
+      after(:create) do |user|
+        create(:role_permission, role: user.role, permission: create(:permission, name: 'RoomLimit'), value: '100')
+      end
+    end
```
farhatahmad, line 64: I wouldn't make this a trait either since we would expect a user to be able to create a number of rooms. Do it in the `after(:create)` on line 14, similar to `CreateRoom`

app/services/permissions_checker.rb
```diff
@@ -56,4 +58,16 @@ def authorize_manage_recordings
 
     @current_user.recordings.find_by(record_id: @record_id).present?
   end
+
+  def authorize_room_limit
+    return false if @user_id.blank?
+
+    user = User.find(@user_id)
+    # return true if room limit has not been reached
+    if RolePermission.joins(:permission).find_by(role_id: user.role_id, permission: { name: 'RoomLimit' }).value.to_i <= user.rooms.count.to_i
```
KH-Amir-TN, line 67: This is not expected to happen but this may crash for unfound users,

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -59,6 +59,11 @@ def public_show
 
       # POST /api/v1/rooms.json
       def create
+        return render_error status: :bad_request, errors: 'RoomLimitError' unless PermissionsChecker.new(
```
KH-Amir-TN, line 62: IK we've asked for this,
But, IMO the permission checker is used more like a private method where it can change its signature in the future supporting new permissions,
It's expected to be used through the `ensure_authroized`,
This introduces both inconsistency, requires more boiler plate code and expands the size of the action,
Future contributors can misuse the permission checker easily,

PR #3974, opened 2022-10-06
Add undeletable roles

scouillard:
## Description

Set "User", "Administrator" and "Guest" as undeletable.
Remove "Delete Role" button if it's one of those 3 roles.
Add custom toast if Admin tries to delete a role that is assigned to users.


app/controllers/api/v1/admin/roles_controller.rb
```diff
@@ -63,6 +63,9 @@ def update
         # Returns: { data: Array[serializable objects] , errors: Array[String] }
         # Does: Deletes a role.
         def destroy
+          default_roles = @role.name == 'User' || @role.name == 'Administrator' || @role.name == 'Guest'
+          return render_error errors: @role.errors.to_a, status: :bad_request if default_roles
```
farhatahmad, line 67: I would return a `:forbidden` here instead since `@role.errors.to_a` wil be blank

app/controllers/api/v1/admin/roles_controller.rb
```diff
@@ -63,6 +63,9 @@ def update
         # Returns: { data: Array[serializable objects] , errors: Array[String] }
         # Does: Deletes a role.
         def destroy
+          default_roles = @role.name == 'User' || @role.name == 'Administrator' || @role.name == 'Guest'
```
farhatahmad, line 66: We can clean this up slightly - `['User','Administrator','Guest'].includes?(@role.name)`

app/javascript/hooks/mutations/admin/roles/useDeleteRole.jsx
```diff
@@ -12,7 +12,13 @@ export default function useDeleteRole({ role, onSettled }) {
   return useMutation(
     () => axios.delete(`/admin/roles/${role.id}.json`),
     {
-      onError: () => { toast.error(t('toast.error.problem_completing_action')); },
+      onError: (error) => {
+        if (error.response?.status === 500) {
```
farhatahmad, line 16: That way you can explicity check for a `403` here

scouillard: Should not have to check for a 403 here since we are removing the "Delete Role" button on frontend for User, Guest and Admin.

app/controllers/api/v1/admin/roles_controller.rb
```diff
@@ -63,6 +63,9 @@ def update
         # Returns: { data: Array[serializable objects] , errors: Array[String] }
         # Does: Deletes a role.
         def destroy
+          default_roles = %w[User Administrator Guest].include?(@role.name)
```
KH-Amir-TN, line 66: For verbosity and readibility,
It may be better to either call leave the logic as is but rename the `default_roles` to `is_default_role` since that what the variable will store and not the list of default roles,
Or, you keep its name but assign only the list to it and make the check on L67,

app/javascript/hooks/mutations/admin/roles/useDeleteRole.jsx
```diff
@@ -12,7 +12,13 @@ export default function useDeleteRole({ role, onSettled }) {
...
+          toast.error(t('toast.error.roles.role_assigned'));
```
KH-Amir-TN, line 17: I'd vote against this,
An internal server error `500` isn't limited to one singular specific case,
If we wanna make sense of the case where a failed delete attepmt is because a role is still in use (which I believe what we should do to avoid giving false negatives which gives a terrible UX).
The API need to be specific by returning something that is in that context unique to this scenario wither through a unique response status code or if it's not feasible or meaninful it can provide an identifier in the response data,

scouillard: Would something like 403 Forbidden be more acceptable?

KH-Amir-TN: Maybe, but I think that it's used in the same context to indicate insufficient permissions?

scouillard: 405 Method Not Allowed would be perfect for this case.

app/controllers/api/v1/admin/roles_controller.rb
scouillard: Ahh I see what you mean

app/javascript/hooks/mutations/admin/roles/useDeleteRole.jsx
KH-Amir-TN: Umm, it seems to be used natively by the web servers and clients.
Was wondering about the implication of the direct usage of it without specifying the `Allow` header,
But after some searching it does not seem to be an issue,
So yeah we can use that,

farhatahmad (approved):

app/javascript/components/admin/roles/forms/EditRoleForm.jsx
```diff
@@ -33,6 +35,21 @@ export default function EditRoleForm({ role }) {
     [role.name],
   );
 
+  function deleteRoleButton() {
+    if (role.name === 'User' || role.name === 'Administrator' || role.name === 'Guest') {
```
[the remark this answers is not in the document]
scouillard: Yeah agreed this could be improved.

drudgede:
I noticed that you could simply rename these roles when I tried to translate them into German (e.g. User into Nutzer). In this case, the delete button reappears. Additionally, the lock-icon disappears to in the list of roles. To be honest, I was not confident enough to press the delete button to see if the role is actually deleted because I only have one production instance. 😅

However, this is a problem because the default user role is empty in the Greenlight settings page when you rename the "User" role. Not sure what happens for new users in this case.

I suggest to make the name of the undeletable roles unrenamable (or at least use an internal ID for the logic, if present).

PR #3975, opened 2022-10-06
Add new UserBoard icon

scouillard:
## Description

Add UserBoard icon.

image

## TODO
Need to find a way to add shared by User in Room without messing the layout - or fix the layout itself.
If we add a line, the height will be different in Room.
image


app/javascript/components/rooms/MeetingBadges.jsx
```diff
@@ -3,17 +3,31 @@ import { Stack } from 'react-bootstrap';
...
   return (
     <Stack direction="horizontal" gap={0} className="room-card-badges">
-      <MeetingStatusBadge />
-      { count >= 1
-        && <MeetingParticipantsBadge count={count} />}
+      { online
+        && (
+          <>
+            <MeetingStatusBadge />
+            <MeetingParticipantsBadge count={count} />
```
KH-Amir-TN, line 15: Are we sure that we'd always have the attendees count when the room is online for all BBB servers??
We can be pessimistic about this since it will not cause any added complexity or reduce readability where we show each badge if its related condition was met and any input is valid,

KH-Amir-TN (approved):
# :fire: :fire: :fire:

app/javascript/components/rooms/MeetingBadges.jsx
scouillard: Yeah agreed. It was like that before but I felt it was getting a bit crazy in there. I'll see what I can do.

scouillard:
I will be adding the new icon to this PR.

app/javascript/components/rooms/RoomsList.jsx
```diff
@@ -35,11 +35,12 @@ export default function RoomsList() {
         {
           // eslint-disable-next-line react/no-array-index-key
           (isLoading && [...Array(8)].map((val, idx) => <Col key={idx} className="mt-0 mb-4"><RoomPlaceHolder /></Col>))
-          || rooms.map((room) => (
-            <Col key={room.friendly_id} className="mt-0 mb-4">
-              {(room.optimistic && <RoomPlaceHolder />) || <RoomCard room={room} />}
-            </Col>
-          ))
+          || rooms?.sort((a, b) => b.online - a.online)
```
KH-Amir-TN, line 38: Can't we sort rooms in the backend?

app/javascript/components/rooms/UserBoardIcon.jsx
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable max-len */
+import React from 'react';
+
+export default function UserBoardIcon() {
+  return (
+    <svg
```
KH-Amir-TN, line 6: :fire: 
We can further leverage this in other areas so it makes more sense to be placed in the `shared_components/icons` directory maybe?

KH-Amir-TN (approved):
# 🔥 🔥 🔥

app/javascript/components/rooms/UserBoardIcon.jsx
scouillard: I don't think it is necessary until we plan use it somewhere else.

app/javascript/components/rooms/RoomsList.jsx
scouillard: Yeah okay

scouillard: Might as well indeed

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -30,6 +30,7 @@ def index
         end
 
         rooms = user_rooms + shared_rooms
+        rooms.sort_by { |room| room.online ? 0 : 1 }
```
farhatahmad, line 33: Shouldn't we do this after calling the RunningMeeting service?

farhatahmad: We'll need to find a better way to handle the sorting and the queries. I'll spend some time later cleaning it up

scouillard: Right.

scouillard: Should we sort via the frontend as it was initially?

farhatahmad:
@scouillard Let me know when the shared icon is switched back

scouillard:
> @scouillard Let me know when the shared icon is switched back

Done

PR #3976, opened 2022-10-06
Add sorting to Rooms

scouillard:
## Description

Add sorting by meeting status to rooms.

### DIFF
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -22,9 +22,9 @@ module Api
       # GET /api/v1/rooms.json
       def index
         # Return the rooms that belong to current user
-        user_rooms = current_user.rooms.search(params[:search])
+        user_rooms = current_user.rooms.order(online: :desc).search(params[:search])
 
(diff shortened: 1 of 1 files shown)


app/controllers/api/v1/rooms_controller.rb
```diff
@@ -22,9 +22,9 @@ class RoomsController < ApiController
       # GET /api/v1/rooms.json
       def index
         # Return the rooms that belong to current user
-        user_rooms = current_user.rooms.search(params[:search])
+        user_rooms = current_user.rooms.order(online: :desc).search(params[:search])
 
-        shared_rooms = current_user.shared_rooms.search(params[:search]).map do |room|
+        shared_rooms = current_user.shared_rooms.order(online: :desc).search(params[:search]).map do |room|
```
farhatahmad, line 27: We should be able to order by more than one value - it  should be `online: :desc` first and then `created_at: :desc` next

scouillard: I think I'd add `last_session: :desc` as second sorting parameter then.

farhatahmad (approved):

PR #3977, opened 2022-10-07
Change flow of require authentication

farhatahmad:

image

### DIFF
--- a/app/controllers/external_controller.rb
+++ b/app/controllers/external_controller.rb
@@ -29,6 +29,9 @@ class ExternalController < ApplicationController
 
     # TODO: - Ahmad: deal with errors
 
+    redirect_location = cookies[:location]
+    return redirect_to redirect_location if redirect_location&.match?('\/rooms\/\w{3}-\w{3}-\w{3}-\w{3}\/join')
+
     redirect_to '/rooms'
   end
 
--- a/app/controllers/external_controller.rb
+++ b/app/controllers/external_controller.rb
@@ -30,6 +30,7 @@ class ExternalController < ApplicationController
     # TODO: - Ahmad: deal with errors
 
     redirect_location = cookies[:location]
+    cookies.delete(:location)
     return redirect_to redirect_location if redirect_location&.match?('\/rooms\/\w{3}-\w{3}-\w{3}-\w{3}\/join')
 
     redirect_to '/rooms'
--- a/app/controllers/external_controller.rb
+++ b/app/controllers/external_controller.rb
@@ -28,10 +28,9 @@ class ExternalController < ApplicationController
     session[:session_token] = user.session_token
 
     # TODO: - Ahmad: deal with errors
-
     redirect_location = cookies[:location]
     cookies.delete(:location)
-    return redirect_to redirect_location if redirect_location&.match?('\/rooms\/\w{3}-\w{3}-\w{3}-\w{3}\/join')
+    return redirect_to redirect_location if redirect_location&.match?('^\/rooms\/\w{3}-\w{3}-\w{3}-\w{3}\/join$')
 
     redirect_to '/rooms'
   end
--- a/app/javascript/components/rooms/room/join/RequireAuthentication.jsx
+++ b/app/javascript/components/rooms/room/join/RequireAuthentication.jsx
@@ -1,0 +1,52 @@
+import React, { useEffect } from 'react';
+import { Button, Form } from 'react-bootstrap';
+import Card from 'react-bootstrap/Card';
+import { useTranslation } from 'react-i18next';
+import { useLocation } from 'react-router-dom';
+import Logo from '../../../shared_components/Logo';
+import useEnv from '../../../../hooks/queries/env/useEnv';
+import Spinner from '../../../shared_components/utilities/Spinner';
+import ButtonLink from '../../../shared_components/utilities/ButtonLink';
+
+export default function RequireAuthentication() {
+  const { t } = useTranslation();
+  const location = useLocation();
+  const { isLoading, data: env } = useEnv();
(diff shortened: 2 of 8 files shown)


app/javascript/components/rooms/room/join/RequireAuthentication.jsx
```diff
@@ -0,0 +1,52 @@
...
+          {
+            env.OPENID_CONNECT ? (
+              <Form action="/auth/openid_connect" method="POST" data-turbo="false">
+                <input type="hidden" name="authenticity_token" value={document.querySelector('meta[name="csrf-token"]').content} />
+                <Button variant="brand-outline-color" className="btn btn-lg m-2" type="submit">{t('authentication.sign_up')}</Button>
+                <Button variant="brand" className="btn btn-lg m-2" type="submit">{t('authentication.sign_in')}</Button>
+              </Form>
+            ) : (
+              <>
+                <ButtonLink to={`/signup?location=${path}`} variant="brand-outline-color" className="btn btn-lg m-2">
```
KH-Amir-TN, line 41: I think the use of the history state here may achieve the desired outcome without the use of the search param  `location`.

app/javascript/components/rooms/room/join/RoomJoin.jsx
```diff
@@ -163,10 +163,10 @@ export default function RoomJoin() {
       {
         env.OPENID_CONNECT ? (
```
KH-Amir-TN, line 164: Don't we need to redirect users back to the join page for these cases?

app/controllers/external_controller.rb
```diff
@@ -29,6 +29,10 @@ def create_user
 
     # TODO: - Ahmad: deal with errors
 
+    redirect_location = cookies[:location]
+    cookies.delete(:location)
+    return redirect_to redirect_location if redirect_location&.match?('\/rooms\/\w{3}-\w{3}-\w{3}-\w{3}\/join')
```
KH-Amir-TN, line 34: The regex isn't specific enough,
A redirection will be made to "https://www.google.com?ignore=/rooms/iam-abl-eto-pas/join" or anything similar,
However, rails `redirect_to` will catch that and raise `ActionController::Redirecting::UnsafeRedirectError` but we shouldn't be counting on that,

spec/controllers/external_controller_spec.rb
```diff
@@ -43,6 +43,45 @@
...
+        get :create_user, params: { provider: 'openid_connect' }
+
+        expect(response).to redirect_to('/rooms/o5g-hvb-s44-p5t/join')
+      end
+
+      it 'doesnt redirect if it doesnt match a room joins format' do
+        request.env['omniauth.auth'] = OmniAuth.config.mock_auth[:openid_connect]
+
+        cookies[:location] = {
+          value: 'https://google.com',
```
KH-Amir-TN, line 63: https://www.google.com?ignore=/rooms/iam-abl-eto-pas/join

app/javascript/hooks/mutations/sessions/useCreateSession.jsx
```diff
@@ -8,14 +8,18 @@ export default function useCreateSession(token) {
...
 
   return useMutation(
     (session) => axios.post('/sessions.json', { session, token }),
     {
       onSuccess: (response) => {
         queryClient.invalidateQueries('useSessions');
         // if the current user does NOT have the CreateRoom permission, then do not re-direct to rooms page
-        if (response.data.data.permissions.CreateRoom === 'false') {
+        if (redirect) {
+          navigate(redirect);
```
KH-Amir-TN, line 21: I had to deal with this and still haven't found a clean way to approach it,
The call to navigate may preceede react state updates and therefore render a component before the contexts updates as intended,
This can render a page before the Auth context gets updated, e.g. it will fail if you'd redirect to '/profile' (same applies even if we manually set the currentUser to its queryCache)

app/controllers/external_controller.rb
KH-Amir-TN: Maybe `%r{\A/rooms/\w{3}-\w{3}-\w{3}-\w{3}/join\z}`?

app/javascript/components/rooms/room/join/RoomJoin.jsx
farhatahmad: Yeah that's a good point - I'll fix this up

app/javascript/components/rooms/room/join/RequireAuthentication.jsx
farhatahmad: `useHistory` is tricky to use in this case, because we'd need to convert our links to use the history state instead of what we currently do. I'll leave it as is for now and maybe we can look at refactoring at a later point

app/javascript/components/rooms/room/join/RoomJoin.jsx
```diff
@@ -1,25 +1,25 @@
...
+import useRoomStatus from '../../../../hooks/mutations/rooms/useRoomStatus';
+import useEnv from '../../../../hooks/queries/env/useEnv';
+import { joinFormConfig, joinFormFields as fields } from '../../../../helpers/forms/JoinFormHelpers';
+import subscribeToRoom from '../../../../channels/rooms_channel';
+import RequireAuthentication from './RequireAuthentication';
+import GGSpinner from '../../../shared_components/utilities/GGSpinner';
+import Logo from '../../../shared_components/Logo';
+import Avatar from '../../../users/user/Avatar';
+import Form from '../../../shared_components/forms/Form';
+import FormControl from '../../../shared_components/forms/FormControl';
```
farhatahmad, line 22: some crazy stuff from the merge conflict - you can ignore it

scouillard (approved):

PR #3981, opened 2022-10-07
Add GG Spinner

scouillard:
## Description

Not sure if you guys will like it.
Found this while browsing icon libraries.
Propose to replace the Spinner component with this one.
Considering all the isLoading spinners will be removed when we have the placeholders.
This spinner matches the spinner in the mock ups - it is nicer than Bootstrap's spinner.
It's easier to add those classes than to try and customize Bootstrap's spinner.

image

### DIFF
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -451,3 +451,37 @@ input[type="range"]:focus::-moz-range-thumb {
     }
   }
 }
+
+//GG spinner
+.gg-spinner {
+  transform: scale(var(--ggs,1))
+}
+.gg-spinner,
+.gg-spinner::after,
+.gg-spinner::before {
+  box-sizing: border-box;
+  position: relative;
+  display: block;
+  width: 40px;
+  height: 40px
+}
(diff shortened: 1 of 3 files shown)


KH-Amir-TN (approved):
That indeed looks better!
It can be used in similar cases, but if we're gonna stick to spinners for fine grained feedbacks (submit button loading,...) the GG spinner may require some tweaks to make it reusable for those scenarios too (e.g. it's a block element we may need it to render an inline element for some cases, we'd need to specify the size,...).

scouillard:
> That indeed looks better! It can be used in similar cases, but if we're gonna stick to spinners for fine grained feedbacks (submit button loading,...) the GG spinner may require some tweaks to make it reusable for those scenarios too (e.g. it's a block element we may need it to render an inline element for some cases, we'd need to specify the size,...).

Agreed. Checking out with you guys before implementing it app wide.

scouillard:
> In okay with this as well, but I'd recommend we keep both. I like this for join, but prefer the other style for "Loading"



> In okay with this as well, but I'd recommend we keep both. I like this for join, but prefer the other style for "Loading"

Sounds good to me

PR #3982, opened 2022-10-07
V2 roles migration: Added `Migrations::create_role` API.

KH-Amir-TN:
```
+Added `Migrations::create_role` API.
+ Added  `#decrypted_params` that will decrypt and verify the request data.
TODO:
  - Add specs.
```

### DIFF
--- a/app/controllers/api/v1/migrations/external_controller.rb
+++ b/app/controllers/api/v1/migrations/external_controller.rb
@@ -4,6 +4,44 @@ module Api
   module V1
     module Migrations
       class ExternalController < ApiController
+        skip_before_action :verify_authenticity_token
+        skip_before_action :ensure_authenticated
+
+        rescue_from ActiveSupport::MessageEncryptor::InvalidMessage do |exception|
+          log_exception exception
+          render_error errors: [Rails.configuration.custom_error_msgs[:missing_params]], status: :bad_request
+        end
+
+        # POST /api/v1/migrations/roles.json
+        # Expects: { role: { :name } }
+        # Returns: { data: Array[serializable objects(recordings)] , errors: Array[String] }
+        # Does: Creates a role.
+        def create_role
+          role = Role.new role_params.merge(provider: 'greenlight')
+
+          return render_error status: :bad_request unless role.save
+
+          render_data status: :created
+        end
+
+        private
+
+        def role_params
+          decrypted_params.require(:role).permit(:name)
+        end
+
+        def decrypted_params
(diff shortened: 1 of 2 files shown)


KH-Amir-TN:
# DEPENDS ON #3980

app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+
+          raise ActiveSupport::MessageEncryptor::InvalidMessage unless encrypted_params.is_a? String
+
+          crypt = ActiveSupport::MessageEncryptor.new(Rails.application.secrets.secret_key_base[0..31])
+          decrypted_params = crypt.decrypt_and_verify(encrypted_params) || {}
+
+          raise ActiveSupport::MessageEncryptor::InvalidMessage unless decrypted_params.is_a? Hash
+
+          ActionController::Parameters.new(decrypted_params)
+        end
```
KH-Amir-TN, line 44: The `decrypted_params` method is an API for validating and decrypting the encrypted request parameters while it can be used in place of the regular `params` method.
And it:
1. Validates the structure of the request params, any valid request should respect the schema `:v2 => {:encrypted_params => String}`
2. Extracts the `encrypted_params` string if 1. passes.
3. Decrypts and verifies the `encrypted_params` ciphertext.
4. If 3. passes it will further verify the serialized plaintext `decrypted_params` is a `Hash`.
5. It will form in an `ActionController::Parameters` instance out of the `decrypted_params` hash.

Output:
1. It can raise `ActiveSupport::MessageEncryptor::InvalidMessage` and `ActionController::ParameterMissing` (in both cases the v3 will log the exception  and render_error a :bad_request) for invalid inputs.
2. It returns a params instance of the decrypted params or an empty hash `{}`.

KH-Amir-TN: So one can add in a new action and just use the `decrypted_params` in place of the regular `params` method which will enable the action with cryptographic verification and decryption while providing a regular params object seamlessly.

app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -4,6 +4,44 @@ module Api
...
+
+          return render_error status: :bad_request unless role.save
+
+          render_data status: :created
+        end
+
+        private
+
+        def role_params
+          decrypted_params.require(:role).permit(:name)
```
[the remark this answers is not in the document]
farhatahmad: :fire:

farhatahmad (approved):

PR #3990, opened 2022-10-12
V2 Resources Migration: Adding`Migrations::create_user` API [WIP].

KH-Amir-TN:


hadicheaito1:
So for this, there would be a rake task in v2 that goes through the users and sends the request to this api endpoint to create the users (assuming there is both deployments of v2 and v3 running)?

KH-Amir-TN:
> So for this, there would be a rake task in v2 that goes through the users and sends the request to this api endpoint to create the users (assuming there is both deployments of v2 and v3 running)?

Exactly,

PR #3991, opened 2022-10-12
Add role validation

scouillard:

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -8,6 +8,8 @@ class Role < ApplicationRecord
   validates :name, presence: true
   validates :provider, presence: true
 
+  validates :name, uniqueness: { scope: :provider }
+
   before_validation :set_random_color, on: :create
 
   scope :with_provider, ->(current_provider) { where(provider: current_provider) }
--- a/spec/factories/role_factory.rb
+++ b/spec/factories/role_factory.rb
@@ -2,7 +2,7 @@
 
 FactoryBot.define do
   factory :role do
-    name { Faker::Construction.role }
+    name { Faker::Lorem.unique.characters(number: 10) }
     provider { 'greenlight' }
   end
 end
--- a/db/migrate/20221012200427_add_index_to_roles.rb
+++ b/db/migrate/20221012200427_add_index_to_roles.rb
@@ -1,0 +1,5 @@
+class AddIndexToRoles < ActiveRecord::Migration[7.0]
+  def change
+    add_index :roles, %i[name provider], unique: true
+  end
+end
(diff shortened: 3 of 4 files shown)


spec/factories/role_factory.rb
```diff
@@ -2,7 +2,7 @@
 
 FactoryBot.define do
   factory :role do
-    name { Faker::Construction.role }
+    name { Faker::Lorem.unique.characters(number: 10) }
```
scouillard, line 5: It's either that or we clear the cache in every single spec.

image

scouillard: Eh after all it might be worth it to clear the cache in every file. I prefer the Construction role name as it helps a lot when debugging.

scouillard: Change Construction.role to Job.title which has the same effect but has a bigger range.

KH-Amir-TN (dismissed):
## :fire: :fire: :fire:

app/models/role.rb
```diff
@@ -8,6 +8,8 @@ class Role < ApplicationRecord
   validates :name, presence: true
   validates :provider, presence: true
 
+  validates :name, uniqueness: { scope: :provider }
```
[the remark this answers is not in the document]
scouillard: Oh yeah good catch

app/models/user.rb
```diff
@@ -174,4 +176,10 @@ def activate!
   def deactivate!
     update! active: false
   end
+
+  def check_user_role_provider
+    return unless role
+
+    errors.add(:user_provider, 'has to be the same as the Role provider') if provider != role.provider
+  end
```
farhatahmad, line 184: I think this is too broad of a check. Can we move this check to `users#update` instead?

scouillard: What do you mean by too broad?
If we add the logic to #update instead, shouldn't we add it to #create as well?
IMO this is redundant, we have specs for that already, it shouldn't happen. 
I feel like this belongs in a Validation.

farhatahmad: Yeah but this check will run on every change to user. We might want to change if to run only if the role is changing, or only do the check in the controller instead

app/models/user.rb
```diff
@@ -174,4 +176,10 @@ def activate!
   def deactivate!
     update! active: false
   end
+
+  def check_user_role_provider
+    return unless role && role_changed?
```
farhatahmad, line 181: We can actually move this check up to the `validate` line. So `validate :check_user_role_provider if role_changed?`

scouillard: Nice

PR #3992, opened 2022-10-13
Fix User join his own room as a moderator

scouillard:
## Description

You would join your own meeting as a regular viewer if the meeting was started by a shared user.
Now you will join all your meetings as an admin.

## Testing Steps

1. Create a room with user1
2. Share the room with user2
3. Go on incognito and log in user2
4. Start the shared room with user2
5. Join the meeting with user1
6. You should be logged in as a moderator

### DIFF
--- a/app/controllers/api/v1/meetings_controller.rb
+++ b/app/controllers/api/v1/meetings_controller.rb
@@ -90,7 +90,8 @@ module Api
       def authorized_as_moderator?(access_code:)
         (params[:access_code].present? && access_code == params[:access_code]) ||
           @room.anyone_joins_as_moderator? ||
-          current_user&.shared_rooms&.include?(@room)
+          current_user&.shared_rooms&.include?(@room) ||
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/meetings_controller.rb
```diff
@@ -90,7 +90,8 @@ def authorized_as_viewer?(access_code:)
       def authorized_as_moderator?(access_code:)
         (params[:access_code].present? && access_code == params[:access_code]) ||
           @room.anyone_joins_as_moderator? ||
-          current_user&.shared_rooms&.include?(@room)
+          current_user&.shared_rooms&.include?(@room) ||
+          current_user&.rooms&.include?(@room)
```
farhatahmad, line 94: You can just do `@room.user_id == current_user.id` to avoid doing another query

scouillard: right thanks

farhatahmad (approved):

PR #3993, opened 2022-10-13
Improve Rooms query

scouillard:


scouillard:
It is possible to do a single query instead of two, as you suggested @farhatahmad 

```
 rooms = Room
             .joins(:shared_accesses)
             .where(user_id: current_user.id)
             .or(Room.joins(:shared_accesses)
                     .where(shared_accesses: { user_id: current_user.id }))
             .order(online: :desc, last_session: :desc)
             .search(params[:search])
             .distinct
```

farhatahmad (approved):

PR #3998, opened 2022-10-14
Fix Rooms sorting (NULLS FIRST is the default for DESC)

scouillard:
## Description

Fix small issue with Rooms sorting where a room with last_session: null would appear at the top of the list.
NULLS FIRST is the default for DESC in PG.
I think we would rather have those at the bottom of the rooms list and have most recently used rooms at the top of the list.

image

### DIFF
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -24,7 +24,8 @@ module Api
         shared_rooms = SharedAccess.where(user_id: current_user.id).select(:room_id)
         rooms = Room.where(user_id: current_user.id)
                     .or(Room.where(id: shared_rooms))
-                    .order(Room.arel_table[:online].desc, Room.arel_table[:last_session].desc.nulls_last)
+                    .order(online: :desc)
+                    .order(Room.arel_table[:last_session].desc.nulls_last)
                     .search(params[:search])
 
         rooms.map do |room|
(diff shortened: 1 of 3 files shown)


hadicheaito1 (approved):

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -24,7 +24,7 @@ def index
         shared_rooms = SharedAccess.where(user_id: current_user.id).select(:room_id)
         rooms = Room.where(user_id: current_user.id)
                     .or(Room.where(id: shared_rooms))
-                    .order(online: :desc, last_session: :desc)
+                    .order(Room.arel_table[:online].desc, Room.arel_table[:last_session].desc.nulls_last)
```
KH-Amir-TN, line 27: That seems to be the best solution,
However, we should use `Arel` only when it's required.
 The `online` sorting can use the regular hash syntax.
`.order(online: :desc).order(Room.arel_table[:last_session].desc.nulls_last)`

KH-Amir-TN (approved):

app/controllers/api/v1/rooms_controller.rb
farhatahmad: Yeah agreed

scouillard: Didnt think about chaining the `.order` thank you!

KH-Amir-TN: You're welcome,

farhatahmad (approved):

PR #4000, opened 2022-10-17
Delete location cookie when leaving the page

farhatahmad:

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,6 +5,8 @@ class ApplicationController < ActionController::Base
 
   # Returns the current signed in User (if any)
   def current_user
+    return @current_user if @current_user
+
     # Overwrites the session cookie if an extended_session cookie exists
     session[:session_token] ||= cookies.encrypted[:_extended_session]['session_token'] if cookies.encrypted[:_extended_session].present?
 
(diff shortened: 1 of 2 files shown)


app/javascript/components/rooms/room/join/RoomJoin.jsx
```diff
@@ -38,10 +38,16 @@ export default function RoomJoin() {
   const path = encodeURIComponent(location.pathname);
 
   useEffect(() => { // set cookie to return to if needed
-    document.cookie = `location=${path};path=/`;
-  });
+    document.cookie = `location=${path};path=/;`;
+
+    return () => { // delete redirect location when unmounting
+      document.cookie = 'location=;path=/;expires=Thu, 01 Jan 1970 00:00:00 GMT';
+    };
```
farhatahmad, line 45: Will run on `componentWillUnmount`

app/javascript/components/rooms/room/join/RoomJoin.jsx
```diff
@@ -38,10 +38,16 @@ export default function RoomJoin() {
...
+  }, []);
 
   const handleJoin = (data) => {
+    document.cookie = 'location=;path=/;expires=Thu, 01 Jan 1970 00:00:00 GMT'; // delete redirect location
```
farhatahmad, line 49: Joining for some reason does not unmount the component, so adding an addition check here

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,8 @@ class ApplicationController < ActionController::Base
 
   # Returns the current signed in User (if any)
   def current_user
+    return @current_user if @current_user
```
farhatahmad, line 8: Will only query current_user once per request

scouillard (approved):

PR #4001, opened 2022-10-17
Resources migrations: Added `migrations:roles` task.

KH-Amir-TN:

### DIFF
--- a/sample.env
+++ b/sample.env
@@ -16,6 +16,13 @@ SECRET_KEY_BASE=
 BIGBLUEBUTTON_ENDPOINT=
 BIGBLUEBUTTON_SECRET=
 
+# The endpoint and "SECRET_KEY_BASE" for your Greenlight v3 instance.
+# Set these if you are trying to migrate your resources to v3.
+# Example:
+#V3_ENDPOINT=https://v3.greenlight.test/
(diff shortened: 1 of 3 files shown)


KH-Amir-TN:
Things that we might like to add:
1. Add a read timeout adequate with the expiration period (Furthermore, we can add retrial in such cases).
2. Give admins the ability to specify a range for partial migrations.
What else you think about guys?

sample.env
```diff
@@ -16,6 +16,13 @@ SECRET_KEY_BASE=
 BIGBLUEBUTTON_ENDPOINT=
 BIGBLUEBUTTON_SECRET=
 
+# The endpoint and shared secret (first 31 characters of V3 secret) for your Greenlight v3 server.
```
farhatahmad, line 19: This should say `secret_key_base` and we should be taking all of the characters and cutting off all unused chars ourself

KH-Amir-TN: Was thinking to leverage the size limit to protect against direct usage of the secret  in case of an exposure.

But since we'll be requesting admins to change v3 secret after migrating their resources we can place the key base itself.

farhatahmad (approved):

PR #4003, opened 2022-10-17
Add Shared Access search restriction (3 characters) + add scrollbar 

scouillard:
## Description

Add restriction to Shared Access user search: need at least 3 characters to initiate the query.
Add scrolling bar to ShareableUsers modal.
Did some fine tuning to fix the height of the modal, and set the trigger height for the scrollbar.
The height is tuned to accommodate for 4 rows of users.
More than 4 rows will trigger the scrollbar.
Less than 4 rows there will be a gap.

## Testing Steps

1. Create a bunch of Users with at least (3) matching characters.
2. Use the SharedAccess search box.
3. Test the UX

image
image
image
image

## TODO
Next PRs:

- Fix Debouncing
- Add pagination to Shared Users?

### DIFF
--- a/app/controllers/api/v1/shared_accesses_controller.rb
+++ b/app/controllers/api/v1/shared_accesses_controller.rb
@@ -30,7 +30,7 @@ module Api
 
       # GET /api/v1/shared_accesses/friendly_id/shareable_users.json
       def shareable_users
-        return render_error status: :bad_request unless params[:search].length >= 3
+        return render_error status: :bad_request unless params[:search].present? && params[:search].length >= 3
 
         # role_id of roles that have SharedList permission set to true
         role_ids = RolePermission.joins(:permission).where(permission: { name: 'SharedList' }).where(value: 'true').pluck(:role_id)
--- a/app/controllers/api/v1/shared_accesses_controller.rb
+++ b/app/controllers/api/v1/shared_accesses_controller.rb
@@ -30,6 +30,8 @@ module Api
 
       # GET /api/v1/shared_accesses/friendly_id/shareable_users.json
       def shareable_users
+        return unless params[:search].length >= 3
+
         # role_id of roles that have SharedList permission set to true
         role_ids = RolePermission.joins(:permission).where(permission: { name: 'SharedList' }).where(value: 'true').pluck(:role_id)
 
--- a/app/controllers/api/v1/shared_accesses_controller.rb
+++ b/app/controllers/api/v1/shared_accesses_controller.rb
@@ -34,10 +34,14 @@ module Api
         role_ids = RolePermission.joins(:permission).where(permission: { name: 'SharedList' }).where(value: 'true').pluck(:role_id)
 
         # Can't share the room if it's already shared or it's the room owner
-        render_data data: User.with_attached_avatar
+        shareable_users = User.with_attached_avatar
                               .where.not(id: [@room.shared_users.pluck(:id) << @room.user_id])
                               .where(role_id: [role_ids])
-                              .search(params[:search]), status: :ok
+                              .search(params[:search])
+
+        pagy, shareable_users = pagy(shareable_users)
+
+        render_data data: shareable_users, meta: pagy_metadata(pagy), status: :ok
       end
 
       private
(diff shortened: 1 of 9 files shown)


hadicheaito1 (dismissed):

app/controllers/api/v1/shared_accesses_controller.rb
```diff
@@ -30,14 +30,18 @@ def show
 
       # GET /api/v1/shared_accesses/friendly_id/shareable_users.json
       def shareable_users
+        return render_error status: :bad_request unless params[:search].length >= 3
```
KH-Amir-TN, line 33: It's better to check first that `params[:serarch]` is present,

app/javascript/components/rooms/room/shared_access/forms/SharedAccessForm.jsx
```diff
@@ -19,47 +19,56 @@ export default function SharedAccessForm({ handleClose }) {
...
+                          <td>
+                            <span className="text-muted"> {user.email} </span>
+                          </td>
+                        </tr>
+                      )));
+                  } if (isThreeCharacters && !shareableUsers?.length) {
+                    return (<tr className="fw-bold"><td>{ t('user.no_user_found') }</td><td /></tr>);
+                  }
+                  return (<tr className="fw-bold"><td colSpan="2">{ t('user.type_three_characters') }</td></tr>);
+                })()
```
KH-Amir-TN, line 67: I think this can be improved,

app/javascript/hooks/queries/shared_accesses/useShareableUsers.jsx
```diff
@@ -1,16 +1,17 @@
...
     search: input,
   };
 
   return useQuery(
-    ['getShareableUsers', input],
+    ['getShareableUsers', { ...params }],
     () => axios.get(`/shared_accesses/${friendlyId}/shareable_users.json`, { params }).then((resp) => resp.data.data),
     {
       keepPreviousData: true,
+      enabled: isThreeCharacters,
```
KH-Amir-TN, line 14: `isThreeCharacters` seems to be an information on the `input` data,
Therefore `useShareableUsers` can extract it without the need to pass in an extra data?

scouillard: It's also used in the view.

scouillard: You would rather duplicate the logic inside the hook?

KH-Amir-TN: I think I'd do that,
First the logic here isn't expensive nor complex,
The `isThreeCharacters` itself will be provided in accordance to the `input`,
Whenever the latter changes the other changes,
The complexity of adding that extra variable in memory may exceed what you're trying to gain in terms of performance,

scouillard: I'm sorry - I don't understand what you mean.

Is adding an extra param, an extra variable in memory that big of a deal?

I'm not a fan of duplicating the logic of the view inside the hook.

scouillard: I think I see what you mean. Because isThreeChar is dependant on input. And both are given to the hook as params. Yeah ok.

I still think adding the variable makes senses and makes it more enjoyable to read/understand. I didn't add it for performance reasons.

KH-Amir-TN: What you're suggesting is good and I'd go that way myself,
However, the notes are caused by the naming schema.
By indicating `isThreeCharacters` we've already gave it a sense and that made a inner dependency for the given arguments (`isThreeCharachters` depends on `input`) and that what led me to spot this in the first place.
However, if you'd simply call that `enabled` instead of `isThreeCharachters` on the hook signature - that would brake the dependency logic.
It's true that `enabled` will map to `isThreeCharachters` in value  but only in this context which is irrelevant to the hook, It's semantically different,
If we're calling it as `isThreeCharachters` in the hook then it's an information on the `input` that we can extract with no complexity,
To further understand this, think about the implication of both schemas when reusing the hook in another context.
One can map the `enabled` to any value to enable/disable the mutation and another will be tight to the what `isThreeCharachers` means.

farhatahmad: Yeah agreed. I would just check the input length directly in the mutation. No need to pass it in

farhatahmad (approved):

PR #4004, opened 2022-10-17
Add Recordings count badge

scouillard:
## Description

Add Recording Count badge as per mock-up.

I figured it was more convenient to add an endpoint for the recordings count, instead of doing an expensive query for the whole Recordings#index in Rooms for (potentially) no reason.

Can do the same thing in Room/Recording if you guys like this solution ie create a room_recording_count endpoint in rooms.

image

### DIFF
--- a/app/controllers/api/v1/recordings_controller.rb
+++ b/app/controllers/api/v1/recordings_controller.rb
@@ -70,6 +70,11 @@ module Api
         render_data status: :ok
       end
 
+      def recordings_count
+        count = current_user.recordings.count
+        render_data data: count, status: :ok
+      end
+
       private
 
       def recording_params
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -487,4 +487,3 @@ input[type="range"]:focus::-moz-range-thumb {
   0% { transform: rotate(0deg) }
   to { transform: rotate(359deg) }
 }
-
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -446,6 +446,8 @@ input[type="range"]:focus::-moz-range-thumb {
   }
 
   .nav-link {
+    padding-left: 8px !important;
+    padding-right: 8px !important;
     &:hover {
       border-bottom: 2px solid var(--brand-color) !important;
     }
@@ -485,3 +487,4 @@ input[type="range"]:focus::-moz-range-thumb {
   0% { transform: rotate(0deg) }
   to { transform: rotate(359deg) }
 }
+
--- a/app/assets/stylesheets/recordings.scss
+++ b/app/assets/stylesheets/recordings.scss
@@ -35,3 +35,7 @@
   cursor: pointer;
 }
 
+.recordings-count-badge {
+  background-color: whitesmoke !important;
+}
+
--- a/app/javascript/components/recordings/RecordingsCountTab.jsx
+++ b/app/javascript/components/recordings/RecordingsCountTab.jsx
@@ -1,0 +1,25 @@
+import React from 'react';
+import { Badge, Stack } from 'react-bootstrap';
+import PropTypes from 'prop-types';
+
+export default function RecordingsCountTab({ count }) {
+  return (
+    <Stack direction="horizontal" gap={0}>
+      <span> Recordings </span>
+      { count > 0
+        && (
+          <Badge className="rounded-pill recordings-count-badge fw-normal ms-2 text-brand">
+            { count }
+          </Badge>
+        )}
+    </Stack>
+  );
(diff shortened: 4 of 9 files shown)


hadicheaito1:
Looks super clean. 

If possible (i think it is), it would be better if we could include the count as part of the response in the existing endpoint that gets the recordings (index in recordings controller). That way we avoid having to add all this code as well as having to make an extra query

scouillard:
> Looks super clean.
> 
> If possible (i think it is), it would be better if we could include the count as part of the response in the existing endpoint that gets the recordings (index in recordings controller). That way we avoid having to add all this code as well as having to make an extra query

That's the point: to avoid doing recordings#index because it's too bulky for what we need. Else we could simply call the og useRecordings and do `recordings.length` on the frontend.

scouillard:
> > Looks super clean.
> > If possible (i think it is), it would be better if we could include the count as part of the response in the existing endpoint that gets the recordings (index in recordings controller). That way we avoid having to add all this code as well as having to make an extra query
> 
> That's the point: to avoid doing recordings#index because it's too bulky for what we need. Else we could simply call the og useRecordings and do `recordings.length` on the frontend.

hadicheaito1:
> > Looks super clean.
> 
> > 
> 
> > If possible (i think it is), it would be better if we could include the count as part of the response in the existing endpoint that gets the recordings (index in recordings controller). That way we avoid having to add all this code as well as having to make an extra query
> 
> 
> 
> That's the point: to avoid doing recordings#index because it's too bulky for what we need. Else we could simply call the og useRecordings and do `recordings.length` on the frontend.

Ohh my bad I forgot that the number is going to show even when just on the Rooms tab. Yeah in that case ignore what I said, what you have makes sense

hadicheaito1 (approved):
And yeah I think it should be added to the Room recordings as well

app/controllers/api/v1/recordings_controller.rb
```diff
@@ -70,6 +70,11 @@ def update_visibility
         render_data status: :ok
       end
 
+      def recordings_count
+        count = current_user.recordings.count
```
KH-Amir-TN, line 74: Can't we just include this to the `SessionsController#index` API and simply extract it fromt the `AuthContext`?

KH-Amir-TN: However, I think we should test this endpoint?

scouillard: > Can't we just include this to the `SessionsController#index` API and simply extract it fromt the `AuthContext`?

IMO this does not belong there. And we would need to "drill down" the value.

KH-Amir-TN: The `AuthContext` is a context, what you mean by drill down?
Contexts were introduced to resolve the "props drilling" issue if that's what you're mentioning,
My idea is that we already have queries that are sent to fetch the current user state,
Anything like the current user name, email, active status, ... along with its permissions and their values can be collected in that API call,
That state is provided to the client application through a context where any component no matter where in the subtree can subscribe to and extract the current user data,
My idea was why making a separate endpoint when we can leverage the  infrastructure that we have through `AuthContext` and via the `SessionsController#index` API.

KH-Amir-TN: I think it's the same of what we've done with the `CurrentUserSerializer` for the current user permissions,

scouillard: I'm not sure what Contexts are doing TBH.
What's the difference between calling useAuth() or useRecordingsCount() to get our value?

scouillard: The only difference I see is the memoization in AuthContext? So that the recording count would be fetch from memory in your solution? Instead of a query.

KH-Amir-TN: An API call (minimum 1) + an extra entry in our queryCache to hold the data + any work involved from react-query to manage that cache.

KH-Amir-TN: However, to be aware  if you're willing to leverage the `SessionsController#index` means that any mutation to the recordings should invoke an invalidation to the `useSessions` cache which may cause an overhead,

scouillard: Am I correct in my assumption above? 

About the useAuth() not firing an extra API call because of the memoization? And that memoization is the core of a Context?

Or the added value of your solution is mostly in the queryCache/React Query business?

farhatahmad:
Yeah I'm not quite sure about this - firing an extra query for little gain might not be worth it

farhatahmad:
I'm not sure I agree with pushing this into AuthContext - this is something strictly related to 1 page and should be refetched rather than store and memoized

farhatahmad (approved):

PR #4007, opened 2022-10-18
Admin Panel fixes

scouillard:
## Description

- Add a (kind of cheap) fix to limit Admin tabs "jumping".

    Set an arbitrary min-height to the admin panel.
    RoomConfigs is pretty tall as well so we can't use pagination to size down the tabs.
    There might a better solution than that - I don't really like setting "random" pixel height but that does it for now.

- Set Pagy to return 10 elements instead of 5.
- Fix the "jumping" in Admin/SiteSettings tabs also.
- This PR does not include the fixes to the different styling of the tables in admin panels

### DIFF
--- a/app/assets/stylesheets/admin_panel.scss
+++ b/app/assets/stylesheets/admin_panel.scss
@@ -1,3 +1,5 @@
+$admin-card-height: 969px;
+
 #admin-panel {
   padding-bottom: $header-height;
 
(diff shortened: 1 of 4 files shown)


scouillard:
Thinking about setting a min-height to Admin Tabs.
What should we use?
I try to avoid setting up random pixel values.

scouillard:
Returning 8 elements instead of 10 is a bit better in terms of design because it helps closing the (blank space) gap IMO.

scouillard:
Tried setting the min-height to the tallest tab (ManageUsers) but the gap is way too big. 
Tried setting Pagy from 10 to 8 users and the gap is still really big. 
Setting Pagy to 5 users would be better design wise but I don't think this would improve the UX (it would worsen it IMO).

farhatahmad (approved):

PR #4008, opened 2022-10-19
V2 Resources Migration: Fixed `Migrations::ExternalController#create_user`to generate password only for local users.

KH-Amir-TN:
## Description

Fixed `Migrations::ExternalController#create_user`to generate password only for local users.

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/controllers/api/v1/migrations/external_controller.rb
+++ b/app/controllers/api/v1/migrations/external_controller.rb
@@ -38,17 +38,17 @@ module Api
         # Does: Creates a user.
 
         def create_user
-          language = if user_params[:language].blank? || user_params[:language] == 'default'
-                       I18n.default_locale
-                     else
-                       user_params[:language]
-                     end
+          user_hash = user_params.to_h
 
-          role = Role.find_by(name: user_params[:role], provider: 'greenlight')
+          user_hash[:language] = I18n.default_locale if user_hash[:language].blank? || user_hash[:language] == 'default'
+
+          role = Role.find_by(name: user_hash[:role], provider: 'greenlight')
 
           return render_error status: :bad_request unless role
 
-          user = User.new(user_params.merge(provider: 'greenlight', password: generate_secure_pwd, language:, role:))
+          user_hash[:password] = generate_secure_pwd if user_hash[:external_id].blank?
+
+          user = User.new(user_hash.merge(provider: 'greenlight', role:))
           return render_error status: :bad_request unless user.save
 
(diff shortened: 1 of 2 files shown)


KH-Amir-TN:
Regarding this subject:
Do we need to send the reset email directly to users??
Instead why don't we leave that announcement responsibility to admins rather then sending an unexpected email which can cause users to report it as a spam or have a bad feedback from the authors,
Users will have their accounts migrated seamlessly,
When admins make the decision to announce the transition completely, they do that by requesting users to reset their passwords them selves by going to the v3 URL and follow the forget password steps.
That can be beneficial to admin that might want to make some tests before announcing or set the right SMTP configs, etc...

KH-Amir-TN:
# REBASED ON 1cd7852e2439e6142f2286dafa88990c43238f47

app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -38,17 +38,17 @@ def create_role
         # Does: Creates a user.
 
         def create_user
-          language = if user_params[:language].blank? || user_params[:language] == 'default'
-                       I18n.default_locale
-                     else
-                       user_params[:language]
-                     end
+          user_hash = user_params.to_h
```
KH-Amir-TN, line 41: This both simples and optimizes the logic,
Calling each time `user_params` is "somehow" an expensive operation compared to caching a hashed version of the params,
Besides we can use the hash for more efficient data retrieval and also assigning,
Using memoization here is an over kill since we may propagate it to all `*_params` methods,

app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -38,17 +38,17 @@ def create_role
...
 
-          role = Role.find_by(name: user_params[:role], provider: 'greenlight')
+          user_hash[:language] = I18n.default_locale if user_hash[:language].blank? || user_hash[:language] == 'default'
+
+          role = Role.find_by(name: user_hash[:role], provider: 'greenlight')
 
           return render_error status: :bad_request unless role
 
-          user = User.new(user_params.merge(provider: 'greenlight', password: generate_secure_pwd, language:, role:))
+          user_hash[:password] = generate_secure_pwd if user_hash[:external_id].blank?
```
KH-Amir-TN, line 49: We'd like our migrated users from to be conform with their regular created counterparts.
So, we create local migrated users a random password but not our external migrated users,

farhatahmad (approved):

PR #4010, opened 2022-10-19
Resources migration: Enabled migration tasks with batch processing.

KH-Amir-TN:
## Description

https://guides.rubyonrails.org/active_record_querying.html#retrieving-multiple-objects-in-batches

### DIFF
--- a/lib/tasks/migrations/migrations.rake
+++ b/lib/tasks/migrations/migrations.rake
@@ -2,28 +2,31 @@
 
 namespace :migrations do
   COMMON = {
-    headers: { "Content-Type" => "application/json" }
+    headers: { "Content-Type" => "application/json" },
+    batch_size: 1000,
   }.freeze
 
   desc "Migrates v2 resources to v3"
   task :roles, [] => :environment do |_task, _args|
     has_encountred_issue = 0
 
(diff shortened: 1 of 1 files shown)


KH-Amir-TN:
# DEPENDS ON #4009

lib/tasks/migrations/migrations.rake
```diff
@@ -2,28 +2,31 @@
...
-        puts cyan "  Name: #{r.name}"
-      else
-        puts red "Unable to migrate Role:"
-        puts yellow "  ID: #{r.id}"
-        puts yellow "  Name: #{r.name}"
-        has_encountred_issue = 1 # At least one of the migrations failed.
-      end
+    Role.select(:id, :name)
+        .where.not(name: Role::RESERVED_ROLE_NAMES)
+        .find_each(batch_size: COMMON[:batch_size]) do |r|
```
farhatahmad, line 15: If we're going to use the default, I would drop the explicit `batch_size: ..` to explicitly show we're using the default value

KH-Amir-TN: I've added it to enable customization to suit admins needs without having to know about the internal of `ActiveRecord`,
You wanna me to remove it?

farhatahmad: ummm no that's a fair point. it's fine as is

PR #4013, opened 2022-10-19
Resources migration: Added `migrations:rooms` task

scouillard:
## Description
Add Rooms rake task for the v2 to v3 migration

### DIFF
--- a/lib/tasks/migrations/migrations.rake
+++ b/lib/tasks/migrations/migrations.rake
@@ -78,14 +78,14 @@ namespace :migrations do
     start, stop = range(args)
     has_encountred_issue = 0
 
-    Room.select(:uid, :name, :bbb_id, :last_session, :user_id)
+    Room.select(:id, :uid, :name, :bbb_id, :last_session, :user_id)
         .find_each(start: start, finish: stop, batch_size: COMMON[:batch_size]).each do |r|
           params = { room: { friendly_id: r.uid,
                              name: r.name,
                              meeting_id: r.bbb_id,
                              last_session: r.last_session.to_datetime,
-                             owner_email: r.user.email,
-                             owner_provider: r.user.provider } }
+                             owner_email: r.owner.email,
+                             owner_provider: r.owner.provider } }
           response = Net::HTTP.post(uri('rooms'), payload(params), COMMON[:headers])
 
           case response
--- a/lib/tasks/migrations/migrations.rake
+++ b/lib/tasks/migrations/migrations.rake
@@ -74,7 +74,7 @@ namespace :migrations do
     exit has_encountred_issue
   end
 
-  task :rooms, [:start, :stop] => :environment do |_task, _args|
+  task :rooms, [:start, :stop] => :environment do |_task, args|
     start, stop = range(args)
     has_encountred_issue = 0
 
--- a/lib/tasks/migrations/migrations.rake
+++ b/lib/tasks/migrations/migrations.rake
@@ -77,7 +77,6 @@ namespace :migrations do
   task :rooms, [] => :environment do |_task, _args|
     has_encountred_issue = 0
 
-    # TODO: Optimize this by running in batches.
     Room.select(:uid, :name, :bbb_id, :last_session, :user_id)
         .find_each(batch_size: COMMON[:batch_size]).each do |r|
           params = { room: { friendly_id: r.uid, name: r.name, meeting_id: r.bbb_id, last_session: r.last_session, user_id: r.user_id } }
--- a/lib/tasks/migrations/migrations.rake
+++ b/lib/tasks/migrations/migrations.rake
@@ -79,7 +79,12 @@ namespace :migrations do
 
     Room.select(:uid, :name, :bbb_id, :last_session, :user_id)
         .find_each(batch_size: COMMON[:batch_size]).each do |r|
-          params = { room: { friendly_id: r.uid, name: r.name, meeting_id: r.bbb_id, last_session: r.last_session, user_id: r.user_id } }
+          params = { room: { friendly_id: r.uid,
+                             name: r.name,
+                             meeting_id: r.bbb_id,
+                             last_session: r.last_session,
+                             owner_email: r.user.email,
+                             owner_provider: r.user.provider } }
           response = Net::HTTP.post(uri('rooms'), payload(params), COMMON[:headers])
 
           case response
--- a/lib/tasks/migrations/migrations.rake
+++ b/lib/tasks/migrations/migrations.rake
@@ -74,11 +74,12 @@ namespace :migrations do
     exit has_encountred_issue
   end
 
-  task :rooms, [] => :environment do |_task, _args|
+  task :rooms, [:start, :stop] => :environment do |_task, _args|
+    start, stop = range(args)
     has_encountred_issue = 0
 
     Room.select(:uid, :name, :bbb_id, :last_session, :user_id)
-        .find_each(batch_size: COMMON[:batch_size]).each do |r|
+        .find_each(start: start, finish: stop, batch_size: COMMON[:batch_size]).each do |r|
           params = { room: { friendly_id: r.uid,
                              name: r.name,
                              meeting_id: r.bbb_id,
--- a/lib/tasks/migrations/migrations.rake
+++ b/lib/tasks/migrations/migrations.rake
@@ -83,9 +83,8 @@ namespace :migrations do
           params = { room: { friendly_id: r.uid,
                              name: r.name,
                              meeting_id: r.bbb_id,
-                             last_session: r.last_session.to_datetime,
-                             owner_email: r.owner.email,
-                             owner_provider: r.owner.provider } }
+                             last_session: r.last_session&.to_datetime,
+                             owner_email: r.owner.email } }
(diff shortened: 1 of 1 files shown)


lib/tasks/migrations/migrations.rake
```diff
@@ -74,6 +74,43 @@ namespace :migrations do
...
+    has_encountred_issue = 0
+
+    Room.select(:uid, :name, :bbb_id, :last_session, :user_id)
+        .find_each(batch_size: COMMON[:batch_size]).each do |r|
+          params = { room: { friendly_id: r.uid,
+                             name: r.name,
+                             meeting_id: r.bbb_id,
+                             last_session: r.last_session.to_datetime,
+                             owner_email: r.user.email,
+                             owner_provider: r.user.provider } }
```
KH-Amir-TN, line 87: Provider is expected to be 'greenlight' may fail the rooms migration,

lib/tasks/migrations/migrations.rake
```diff
@@ -74,6 +74,43 @@ namespace :migrations do
     exit has_encountred_issue
   end
 
+  task :rooms, [] => :environment do |_task, _args|
```
KH-Amir-TN, line 77: If we're enabling partitioning (range) on users and since on v2 a user will have at least one room we may need partitioning here either,

scouillard: Why would we enable partitioning? I saw that in users migration

lib/tasks/migrations/migrations.rake
scouillard: Won't that be removed for production?

lib/tasks/migrations/migrations.rake
KH-Amir-TN: I believe we enabled that task because it's expected to migrate a large number of users,
For the same reason the number of rooms is expected to be even higher,

lib/tasks/migrations/migrations.rake
KH-Amir-TN: AFAIK it's production ready,
Should have @farhatahmad input on this,

lib/tasks/migrations/migrations.rake
```diff
@@ -74,11 +74,12 @@ namespace :migrations do
     exit has_encountred_issue
   end
 
-  task :rooms, [] => :environment do |_task, _args|
+  task :rooms, [:start, :stop] => :environment do |_task, _args|
```
KH-Amir-TN, line 77: Shouldn't we rename the second block param to `args` instead?

lib/tasks/migrations/migrations.rake
```diff
@@ -74,6 +74,44 @@ namespace :migrations do
     exit has_encountred_issue
   end
 
+  task :rooms, [:start, :stop] => :environment do |_task, args|
+    start, stop = range(args)
+    has_encountred_issue = 0
+
+    Room.select(:uid, :name, :bbb_id, :last_session, :user_id)
```
KH-Amir-TN, line 81: It's always recommended to include the `:id` attribute (or the configured primary key attribute).
However, when doing batch processing it's required to provide the `:id` attribute since batching will occure by ordering, limitting and offseting records based on their ids.
This will throw an exception,

lib/tasks/migrations/migrations.rake
```diff
@@ -74,6 +74,44 @@ namespace :migrations do
...
+        .find_each(start: start, finish: stop, batch_size: COMMON[:batch_size]).each do |r|
+          params = { room: { friendly_id: r.uid,
+                             name: r.name,
+                             meeting_id: r.bbb_id,
+                             last_session: r.last_session.to_datetime,
```
KH-Amir-TN, line 86: This will cause an issue for rooms that has no `last_session`,
Besides, the conversion to `DateTime` isn't required we can marshal the `ActiveSupport::TimeWithZone` object directly and in both cases time zome will be conserved,
However, this is more then 2 times faster then if we marshal the `ActiveSupport::TimeWithZone` so good job on that!

lib/tasks/migrations/migrations.rake
```diff
@@ -74,6 +74,44 @@ namespace :migrations do
...
+                             owner_email: r.user.email,
```
KH-Amir-TN, line 87: I think you've meant `r.owner.email`

lib/tasks/migrations/migrations.rake
KH-Amir-TN: I believe we need to remove the specification of the room provider,
However, besides failing to migrate rooms this will also fail because the room owner is `r.owner` not `r.user` on v2,

lib/tasks/migrations/migrations.rake
KH-Amir-TN: I suspect that this is subject to `N+1` problem along with the users task,
We can optimize them in a new PR,

KH-Amir-TN (changes requested):

lib/tasks/migrations/migrations.rake
scouillard: I'm pretty confident I tested `r.user` but I can double check.

scouillard: Yeah ok I will change to owner. It's weird I'm sure I tested it

lib/tasks/migrations/migrations.rake
KH-Amir-TN: Umm, not sure what you mean.
As I recall both works, the `last_session` will be represented as an `ActiveSupport::TimeWithZone` on both versions.
However, converting that to a `DateTime` instance seems to be better for marshalling.
You can leave it as is and better handle the `Nil` case or sacrifice a bit of performance by omitting the conversion.
You can further test the task on your local setup,

scouillard: Added a safe operator on last_session.
I will run a test by the end of the week.

KH-Amir-TN: No worries, I'm already testing it.

lib/tasks/migrations/migrations.rake
```diff
@@ -74,6 +74,43 @@ namespace :migrations do
     exit has_encountred_issue
   end
 
+  task :rooms, [:start, :stop] => :environment do |_task, args|
+    start, stop = range(args)
+    has_encountred_issue = 0
+
+    Room.select(:id, :uid, :name, :bbb_id, :last_session, :user_id)
```
KH-Amir-TN, line 81: I forgot this,
We'd need to migrate the rooms for the migrated users.
This will migrate rooms for pending, denied and deleted users.

scouillard: Good call, I will try to fix this by the end of the day.

scouillard: Discussed this with @farhatahmad and we will migrate all the rooms.
In v3, the rooms from banned/pending users won't be created as the user will not be find - it will return a bad_request error.

scouillard: EDIT: Working on a solution to store the migrated Users in a variable. Coming soon

scouillard: Done - add migrated_users as :namespace instance variable.

scouillard: EDIT: Rake task not working like I expected, reverting last commit.

KH-Amir-TN: I don't think that we're not synchronized on this,
The migration system in place will NOT migrate any pending, denied, deleted or super admin users.
A room cannot be migrated to an inexistent user on v3 otherwise will get a bad request.
There's no way to migrate all rooms for inexistent users and a pending, denied,... users will be filtered out on users migration task.
For me all I see is that we either migrate all users including those inactive which is not aligned with what I discussed with @farhatahmad .
Or, we'll be having false negatives that will not resolve on reruns and a waste of performance due to errors caused by migrating rooms for inexistent users that will never exist and that will also ruin the migration auto repair property (rerunning the migration task will fix any failed individual migrations).

KH-Amir-TN: I don't think that we're synchronized on this,
The migration system in place will NOT migrate any pending, denied, deleted or super admin users.
A room cannot be migrated to an inexistent user on v3 otherwise will get a bad request.
There's no way to migrate all rooms for inexistent users and a pending, denied,... users will be filtered out on users migration task.
For me all I see is that we either migrate all users including those inactive which is not aligned with what I discussed with @farhatahmad .
Or, we'll be having false negatives that will not resolve on reruns and a waste of performance due to errors caused by migrating rooms for inexistent users that will never exist and that will also ruin the migration auto repair property (rerunning the migration task will fix any failed individual migrations).

scouillard:
As per @KH-Amir-TN we could merge this as is, and provide a new PR for the banned/pending user's rooms solution

lib/tasks/migrations/migrations.rake
farhatahmad:
```
role_ids = RESERVED_ROLE_NAMES.ids

Room.includes(:user)
.select(:id, :uid, :name, :bbb_id, :last_session, :user_id)
.where.not(user: { role_id: role_ids} )
.find_each(start: start, finish: stop, batch_size: COMMON[:batch_size]).each do |r|
```

PR #4014, opened 2022-10-19
V2 roles migration: Added `Migrations::create_room` API + specs

scouillard:
## Description

Add Migrations#create_room endpoint for the v2 to v3 migration.
Add specs for #create_room

### DIFF
--- a/app/controllers/api/v1/migrations/external_controller.rb
+++ b/app/controllers/api/v1/migrations/external_controller.rb
@@ -73,8 +73,10 @@ module Api
           room = Room.new(room_params.except(:owner_email, :owner_provider).merge({ user: }))
 
           # Redefines the validations method to do nothing
+          # rubocop:disable Lint/EmptyBlock
           room.define_singleton_method(:set_friendly_id) {}
           room.define_singleton_method(:set_meeting_id) {}
+          # rubocop:enable Lint/EmptyBlock
 
           return render_error status: :bad_request unless room.save
 
--- a/app/controllers/api/v1/migrations/external_controller.rb
+++ b/app/controllers/api/v1/migrations/external_controller.rb
@@ -61,6 +61,22 @@ module Api
           render_data status: :created
         end
 
+        # POST /api/v1/migrations/rooms.json
+        # Expects: { room: { :name, :friendly_id, :meeting_id, :last_session } }
+        # Returns: { data: Array[serializable objects] , errors: Array[String] }
+        # Does: Creates a room.
+        def create_room
+          user = User.find_by(email: room_params[:owner_email], provider: room_params[:owner_provider])
+
+          return render_error status: :bad_request unless user
+
+          room = Room.new(room_params.except(:owner_email, :owner_provider).merge({ user: }))
+
+          return render_error status: :bad_request unless room.save
+
+          render_data status: :created
+        end
+
         private
 
         def role_params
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -61,6 +61,28 @@ def create_user
...
+
+          return render_error status: :bad_request unless user
+
+          room = Room.new(room_params.except(:owner_email, :owner_provider).merge({ user: }))
+
+          # Redefines the validations method to do nothing
+          # rubocop:disable Lint/EmptyBlock
+          room.define_singleton_method(:set_friendly_id) {}
+          room.define_singleton_method(:set_meeting_id) {}
+          # rubocop:enable Lint/EmptyBlock
```
farhatahmad, line 79: Hmm very interesting. I never thought about this being a problem

app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -61,6 +61,28 @@ def create_user
           render_data status: :created
         end
 
+        # POST /api/v1/migrations/rooms.json
+        # Expects: { room: { :name, :friendly_id, :meeting_id, :last_session } }
+        # Returns: { data: Array[serializable objects] , errors: Array[String] }
+        # Does: Creates a room.
+        def create_room
+          user = User.find_by(email: room_params[:owner_email], provider: room_params[:owner_provider])
```
farhatahmad, line 69: I think it's best if we decrypt the params and assign it to a variable first `room_par = room_params`

farhatahmad (approved):

app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -61,6 +61,30 @@ def create_user
...
+          user = User.find_by(email: room_hash[:owner_email], provider: room_hash[:owner_provider])
+
+          return render_error status: :bad_request unless user
+
+          room = Room.new(room_hash.except(:owner_email, :owner_provider).merge({ user: }))
+
+          # Redefines the validations method to do nothing
+          # rubocop:disable Lint/EmptyBlock
+          room.define_singleton_method(:set_friendly_id) {}
+          room.define_singleton_method(:set_meeting_id) {}
```
KH-Amir-TN, line 80: Good use of ruby that saves us from impacting our codebase because of this temporary feature,
However, having the meeting ids propagating from v2 makes sense but do we need the friendly ids?

app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -61,6 +61,30 @@ def create_user
...
+          # rubocop:enable Lint/EmptyBlock
+
+          return render_error status: :bad_request unless room.save
```
KH-Amir-TN, line 83: We'd need to add in idempotency to this API in another PR,

app/controllers/api/v1/migrations/external_controller.rb
scouillard: I guess we don't really need them, but might as well?

KH-Amir-TN: That worked because we conserved somehow the same schema for both friendly ids on v2 and v3,
A cleaner approach is to pass only what v3 requires from v2 and generate what v3 can generate,

scouillard: My reasoning was that it would keep the URLs/bookmarks of the rooms

scouillard: Yeah the URLs will be different from v2. I can't think of a reason to keep the friendly_ids neither.

KH-Amir-TN: Well, good reasoning.
However, that may not help since on v2 rooms are accessed directly on the root path when on v3 they're under `/rooms`.
Besides, most of v2 deployments are hosting greenlight under a root mainly `/b` or `/gl`,

PR #4015, opened 2022-10-20
Resources migration: Minor improvement to tasks.

KH-Amir-TN:
## Description

 + Added capitalization on migrated roles names.
 + Auto mapping of v2 default roles names to v3 names.
 + Better filtering to migrated users records excluding (deleted, pending, denied, super_admin) users.


KH-Amir-TN:
# DEPENDS ON #4010

lib/tasks/migrations/migrations.rake
```diff
@@ -1,27 +1,33 @@
...
-        puts green "Succesfully migrated Role:"
-        puts cyan "  Name: #{r.name}"
-      else
-        puts red "Unable to migrate Role:"
-        puts red "  Name: #{r.name}"
-        has_encountred_issue = 1 # At least one of the migrations failed.
-      end
+    Role.select(:id, :name)
+        .where.not(name: Role::RESERVED_ROLE_NAMES)
+        .find_each(batch_size: COMMON[:batch_size]) do |r|
```
farhatahmad, line 16: I think you can do `in_batches_of` - I would also recommend lowering the size to 50 or 100

farhatahmad:
@KH-Amir-TN conflits

lib/tasks/migrations/migrations.rake
KH-Amir-TN: Please refer to #4010

KH-Amir-TN: > I think you can do `in_batches_of` - I would also recommend lowering the size to 50 or 100

`find_in_batches_of` will make the same query as `find_each` but instead of instantiating a model for `each` record in the result, it will instead (like `all`) instantiate all of the loaded raw records and yield that to given block as an array,
Since, that isn't (as I'm aware of) will optimize the queries but simply eager load the models and cause us to further iterate the given array I used `find_each` instead,

The batch size is the defaulted value from the lib writers, that indicated me that a 1000 record is acceptable to be loaded into memory besides `find_each` will not have to create a 1000 model instance and probably use only one for each iteration (only assumption),

Do you wanna default that to a 100?
To note, that will cause more queries to fire,

lib/tasks/migrations/migrations.rake
```diff
@@ -39,10 +40,14 @@ namespace :migrations do
     start, stop = range(args)
 
     has_encountred_issue = 0
+    filtered_roles_names = Role::RESERVED_ROLE_NAMES - %w[admin user]
```
KH-Amir-TN, line 43: Do you want to use `Role::RESERVED_ROLE_NAMES` as the source of truth for this,
Or we hardcode it avoiding having issues on custom deployments that might have altered the list of reserved role names,

lib/tasks/migrations/migrations.rake
```diff
@@ -111,4 +116,8 @@ namespace :migrations do
 
     [start, stop]
   end
+
+  def infer_role_name(name)
```
KH-Amir-TN, line 120: Default role names (user and admin) names on v2 will map their names to v3 configuration,
Other roles will have the same name on v3 but capitalized,

lib/tasks/migrations/migrations.rake
```diff
@@ -39,10 +40,14 @@ namespace :migrations do
     start, stop = range(args)
 
     has_encountred_issue = 0
+    filtered_roles_names = Role::RESERVED_ROLE_NAMES - %w[admin user]
 
     User.select(:id, :uid, :name, :email, :social_uid, :language, :role_id)
+        .joins(:role)
+        .where.not(roles: { name: filtered_roles_names }, deleted: true)
```
KH-Amir-TN, line 47: No deleted, pending, denied or super_admin users will be migrated,

farhatahmad: Do we need to explicitly check `deleted: true` ? Isn't that the default scope?

KH-Amir-TN: Yup, I did it to be independent of the original version of Greenlight v2 and being more specific to be compatible with the largest number possible of deployment (most of other forked version),
Isn't that desired?

PR #4016, opened 2022-10-20
V2 Resources Migration: Made `Migrations::ExternalController1` APIs idempotent.

KH-Amir-TN:
## Description

Added idempotency to `Migrations::ExternalController1` APIs .
## Testing Steps

> 1. Pull the code.


KH-Amir-TN:
# DEPENDS ON #4008

app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -38,17 +42,19 @@ def create_role
...
 
         def create_user
-          language = if user_params[:language].blank? || user_params[:language] == 'default'
-                       I18n.default_locale
-                     else
-                       user_params[:language]
-                     end
+          user_hash = user_params.to_h
+
+          return render_data status: :created if User.exists? email: user_hash[:email], provider: 'greenlight'
```
KH-Amir-TN, line 47: This is an approach where we make an prefight query to check wither or not the resource has been created,
This has the following advantages:
1. It's simpler.
2. Whenever a resource exists it will fail in one query => faster duplicated request check + no trigger of additional logic (other queries mainly).
But it suffers from the following disadvantages:
1. For the initial/original requests this caused an additional query.

app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -38,17 +42,19 @@ def create_role
...
+          user_hash[:language] = I18n.default_locale if user_hash[:language].blank? || user_hash[:language] == 'default'
+
+          role = Role.find_by(name: user_hash[:role], provider: 'greenlight')
 
           return render_error status: :bad_request unless role
 
-          user = User.new(user_params.merge(provider: 'greenlight', password: generate_secure_pwd, language:, role:))
+          user_hash[:password] = generate_secure_pwd if user_hash[:external_id].blank?
+
+          user = User.new(user_hash.merge(provider: 'greenlight', role:))
```
KH-Amir-TN, line 57: Another approach is the following:
```
          user = User.create(user_hash.merge(provider: 'greenlight', role:))

          return render_data status: :created if user.errors.size == 1 && user.errors.of_kind?(:email, :taken)

          return render_error status: :bad_request unless user.persisted?
```
This offers the following advantages:
1. For initial requests it will have less queries (one less to be specific, around 0.5 ms).

But suffers from the following disadvantages:
1. It requires more boilerplate code and complexity.
2. For all duplicated requests it will cause way more queries and logic to fire to catch the duplicated attempt.

KH-Amir-TN (commented):
Since, the first approach offered more benefits (simpler, cleaner) and since the extra query will be an issue only for 1 request (constant vs the duplicated requests which likely to happen redundantly) per resource which itself is expected (creating resources always require time) and the time itself isn't much (around 0.5 ms) I chose the first approach,
What do you think guys?

PR #4017, opened 2022-10-20
Initial setup for registration method

farhatahmad:
## Description

The initial steps to setup Registration Method. Will be followed up with multiple PRs


app/models/invitation.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class Invitation < ApplicationRecord
+  has_secure_token :token
```
KH-Amir-TN, line 4: I like this but I'm more skeptical about it,
- We're limited to tokens of `24` characters in length,
- This does guarantees no collisions to occur (along with the db unique index)  but no counter measures to follow (To be specific enough this is unlikely to happen but the probability of collisions increases with the reserved values size):
If that's normal why don't we consider that pattern for other usage of tokens that requires the user to initiate the generation process (like with reset password and verify account) and save ourselves the overhead of making those `generate_*` methods and catch the collision cases and retry the generation process?

KH-Amir-TN (approved):

app/models/invitation.rb
```diff
@@ -0,0 +1,9 @@
...
+
+  validates :email, presence: true, uniqueness: { scope: :provider }
```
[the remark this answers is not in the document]
farhatahmad: Maybe in a later PR I can come back to this

app/models/invitation.rb
farhatahmad: Let's talk about this in our meeting today

PR #4021, opened 2022-10-21
Added invite logic for local sign in 

farhatahmad:
## Testing Steps

Still a bit hard to test. You can manually generate an Invitation and try using it (will be testable next PR with emails)

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -26,12 +26,19 @@ module Api
         # TODO: amir - ensure accessibility for unauthenticated requests only.
         params[:user][:language] = I18n.default_locale if params[:user][:language].blank?
 
-        user = UserCreator.new(user_params:, provider: current_provider, role: default_role).call
+        registration_method = SettingGetter.new(setting_name: 'RegistrationMethod', provider: current_provider).call
+
+        return render_error errors: 'InviteInvalid' if registration_method == SiteSetting::REGISTRATION_METHODS[:invite] && !valid_invite_token
+
+        user = UserCreator.new(user_params: user_params.except(:invite_token), provider: current_provider, role: default_role).call
 
         # TODO: Add proper error logging for non-verified token hcaptcha
         return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
         if user.save
+          # Delete invitation (ignore whether it exists or not)
+          Invitation.delete_by(email: user_params[:email], provider: current_provider, token: user_params[:invite_token]) if registration_method == SiteSetting::REGISTRATION_METHODS[:invite]
+
           user.generate_session_token!
           session[:session_token] = user.session_token unless current_user # if this is NOT an admin creating a user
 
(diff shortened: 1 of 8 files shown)


app/javascript/components/home/HomePage.jsx
```diff
@@ -27,8 +33,16 @@ export default function HomePage() {
...
   );
 
+  useEffect(() => {
+    document.cookie = `token=${inviteToken};path=/;`;
+  }, [inviteToken]);
+
   if (isLoading) return <Spinner />;
 
+  function showSignUp() {
+    return registrationMethod !== 'invite' || !!inviteToken;
```
scouillard, line 43: Kind of hard to follow.
I would personally prefer a `===` operator instead and adding an extra clause.
SignUp will show up if the registration method is "Open" or "Approval" OR if the inviteToken (the URL param) is blank?

scouillard (approved):

app/javascript/components/home/HomePage.jsx
farhatahmad: Yeah the double bang is slightly annoying. It won't show Signup if the registration method is invite and there is no invite token. I'll add a comment here

app/controllers/api/v1/users_controller.rb
```diff
@@ -97,12 +107,18 @@ def change_password
...
       end
 
       def change_password_params
         params.require(:user).permit(:old_password, :new_password)
       end
+
+      def valid_invite_token
+        return false if user_params[:invite_token].blank?
+
+        Invitation.exists?(email: user_params[:email], provider: current_provider, token: user_params[:invite_token])
```
KH-Amir-TN, line 120: I believe this will have the validity check in the future?

app/controllers/api/v1/users_controller.rb
```diff
@@ -26,12 +26,22 @@ def create
...
+
+        user = UserCreator.new(user_params: user_params.except(:invite_token), provider: current_provider, role: default_role).call
 
         # TODO: Add proper error logging for non-verified token hcaptcha
         return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
         if user.save
+          # Delete invitation (ignore whether it exists or not)
+          if registration_method == SiteSetting::REGISTRATION_METHODS[:invite]
+            Invitation.delete_by(email: user_params[:email], provider: current_provider,
```
KH-Amir-TN, line 41: This isn't a problem here, but it's recommended to avoid the use of `delete_by` unless required since it bypasses dependencies,

app/javascript/components/home/HomePage.jsx
```diff
@@ -27,8 +33,16 @@ export default function HomePage() {
     [currentUser.signed_in],
   );
 
+  useEffect(() => {
+    document.cookie = `token=${inviteToken};path=/;`;
```
KH-Amir-TN, line 37: I suspect that this will always create the cookie,

app/javascript/components/home/HomePage.jsx
KH-Amir-TN: Agreed,
Besides, in JS there's no need to cast a value to a boolean for conditional checks,
`!!inviteToken` should be the same as `inviteToken` in a condition,

app/javascript/components/home/HomePage.jsx
KH-Amir-TN: Why we don't host this logic on signup where it will check the validity of the request,
It will further set any required cookies and query params for the signup request,

KH-Amir-TN: So the invitation URL is a `.../signup?inviteToken=xxx`

app/models/invitation.rb
```diff
@@ -5,5 +5,5 @@ class Invitation < ApplicationRecord
 
   validates :email, presence: true, uniqueness: { scope: :provider }
   validates :provider, presence: true
-  validates :token, presence: true, uniqueness: true
+  validates :token, uniqueness: true
```
KH-Amir-TN, line 8: Is this removed beacuse it's guranteed to be present?

PR #4026, opened 2022-10-24
Add Debouncing to SearchBarQuery

scouillard:
## Description

- Add debouncing to SearchBarQuery component
- Rename [input, setInput] state to [searchInput, setSearchInput] to fix inconsistencies
- Remove SearchBar component as it's not needed anymore since we removed frontend searches.

## Testing Steps

Use the SearchBox and look at the Network tab.


scouillard:
We could use useEffect instead of the `lodash.debounce` library, something like [this guy on stackoverflow](https://stackoverflow.com/a/72672732).

But it would require somewhat major refactoring of our components related to search.

I kind of like how the SearchBarQuery component is working.

Maybe you guys don't share this opinion and then we could rewrite it and make use of the useEffect/timeout solution instead.

farhatahmad (approved):

app/javascript/components/shared_components/search/SearchBar.jsx
```diff
@@ -1,20 +1,39 @@
-import React from 'react';
+import React, { useEffect, useMemo } from 'react';
 import PropTypes from 'prop-types';
+import debounce from 'lodash.debounce';
```
KH-Amir-TN, line 3: Is there a good reason of including a library for a functionality that we can implement?

app/javascript/components/shared_components/search/SearchBar.jsx
```diff
@@ -1,20 +1,39 @@
...
+import { useTranslation } from 'react-i18next';
+
+export default function SearchBar({ searchInput, setSearchInput }) {
+  const { t } = useTranslation();
+
+  const onChangeHandler = (event) => {
+    setSearchInput(event.target.value);
+  };
+
+  const debouncedOnChangeHandler = useMemo(
```
KH-Amir-TN, line 13: Isn't a `useCallback` more appropriate here if we'd like to memoize the handler (which may not be required)?
The event can be passed to the callback,

app/javascript/components/shared_components/search/SearchBar.jsx
scouillard: Would take a lot of refactoring, as commented above

app/javascript/components/shared_components/search/SearchBar.jsx
scouillard: useCallback would be used on every refetch

app/javascript/components/shared_components/search/SearchBar.jsx
KH-Amir-TN: I wasn't referring to the code that we have but rather the use of library for one functionality that we can implement, 
But after giving a second look, my bad you were using only including one module of the library and according to our dependencies we're already using lodash (which is expected for such a lib),
So, overall that makes sense we can come back to this later because the lodash debounce is somehow causing us to introduce complexity,

app/javascript/components/shared_components/search/SearchBar.jsx
KH-Amir-TN: Umm, what do you mean?

scouillard: I meant re-render, not refetch my bad.
https://dmitripavlutin.com/react-throttle-debounce/

KH-Amir-TN: Umm, I still don't see how `each time the component re-renders, a new instance of the debounced function is created by the debounce(changeHandler, 300)` occurring.

KH-Amir-TN: They're the same internally,
One will simply memoize the callback while the other memoizes its returned value,
FYI check `mountMemo, updateMemo, mountCallback and updateCallback` in https://github.dev/facebook/react/tree/main/packages/react/src

PR #4027, opened 2022-10-25
Added more invitation stuff

farhatahmad:
## Description

Added the functionality that sends an email to an invited user through the manage users tab.

## Testing Steps

1- Sign in with and admin and go to Site Settings
2- Enable `Join By Invitation` under Registration
3- Go to Manage Users and click `Invite Users`
4- Enter a list of emails (comma separated)
5- Confirm they get invitations created and emailed out (through the console)

image

### DIFF
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -94,11 +94,6 @@ input.search-bar {
   background-color: var(--brand-color-light);
 }
 
-//Heroicons
-.hi-xs {
-  width: 18px;
-}
-
 //default Heroicon width
 .hi-xs {
   width:16px;
@@ -106,6 +101,7 @@ input.search-bar {
 
 .hi-s {
   width: 24px;
+  padding-bottom: 1px;
 }
 
 .hi-m {
--- a/app/controllers/api/v1/admin/invitations_controller.rb
+++ b/app/controllers/api/v1/admin/invitations_controller.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module Admin
+      class InvitationsController < ApiController
+        before_action only: %i[create] do
+          ensure_authorized('ManageUsers')
+        end
+
+        # POST /api/v1/admin/invitations
+        def create
+          params[:invitations][:emails].split(',').each do |email|
+            invitation = Invitation.find_or_initialize_by(email:, provider: current_provider).tap do |i|
+              i.updated_at = Time.zone.now
+              i.save!
+            end
+
+            UserMailer.with(
+              email:,
+              name: current_user.name,
+              signup_url: root_url(inviteToken: invitation.token)
+            ).invitation_email.deliver_later
+          rescue StandardError => e
+            logger.error "Failed to send invitation to #{email} - #{e}"
+          end
+
+          render_data status: :ok
+        rescue StandardError => e
+          logger.error "Failed to send invitations to #{params[:invitations][:emails]} - #{e}"
+          render_error status: :bad_request
+        end
+      end
+    end
+  end
+end
--- a/app/javascript/components/admin/manage_users/forms/InviteUserForm.jsx
+++ b/app/javascript/components/admin/manage_users/forms/InviteUserForm.jsx
@@ -1,0 +1,41 @@
+import React from 'react';
+import { Button, Stack } from 'react-bootstrap';
+import { useForm } from 'react-hook-form';
+import PropTypes from 'prop-types';
+import { useTranslation } from 'react-i18next';
+import FormControl from '../../../shared_components/forms/FormControl';
+import Form from '../../../shared_components/forms/Form';
+import Spinner from '../../../shared_components/utilities/Spinner';
+import useCreateInvitation from '../../../../hooks/mutations/admin/manage_users/useCreateInvitation';
+import { createInvitaionFormFields } from '../../../../helpers/forms/CreateInvitationHelpers';
+
+export default function InviteUserForm({ handleClose }) {
+  const { t } = useTranslation();
+  const methods = useForm();
+  const createInvitation = useCreateInvitation({ onSettled: handleClose });
+  const { isSubmitting } = methods.formState;
+  const fields = createInvitaionFormFields;
+
+  fields.emails.placeHolder = t('admin.manage_users.enter_user_email');
+
+  return (
+    <Form methods={methods} onSubmit={createInvitation.mutate}>
+      <FormControl field={fields.emails} type="text" />
+
+      <Stack className="mt-1" gap={1}>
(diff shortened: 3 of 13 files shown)


app/assets/stylesheets/application.bootstrap.scss
```diff
@@ -94,18 +94,14 @@ input.search-bar {
...
-}
-
 //default Heroicon width
 .hi-xs {
   width:16px;
 }
 
 .hi-s {
   width: 24px;
+  padding-bottom: 1px;
```
farhatahmad, line 104: Vertically centers the icons with the text beside them

scouillard (approved):

app/controllers/api/v1/admin/invitations_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+              i.updated_at = Time.zone.now
+              i.save!
+            end
+
+            UserMailer.with(
+              email:,
+              name: current_user.name,
+              signup_url: root_url(inviteToken: invitation.token)
+            ).invitation_email.deliver_later
+          rescue StandardError => e
```
KH-Amir-TN, line 24: This is to ensure the completion of the task even when failing to send an email to some users?

app/controllers/api/v1/admin/invitations_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+            logger.error "Failed to send invitation to #{email} - #{e}"
+          end
+
+          render_data status: :ok
+        rescue StandardError => e
+          logger.error "Failed to send invitations to #{params[:invitations][:emails]} - #{e}"
```
KH-Amir-TN, line 30: I think this may be hard to follow,
With the current implementation this rescue block seems to be triggered beacuase of `params[:invitations][:emails].split(',').each` which will be probabily caused by an `undefined method 'xxx' for nil:NilClass (NoMethodError)`.
That can easily be validated before proceeding in the logic like we used to do and further clean this.
Besides the same exception can happen here in the resucer and throw it again causing the global rescuer to handle this which seems not to be the case since it removes the need for this,

KH-Amir-TN: Besides are we willing to treat here a generic exception (almost all running time exceptions) as a bad request and brake from all global rescuing?

app/controllers/api/v1/admin/invitations_controller.rb
farhatahmad: Correct

app/javascript/components/admin/manage_users/forms/InviteUserForm.jsx
```diff
@@ -0,0 +1,41 @@
...
+import { useTranslation } from 'react-i18next';
+import FormControl from '../../../shared_components/forms/FormControl';
+import Form from '../../../shared_components/forms/Form';
+import Spinner from '../../../shared_components/utilities/Spinner';
+import useCreateInvitation from '../../../../hooks/mutations/admin/manage_users/useCreateInvitation';
+import { createInvitaionFormFields } from '../../../../helpers/forms/CreateInvitationHelpers';
+
+export default function InviteUserForm({ handleClose }) {
+  const { t } = useTranslation();
+  const methods = useForm();
```
KH-Amir-TN, line 14: It's recommended to provide defaultValues when using `react-hook-form`,
Provided in the first rule in https://react-hook-form.com/api/useform

app/javascript/components/admin/manage_users/forms/InviteUserForm.jsx
```diff
@@ -0,0 +1,41 @@
...
+  fields.emails.placeHolder = t('admin.manage_users.enter_user_email');
+
+  return (
+    <Form methods={methods} onSubmit={createInvitation.mutate}>
+      <FormControl field={fields.emails} type="text" />
+
+      <Stack className="mt-1" gap={1}>
+        <Button variant="brand" className="w-100 mb- mt-1" type="submit" disabled={isSubmitting}>
+          { t('admin.manage_users.send_invitation') }
+          { isSubmitting && <Spinner /> }
```
KH-Amir-TN, line 28: This is fine for an `onSubmit` form, but we're avoiding the use of the `submitting` formState as a pattern and using the mutation states instead.

app/javascript/helpers/forms/CreateInvitationHelpers.jsx
```diff
@@ -0,0 +1,18 @@
...
+export const validationSchema = yup.object({
+  emails:
+    yup.string()
+      .required('Please enter your email.')
+      .email('Entered value does not match email format.'),
+});
+
+export const createInvitaionFormFields = {
+  emails: {
+    label: 'Emails',
```
KH-Amir-TN, line 12: This is irrelevant to this PR but I think we'd need to refactor the forms templates to support `i18n` and to improve it,
Here the labels will not be localized,
We can discuss during meetings,

app/javascript/helpers/forms/CreateInvitationHelpers.jsx
```diff
@@ -0,0 +1,18 @@
+import * as yup from 'yup';
+
+export const validationSchema = yup.object({
```
KH-Amir-TN, line 3: I think we're not using this?

app/views/user_mailer/invitation_email.html.erb
```diff
@@ -0,0 +1,18 @@
...
+<div style="text-align:center; font-family:'Source Sans Pro', -apple-system, BlinkMacSystemFont, 'Segoe UI', 'Helvetica Neue', Arial, sans-serif">
+  <div style="display:inline-block; background-color:#F5F7FB; border:1px solid #d3d3d3; padding: 25px 70px">
+
+    <h1 style="margin-bottom:30px">Invitation to join BigBlueButton</h1>
+
+    <p>You have been invited to your own personal space by <strong><%= @name %></strong>.</p>
+
+    <p>To signup using your email, click the button below and follow the steps.</p>
+
+    <p style="margin-bottom:35px;">The invite is only valid for 48 hours.</p>
```
KH-Amir-TN, line 12: Is the validity period not configurable?

app/controllers/api/v1/admin/invitations_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+      class InvitationsController < ApiController
+        before_action only: %i[create] do
+          ensure_authorized('ManageUsers')
+        end
+
+        # POST /api/v1/admin/invitations
+        def create
+          params[:invitations][:emails].split(',').each do |email|
+            invitation = Invitation.find_or_initialize_by(email:, provider: current_provider).tap do |i|
+              i.updated_at = Time.zone.now
```
KH-Amir-TN, line 15: This isn't an issue for this scenario, but we can find or create the record and then `touch` it?
That will probably fire an extra query on creation but it should be light weight,
The advantage is that rails will propaperly manage that and if the model has any associations it will touch it either (this isn't the case probably),

PR #4028, opened 2022-10-25
Adding status column to user model

hadicheaito1:
## Description

- Adding status enum column to users model
- renaming active, activation_digest, and activation_sent_at columns
NOTE: There are many things such as function names that still refer to the active naming that could also be changed in the future. I just tried to change what was necessary in this PR


scouillard:
Could we change all the "active" related stuff in this PR? I feel like this will never be done if not now.

farhatahmad (approved):

farhatahmad:
Hadi will follow this up with another PR - I just need this for the stuff I'm working on

Issue #3800, opened 2022-08-23
Error 500


ali-alhaidary:
**Describe the bug**
when selecting Arabic in user profile, error 500 is displayed. v2.13.0

**To Reproduce**
1. Go to user profile
2. Click on language
3. Scroll down to Arabic
4. Save the update
5. click on home

**Expected behavior**
Arabic home screen

**Actual behavior**
error 500 is displayed

**Screenshots**
![Screenshot 2022-08-22 at 22-30-53 بك بلو بتن](https://user-images.githubusercontent.com/75235623/186250305-dd81c491-134f-4144-8d8d-5c5212c07a43.png)

GUFFMHRZ:
500 Internal Server Error
If you are the administrator of this website, then please read this web application's log file and/or the web server's log file to find out what went wrong.

I think i have the same issue.

Something to note:
I have no trouble on our testservers, which have a single Greenlight instance with its own postgres container.
In our production environment, where the error occurs, 2 greenlight instances share one postgres-db(non-docker)

ali-alhaidary:
Thank you, however, the error was on the link of the v2.13 email 
announcement and not on our local server.

Ali

etopian:
issue is that the locale yml file is wrong for both arabic and iraqi arabic. it uses b as a token when it should be %b. config/locales/ar.yml

etopian:
The fix is like the following json, it requires removing the % from the end and adding it behind the b like so:

both files:
./config/locales/ar.yml
./config/locales/ar_IQ.yml

```json
  time:
    formats:
      default: "%b %d, %Y %-I:%M%P"
```

etopian:
Likely caused by some editor dealing with RTL.

etopian:
https://github.com/bigbluebutton/greenlight/pull/4055

farhatahmad:
Please note: Greenlight v3 has been released. With this new version, many of the issues and bugs that were present in v2 have been resolved.

As a result, we will no longer be providing updates or support for v2 (except for major security issues), and we will be closing any outstanding bug reports / feature requests related to v2. While we understand that some of you may still be using v2, we highly encourage you to upgrade to v3 to take advantage of the improved features and stability. If your request/bug still applies to v3, please open a new issue for it

PR #4055, opened 2022-10-31
Fix time formatting in Arabic and Iraqi Arabic locales

etopian:
## Description
The time formatting in the Arabic and Iraqi Arabic locales is broken because the time string format seems to have shifted a % from left to right.

## Testing Steps
Go to a room and switch the current user's language to Arabic or Iraqi Arabic.


etopian:
https://github.com/bigbluebutton/greenlight/issues/3800

farhatahmad (approved):

PR #4057, opened 2022-10-31
Recording consent

farhatahmad:
## Description

Added a checkbox on Room Join for rooms where `recording` is enabled. At the moment, it is not mandatory, but we should find a way to make it that way

image


scouillard:
Could we rename the 'record' setting to 'recordConsent'?

scouillard (approved):

PR #4058, opened 2022-10-31
Add RolePermission creation on Role create

scouillard:
## Description

Problem: Creating a new role would not create the subsequent RolePermissions.
Solution: Add a callback method on create in Role model.

## Testing Steps

Create a new Role.
Change/edit the Role Permissions.


scouillard:
Yeah ok, I was on the fence about it.

scouillard:
Thought about adding the Permission creation directly in spec_helper.r
Would that be ethical?
Decided to add the before block in every spec file that needs it (only x4) but could be a solution for other cases.

```
RSpec.configure do |config|
  # Those permissions are enabled by default on new Roles
  config.before do
    create(:permission, name: 'CreateRoom')
    create(:permission, name: 'CanRecord')
    create(:permission, name: 'SharedList')
    create(:permission, name: 'RoomLimit')
  end
```

scouillard:
In specs, the `before` block should be called before `let!` to avoid [unexepected behaviour ](https://dev.to/prathamesh/rspec-let-and-before-order-dependent-behavior-4gp6#:~:text=let!%20is%20called%20before%20every,run%20in%20a%20specific%20block.)

farhatahmad (approved):

farhatahmad:
We'll need to come back to the tests for this - there has to be a better way to deal with this issue

PR #4059, opened 2022-10-31
Add AdminPanel component to redirect to proper AdminPanel tab

scouillard:
## Description

Problem: AdminPanel would always redirect to AdminPanel/ManageUsers. It is a problem because you can have a certain permission without having the ManageUser permission.
Solution: Create new AdminPanel component that will redirect to the proper AdminPanel tab depending on the User's permissions.

## Testing Steps

Add/Remove certain permissions, and go to the AdminPanel.

### DIFF
--- a/app/javascript/components/admin/AdminNavSideBar.jsx
+++ b/app/javascript/components/admin/AdminNavSideBar.jsx
@@ -23,7 +23,7 @@ export default function AdminNavSideBar() {
       )}
       {(currentUser.permissions.ManageRooms === 'true') && (
         <Nav.Item>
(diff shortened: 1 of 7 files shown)


Gemfile.lock
```diff
@@ -401,3 +419,9 @@ DEPENDENCIES
   tzinfo-data
   web-console
   webdrivers
+
+RUBY VERSION
+   ruby 3.1.0p0
+
+BUNDLED WITH
+   2.3.10
```
farhatahmad, line 427: You need to remove this

farhatahmad (approved):

PR #4067, opened 2022-11-02
Admin Endpoints verification

scouillard:
## Description

Reports so far:
1. ManageRole needs to be revisited. A user with the ManageRole permission could set himself as an admin (and nuke the admins).

TODO: **_non-admin endpoints_**

  2. CreateRoom permission check needs to be revisited. For example, in Rooms and Recordings, we are doing a CreateRoom permission check to display or not the rooms and recordings tabs. But, "admin" with ManageRooms and ManageRecordings  permissions will need access to Recordings#index and Rooms#index.
 
  4. Need to find a template solution for non-permission related endpoint authorization. For example, trying to delete a SharedAccess that is not yours. Should we rework `ensure_authorized`? Create another generalised method? Add custom `before_action` in controllers?

### DIFF
--- a/spec/controllers/rooms_configurations_controller_spec.rb
+++ b/spec/controllers/rooms_configurations_controller_spec.rb
@@ -7,7 +7,6 @@ RSpec.describe Api::V1::RoomsConfigurationsController, type: :controller do
 
   before do
     request.headers['ACCEPT'] = 'application/json'
-    create_default_permissions
     sign_in_user(user)
   end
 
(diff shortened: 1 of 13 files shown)


spec/controllers/rooms_configurations_controller_spec.rb
```diff
@@ -7,6 +7,7 @@
 
   before do
     request.headers['ACCEPT'] = 'application/json'
+    create_default_permissions
```
scouillard, line 10: I think we have to add the 'CreateRoom' permission check on (most of our) endpoints and consequently add the `create_default_permissions` on (most of our) spec.

farhatahmad (approved):

PR #4070, opened 2022-11-03
Layout fix

scouillard:
## Description

Add BackgroundBuffer component that checks if user is on /rooms or /rooms/ and if so, will toggle a "white background buffer".

### DIFF
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -22,9 +22,18 @@ body {
 .header {
   height: $header-height;
   background: white;
-  vertical-align: top;
-  display: block;
 }
+
+.background-sm-buffer {
+  background-color: white;
+  height: 70px;
+}
+
+.background-lg-buffer {
+  background-color: white;
+  height: 300px;
+}
+
 
 // TODO - samuel: make responsive/add breakpoints
 .btn-xlg {
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -22,6 +22,8 @@ body {
 .header {
   height: $header-height;
   background: white;
+  vertical-align: top;
+  display: block;
 }
 
 // TODO - samuel: make responsive/add breakpoints
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -26,12 +26,12 @@ body {
 
 .background-sm-buffer {
   background-color: white;
-  height: 70px;
+  height: $background-sm-buffer-height;
 }
 
 .background-lg-buffer {
   background-color: white;
-  height: 300px;
+  height: $background-lg-buffer-height;
 }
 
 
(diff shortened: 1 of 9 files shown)


app/assets/stylesheets/helpers.scss
```diff
@@ -4,18 +4,15 @@
 }
 
 .wide-background-room {
-  background: white;
-  margin: 0 -600rem;
-  padding: 0.25rem 600rem;
-  //TODO - samuel: need to find a solution for this
-  height: 302px;
+  margin-top: -299px;
```
scouillard, line 7: This is very annoying, can't figure out what that extra pixel is but I had to let it go

app/assets/stylesheets/rooms.scss
```diff
@@ -1,6 +1,6 @@
 #room-card {
   border: 2px solid transparent !important;
-  width: 305px;
+  width: 311px;
```
scouillard, line 3: This should be tuned for different breakpoints.
Haven't started looking at responsiveness yet.

app/javascript/App.jsx
```diff
@@ -29,6 +31,7 @@ export default function App() {
   return (
     <>
       {currentUser?.signed_in && <Header /> }
+      {location?.pathname.startsWith('/rooms') && <BackgroundBuffer location={location.pathname} />}
```
scouillard, line 34: I thought about putting all the logic in <BackgroundBuffer>. And add a `return null` inside the component if the condition is not met. Wanted to stay consistent with the <Header> logic. IDK if it would impact the performance, but it might be a bit cleaner.

farhatahmad: I think its actually preferred that way - that way all of the logic is contained in one container

scouillard: All right I will put the hook back in

app/assets/stylesheets/application.bootstrap.scss
```diff
@@ -152,6 +164,7 @@ input.search-bar {
 }
 
 #nav-user-dropdown {
+  padding-right: 0;
```
farhatahmad, line 167: Can this be moved to a class?

app/javascript/components/shared_components/BackgroundBuffer.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import Container from 'react-bootstrap/Container';
+import { useLocation } from 'react-router-dom';
+
+export default function BackgroundBuffer() {
+  const location = useLocation();
+
+  if (location?.pathname.startsWith('/rooms/')) {
```
farhatahmad, line 8: What happens if you go to `localhost:3000/rooms/`? I think we need to make this a bit smarter

app/assets/stylesheets/application.bootstrap.scss
scouillard: Nope

app/javascript/components/shared_components/BackgroundBuffer.jsx
scouillard: Nice catch, I'll see what we could do.

farhatahmad (approved):

PR #4071, opened 2022-11-03
Locked down all endpoints

farhatahmad:

### DIFF
--- a/app/controllers/api/v1/rooms_configurations_controller.rb
+++ b/app/controllers/api/v1/rooms_configurations_controller.rb
@@ -3,6 +3,10 @@
 module Api
   module V1
     class RoomsConfigurationsController < ApiController
+      before_action only: %i[index] do
+        ensure_authorized(%w[CreateRoom ManageSiteSettings ManageRoles], friendly_id: params[:friendly_id])
+      end
+
       # GET /api/v1/rooms_configurations.json
       # Expects: {}
       # Returns: { data: Array[serializable objects] , errors: Array[String] }
--- a/spec/factories/role_factory.rb
+++ b/spec/factories/role_factory.rb
@@ -4,5 +4,14 @@ FactoryBot.define do
   factory :role do
     name { Faker::Job.unique.title }
     provider { 'greenlight' }
+
+    after(:create) do |role|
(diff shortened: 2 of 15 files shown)


app/controllers/api/v1/rooms_configurations_controller.rb
```diff
@@ -3,6 +3,10 @@
 module Api
   module V1
     class RoomsConfigurationsController < ApiController
+      before_action only: %i[index] do
+        ensure_authorized(%w[CreateRoom ManageSiteSettings ManageRoles], friendly_id: params[:friendly_id])
+      end
```
scouillard, line 8: A user with a SharedRoom wouldn't be able to change the settings of a SharedRoom because it's neither it's own room, nor does he has the ManageSiteSettings permission.

spec/factories/role_factory.rb
```diff
@@ -4,5 +4,14 @@
...
     provider { 'greenlight' }
+
+    after(:create) do |role|
+      perm = Permission.find_or_create_by(name: 'CreateRoom')
+      perm2 = Permission.find_or_create_by(name: 'RoomLimit')
+      perm3 = Permission.find_or_create_by(name: 'SharedList')
+      RolePermission.find_or_create_by(permission: perm, role:, value: 'true')
+      RolePermission.find_or_create_by(permission: perm2, role:, value: '100')
+      RolePermission.find_or_create_by(permission: perm3, role:, value: 'true')
+    end
```
scouillard, line 15: Nice, didn't think about `find_or_create_by`

scouillard:
It seems you like to specify the protected endpoints which need a permission check ie `before_do only: %i[index recordings_count]` even if all/most the endpoints require a that permission check.
I was debating this yesterday with the Admin endpoints.

farhatahmad:
It's to avoid calling `ensure_authenticated` more than once

app/controllers/api/v1/rooms_configurations_controller.rb
scouillard: Ah, I didn't know that.
Yeah, for now it will crash if the user is going to the Room settings

scouillard (approved):

PR #4073, opened 2022-11-04
Allow ManageRooms to view other rooms

scouillard:
## Description

Problem: User with ManageRoom permissions can't access Presentation, Shared and Settings tabs other users Rooms because it is not shared.
Solution: Add a permission checker to Room view. If user has ManageRoom permission, tabs should appear.
Issue: The `RoomsConfigurations#index` (needed in Room Settings tab) is currently restricted to users with CreateRoom ManageSiteSettings and ManageRoles. **_Is it really needed? Could we unlock this endpoint?_**

### DIFF
--- a/app/controllers/api/v1/rooms_configurations_controller.rb
+++ b/app/controllers/api/v1/rooms_configurations_controller.rb
@@ -3,6 +3,7 @@
 module Api
   module V1
     class RoomsConfigurationsController < ApiController
+      # TODO: samuel - Review this, RoomsConfig#index is needed when user with ManageRooms what to access another users room
       before_action only: %i[index] do
         ensure_authorized(%w[CreateRoom ManageSiteSettings ManageRoles], friendly_id: params[:friendly_id])
       end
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/rooms_configurations_controller.rb
```diff
@@ -3,6 +3,7 @@
 module Api
   module V1
     class RoomsConfigurationsController < ApiController
+      # TODO: samuel - Review this, RoomsConfig#index is needed when user with ManageRooms what to access another users room
       before_action only: %i[index] do
         ensure_authorized(%w[CreateRoom ManageSiteSettings ManageRoles], friendly_id: params[:friendly_id])
       end
```
scouillard, line 9: Is this really needed?

The data from /api/v1/rooms_configurations:
`{"data":{"record":"false","muteOnStart":"optional","guestPolicy":"optional","glAnyoneCanStart":"optional","glAnyoneJoinAsModerator":"optional","glRequireAuthentication":"optional","glModeratorAccessCode":"optional","glViewerAccessCode":"optional"}}`

scouillard: Quick thought before the weekend: maybe adding an OR in ensure_authorized would be an idea?
This user either has CreateRoom || ManageSiteSettings || etc.

PR #4086, opened 2022-11-08
Basic protected routes for admin

farhatahmad:
## Description

Simple permission checks for the admin side. Render a 404 if the user doesn't have the expected permission


scouillard:
I feel like all this logic belongs in `main.jsx` with the `Route` components

farhatahmad:
@scouillard The high level protections will happen there, but don't want to put too much logic in that file. Otherwise it would be super convoluted in the main file

scouillard (approved):

PR #4088, opened 2022-11-08
Fix Admin Roles table

scouillard:
## Description

Fix Role table
Remove "Role badge" in Role table for a colored shield icon.
Change the "Role badge" to have a colored dot instead of the whole badge colored background
Remove hover from Admin tables

image
image


scouillard:
Any plan for the Role color scheme? It's not on the mock ups, but I'd like to keep it in.

farhatahmad (approved):

PR #4104, opened 2022-11-09
Add disabled to Pagination button

scouillard:
## Description

A bit of an annoying fix:

1. Semantic-ui pagination buttons are a button inside of an anchor. So we need to disable the anchor with `aria-disabled=true` in the Pagination component. 
image

2. We need to disable the button itself in PaginationButton component.
3. We need to overwrite the custom button in app.scss

image

### DIFF
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -217,6 +217,12 @@ input.search-bar {
   &:focus {
     box-shadow: 0 0 0 0.25rem var(--brand-color-light);
   }
+
+  &:disabled {
+    color: gray !important;
+    pointer-events: none !important;
+    cursor: not-allowed !important;
+  }
 }
 
 .btn-brand-outline-color {
--- a/app/javascript/components/shared_components/Pagination.jsx
+++ b/app/javascript/components/shared_components/Pagination.jsx
@@ -8,6 +8,20 @@ export default function Pagination({
 }) {
   const handlePage = (e, { activePage }) => setPage(activePage);
 
+  const disabledPrevious = () => {
+    if (page === 1) {
+      return true;
+    }
+    return false;
+  };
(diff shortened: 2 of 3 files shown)


app/javascript/components/shared_components/Pagination.jsx
```diff
@@ -1,13 +1,27 @@
...
 export default function Pagination({
   page, totalPages, setPage,
 }) {
   const handlePage = (e, { activePage }) => setPage(activePage);
 
+  const disabledPrevious = () => {
+    if (page === 1) {
+      return true;
+    }
+    return false;
```
farhatahmad, line 15: you can just say `return page ===1`

app/javascript/components/shared_components/Pagination.jsx
```diff
@@ -1,13 +1,27 @@
...
+  };
+
+  const disabledNext = () => {
+    if (page === totalPages) {
+      return true;
+    }
+    return false;
```
farhatahmad, line 22: `return page === totalPages`

app/assets/stylesheets/application.bootstrap.scss
```diff
@@ -217,6 +217,12 @@ input.search-bar {
   &:focus {
     box-shadow: 0 0 0 0.25rem var(--brand-color-light);
   }
+
+  &:disabled {
+    color: gray !important;
+    pointer-events: none !important;
+    cursor: not-allowed !important;
+  }
```
farhatahmad, line 225: Theres a bootstrap `disabled` property that you can use instead https://react-bootstrap.github.io/components/buttons/#disabled-state

scouillard: It has the disabled property already but it is being overwritten by the custom button style.

farhatahmad (approved):

PR #4116, opened 2022-11-10
Fix RoomsList/RoomCard responsiveness

scouillard:
## Description

+ add `responsive` tag to our tables so that they have horizontal scroll below a certain width.

### DIFF
--- a/app/assets/stylesheets/rooms.scss
+++ b/app/assets/stylesheets/rooms.scss
@@ -1,20 +1,40 @@
 #room-card {
   border: 2px solid transparent !important;
-  width: 311px;
+  width: 305px;
+
+  @include media-breakpoint-down(xxl) {
+    width: 261px;
+  }
+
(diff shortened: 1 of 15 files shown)


app/assets/stylesheets/rooms.scss
```diff
@@ -1,20 +1,40 @@
...
+    width: 216px;
+  }
+
+  @include media-breakpoint-down(md) {
+    width: 245px;
+  }
+
+  @include media-breakpoint-down(sm) {
+    min-width: 245px;
+    width: 100%;
```
farhatahmad, line 19: Why not width 100% them and then set the width using the row and column bootstrap system?

scouillard: It feels better with a fixed width.

Having them at 100% will stretch out the card, is unpredictable.

The card should be more of a square than a rectangle.

farhatahmad (approved):

PR #4135, opened 2022-11-16
Fix Spinners

scouillard:
## Description

First PR regarding the improvement of the loading state UI in GLv3.

Remove return Spinner on loading
Add return null on loading
Style the loading button Spinners

Remaining Spinners are either

- in the loading buttons 
- the gg spinner in Room join
- going to be replaced for placeholder

Some of the `if (isLoading) return null` could be removed when we enter the refactoring phase but it doesn't break anything for now.

## Testing Steps

- Navigate through the app
- Nothing should break
- The UX should be improved

### DIFF
--- a/app/javascript/App.jsx
+++ b/app/javascript/App.jsx
@@ -25,7 +25,7 @@ export default function App() {
   document.documentElement.style.setProperty('--brand-color', brandColor);
   document.documentElement.style.setProperty('--brand-color-light', brandColorLight);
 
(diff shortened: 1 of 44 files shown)


farhatahmad:
Ah my mistake - I misread the changes

scouillard:
> Ah my mistake - I misread the changes

My PRs description could be more concise

farhatahmad (approved):

PR #4143, opened 2022-11-18
Dropdowns: Created initial custom Select and Option UI components.

KH-Amir-TN:
## Description

This PR will introduce two  new custom shared UI components `Select` and `Option`.
- Select: is a UI component that introduces an API to render a custom select like component based on the Bootstrap `Drodown` API.
- Option: is a UI component that introduces an API to render a custom option like component based on the Bootstrap `Dropdown.Item` API.
Option component is meant to be nested inside of a `Select` component and therefore have no meaning to be used on its own.
It will integrate itself to its parent Select context seamlessly.
Both of Select and Option UI components will form together the new custom Select API.

The `Select` API is still a **WIP** and cannot be considered as stable, however this PR can be considered as an initial version (70-80% done).
The end result will not differ a lot from what is introduced on this PR but rather complete it.

## Usage:
Example of using the current `Select` API:
```
<Select id="awesome-select" variant="danger" defaultOpt={{ title: 'One', value: '1' }}>
  <Option value="1">1</Option>
  <Option value="2">2</Option>
  <Option value="3">3</Option>
 </Select>

<Select onChange={(selectedVal) => updateNationality(selectedVal)} defaultOpt={{ title: 'Tunisia', value: 'TN' }}>
        {
          countries.map((country) => <Option key={country.code} value={country.code}>{country.name}</Option>)
        }
 </Select>
```
## Testing Steps

> 1. Pull the code.

image
image

### DIFF
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -225,6 +225,17 @@ input.search-bar {
   }
 }
 
+// Custom Select component.
+.select{
+  .dropdown-item.active, .dropdown-item:active  {
+    background-color: var(--brand-color) !important;
+  }
+  
+  .dropdown-item.active:hover {
+    color: $black !important;
+  }  
+}
+
 //Brand
 :root {
   --brand-color: '';
@@ -387,6 +398,10 @@ input[type='checkbox'] {
 
 .form-control:focus {
   box-shadow: none;
+}
+
+.form-control:focus, .form-control:hover {
+  border-color: var(--brand-color) !important;
 }
 
 
--- a/app/javascript/components/shared_components/utilities/Option.jsx
+++ b/app/javascript/components/shared_components/utilities/Option.jsx
@@ -1,0 +1,19 @@
+import React, { useCallback, useContext } from 'react';
+import PropTypes from 'prop-types';
+import { Dropdown } from 'react-bootstrap';
+import { ACTIONS, SelectContext } from './Select';
+
+export default function Option({ children: title, value }) {
+  const { selected, dispatch } = useContext(SelectContext);
+  const handleClick = useCallback(() => { dispatch({ type: ACTIONS.SELECT, msg: { title, value } }); }, [dispatch]);
+  const isActive = selected.title === title && selected.value === value;
+
+  return (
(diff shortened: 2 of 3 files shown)


KH-Amir-TN:
https://user-images.githubusercontent.com/29759616/202698804-928d29cd-d539-4ba8-982f-ee5ce61ab243.mp4

KH-Amir-TN:
## Note: The current version of the API is optimistic, meaning that few assumptions were made and aren't guaranteed.
1. The Select component currently does not keep hold of a collection of its registered options like the native `HTMLSelectElement.options` API.
2. Because of 1. There's no direct way to find a child `Option` data (like its title) based on a single piece of data like a defaultValue or a selectedIndex.
Therefore, the defaultValue currently is independent of the options list and isn't controlled or deterministic.
3. Because of 1. there's also no direct way of guaranteeing that the children subtree is formed by legit components (Option and Dropdown.Divider) and can have any node in place so currently an assumption is made to effectively place adequate components.
4. Because of 1. There's no guarantee over the uniqueness of options values and titles, an assumption here is made as well to count on having unique values.
5. Because of 1. The Select component cannot have a clean state initially (cannot properly select a default a option since we'd require having the exact select value and title which may not be both known).
6. Because of 1. the Select component cannot be fully controlled (there's no way of providing a value and control the component state).
7. Because of 6. some features of react-hook-form that requires controlling the component like resetting field(s) value(s) will not work with the current implementation (The formValues will be reset but the Select component will be out of sync since the form cannot control its value and therefore its state).
8. The Select and Option component were meant to be used as and in place of (to a certain extent) regular bootstrap and native select and option components.
The Select and Option components unlike the boostrap ones will not render an actual native select and option elements or any input like elements, they will only reassemble in their APIs and UX.
9. Because of 8 the Select component faces DOM challenges such us `Events`, the current event handlers `onBlur` and `onChange` aren't actually emitting an `Event`.
Therefore, expecting the bubbling of a regular or a synthetic events `onChange|onBlur` by the `Select` is error prone in this current implementation.
10. Because of 9. The current use of the `FormSelect` on a `Form` submitted `onChange|onBlur` will not work as expected.

farhatahmad (approved):

PR #4169, opened 2022-11-21
Select API: Upgraded the API to meet requirements.

KH-Amir-TN:
## Description

This PR upgrades the **Select** API to meet the requirements and notes mentioned on https://github.com/bigbluebutton/greenlight/pull/4143#issuecomment-1319897032.
This is achieved through integrating and using the `createOptionsFromChildren` function introduced on #4168 while making the adequate changes to the API.
From now on the **Select** API:
1. Will be Aware of its list of options, their associated values and titles.
2. Will be Less optimistic and less assumption based and therefore more stable and deterministic (less room for error).
3. Will be capable of having a clean state through its entire lifecycle because of its ability to select an option based on a value.
4. Can be controlled externally through its `value` prop.

## Testing Steps

> 1. Pull the code.
> 2. Install the dependencies `bundle install && npm|yarn install`.
> 3. Clean the previous assets build by running `rm app/assets/builds/*` (This won't remove .keep since it's hidden).
> 4. Clean the database and tmp files for a better isolation by running `rails tmp:clear && rails db:schema:cache:clear && rails db:drop && rails db:create && rails db:migrate:with_data`
> 5. Run the linter and specs `bundle exec rubocop --parallel && bundle exec rspec && npx eslint app/javascript/* --ext

PR #4170, opened 2022-11-21
Select API: Added custom form control component `FormSelect`.

KH-Amir-TN:
## Description

This PR adds the custom form control `FormSelect`.
**FormSelect** is a wrapping component around the **Select** UI component that will integrate and register it seamlessly with the closest react-hook form.
**FormSelect** will provide a simplified declarative experience compared to interacting directly with the **Select** API and hook-forms.

 **FormSelect** is a form control component and therefore it is a case of the regular **FormControl** that uses its same structure and design (a future refactoring can reduce duplication between both components).

## Usage:
Example of using the current `FormSelect` API:
```
 <Form methods={methods} onSubmit={updateUser.mutate}>
      <FormControl field={fields.name} type="text" />
      <FormSelect field={fields.language} variant='primary'>
        {
          Object.keys(LOCALES).map((code) => <Option key={code} value={code}>{LOCALES[code]}</Option>)
        }
      </FormSelect>
 </Form>
```

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/javascript/components/shared_components/forms/controls/FormSelect.jsx
+++ b/app/javascript/components/shared_components/forms/controls/FormSelect.jsx
@@ -1,0 +1,75 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Form as BootStrapForm } from 'react-bootstrap';
+import { useController } from 'react-hook-form';
+import Select from '../../utilities/Select';
+
+export default function FormSelect({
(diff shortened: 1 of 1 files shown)


KH-Amir-TN:
# DEPENDS ON #4169

farhatahmad (approved):

PR #4179, opened 2022-11-23
SuperAdmin permission check

hadicheaito1:
## Description

- Adding check in permission checker service
  - This checks if the `current_user` role is SuperAdmin by checking name and provider
## TODO
- rspec tests need to be done, such as testing if SuperAdmin can get the users from any provider, if SuperAdmin can get the server rooms for any provider, etc.


hadicheaito1:
waiting on #4147

farhatahmad (approved):

PR #4203, opened 2022-11-23
Upgrade to ReactRouter 6.4.3

scouillard:
## Description

Still in progress.

The `useRouteError` of Router 6.4 does not work like I expect it to.

After 3 unsuccessful queries, React Query will return an error.
image

But API errors are never caught.

However, the TypeScript errors are always caught.
image

## ErrorBoundary component
Investigating the possibility of a component `ErrorBoundary.jsx` with all our Error pages rendering logic

```
export default function ErrorBoundary() {
  const error = useRouteError();
  
  if (isRouteErrorResponse(error)) {
    if (error.status === 401) {
      return (
        <div> Unauthorized </div>
      );
    }
    if (error.status === 403) {
      return (
        <div> Forbidden </div>
      );
    }
    if (error.status === 404) {
      return (
        <NotFoundPage />
      );
    }
  }

  return <DefaultErrorPage />;
}
```


scouillard:
Maybe @KH-Amir-TN you would know why the Router catches only the Typescript error while I'm expecting it to catch our API errors. 

In practice, it works fine because the API error should eventually turn into a Typescript error.

app/javascript/routes/AuthenticatedOnly.jsx
```diff
@@ -9,13 +9,13 @@ export default function AuthenticatedOnly() {
...
     return <VerifyAccount currentUser={currentUser} />;
   }
 
   if (!currentUser.signed_in) {
     toast.error(t('toast.error.signin_required'));
-    return <Navigate to="/" />;
+    return <Navigate to="/signin" />;
   }
 
   return <Outlet />;
```
scouillard, line 21: This should be it's own PR

app/javascript/routes/AuthenticatedOnly.jsx
```diff
@@ -9,13 +9,13 @@ export default function AuthenticatedOnly() {
...
 
-  if (!currentUser.verified) {
+  if (currentUser.signed_in && !currentUser.verified) {
     return <VerifyAccount currentUser={currentUser} />;
   }
 
   if (!currentUser.signed_in) {
     toast.error(t('toast.error.signin_required'));
-    return <Navigate to="/" />;
+    return <Navigate to="/signin" />;
```
farhatahmad, line 18: Fixing this in a separate PR

app/javascript/main.jsx
```diff
@@ -28,63 +28,74 @@ import EditUser from './components/admin/manage_users/EditUser';
...
+import NotFoundPage from './components/errors/NotFoundPage';
 
-const queryClient = new QueryClient();
+const queryClientConfig = {
+  defaultOptions: {
+    queries: {
+      useErrorBoundary: true,
+    },
+  },
+};
```
farhatahmad, line 45: @scouillard I think this was the only missing piece. I tested this and it works as expected

PR #4210, opened 2022-11-24
Add redirect to /signin on AuthenticatedOnly

scouillard:
## Description

User should be redirected to `/signin` form instead of `/` homepage if he needs to be authenticated.


farhatahmad:
It redirects to `/` on purpose. What if they're using keycloak?

scouillard:
Was looking for an easy fix to test the Git setup on my new laptop. Closed.

PR #4217, opened 2022-11-25
Creating default room for user

hadicheaito1:
## Description

- Creating default room for user after user is created, after user role is updated, after user's current role CreateRoom permission changed from false to true and user has no rooms

### DIFF
--- a/app/controllers/api/v1/admin/role_permissions_controller.rb
+++ b/app/controllers/api/v1/admin/role_permissions_controller.rb
@@ -7,6 +7,7 @@ module Api
         before_action do
           ensure_authorized('ManageRoles')
         end
+        before_action :default_room, only: %i[update]
 
         def index
           roles_permissions = RolePermission.joins(:permission)
@@ -30,6 +31,19 @@ module Api
 
         def role_params
           params.require(:role).permit(:role_id, :name, :value)
+        end
+
+        def default_room
+          return unless role_params[:name] == 'CreateRoom' && role_params[:value] == true
+
+          # find users with role id and create a room for them if their room count is <= 0
+          User.where(role_id: role_params[:role_id]).find_in_batches do |group|
+            group.each do |user|
+              next if user.rooms.count.positive?
+
+              Room.create(name: "#{user.name}'s Room", user_id: user.id)
+            end
+          end
         end
       end
     end
--- a/app/controllers/api/v1/admin/role_permissions_controller.rb
+++ b/app/controllers/api/v1/admin/role_permissions_controller.rb
@@ -7,8 +7,6 @@ module Api
         before_action do
           ensure_authorized('ManageRoles')
         end
-        before_action :default_room, only: %i[update]
-
         def index
           roles_permissions = RolePermission.joins(:permission)
                                             .where(role_id: params[:role_id])
@@ -24,6 +22,7 @@ module Api
           return render_error status: :not_found unless role_permission
           return render_error status: :bad_request unless role_permission.update(value: role_params[:value].to_s)
 
+          create_default_room
           render_data status: :ok
         end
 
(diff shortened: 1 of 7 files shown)


hadicheaito1:
What do you guys think of putting the user's name in the default room?
For example is the new user is "Alphonse", the default room would be "Alphonse's Room"

hadicheaito1:
The rspec tests are failing for obvious reasons. Do you guys know of a clean way to disable specific callbacks in rspec? That would be much cleaner in this case vs manually making the tests expect the home rooms as well.

hadicheaito1:
> The solution is a little more complicated than this. Here's some questions:
> 
> * What if the user has a role with `CreateRooms` = `false`?
> * What if their role changes from one with `CreateRooms` = `false` to `true`?

Good point. The last push fixed those.

There is another case which is even trickier:
- If the `CreateRoom` permission is changed from `false` to `true`, and if there is a user with that role with no rooms, a room should also be created

I could create a check before or after the `update` in `role_permissions_controller`, then go through all the users with the role that the `CreateRooms` permissions was changed for, and then create a room for them, but this isn't efficient. Any idea on a cleaner way?

app/models/user.rb
```diff
@@ -96,6 +100,12 @@ def self.reset_token_expired?(sent_at)
     Time.current > (sent_at.in(RESET_TOKEN_VALIDITY_PERIOD))
   end
 
+  def default_room
+    return unless PermissionsChecker.new(permission_names: 'CreateRoom', user_id: id, current_user: self, friendly_id: nil, record_id: nil).call
+
+    Room.create(name: "#{name}'s Room", user_id: id) if rooms.count <= 0
```
farhatahmad, line 106: Just for performance sake, I would move the `if rooms.count <=0` to the top of the function. That way, we can do a quick check to make sure they don't already have a room

farhatahmad:
@hadicheaito1 I think that's probably your best option

farhatahmad:
You can do 1 create call along with a `find_in_batches` to improve performance

app/controllers/api/v1/admin/role_permissions_controller.rb
```diff
@@ -31,6 +32,19 @@ def update
         def role_params
           params.require(:role).permit(:role_id, :name, :value)
         end
+
+        def default_room
+          return unless role_params[:name] == 'CreateRoom' && role_params[:value] == true
+
```
hadicheaito1, line 38: I orignally had this check as well here : 
```
role_permission = RolePermission.joins(:permission).find_by(role_id: role_params[:role_id], permission: { name: role_params[:name] })

return unless role_permission.value.to_s == 'false'
```
I removed it since technically there shouldn't be a case where the value for `CreateRooms` is already `True`and the person is sending a request to change it to `True`

Lmk if you think this extra check is needed to prevent a room being made when not needed @farhatahmad

app/controllers/api/v1/admin/role_permissions_controller.rb
```diff
@@ -31,6 +32,19 @@ def update
...
+          # find users with role id and create a room for them if their room count is <= 0
+          User.where(role_id: role_params[:role_id]).find_in_batches do |group|
```
hadicheaito1, line 40: The default batch size is 1000. Should we lower it to like 500?

farhatahmad: default is fine - but you're doing the find in batches and then doing 1 create for each user in that find. You can instead do 1 create call per batch

app/controllers/api/v1/admin/role_permissions_controller.rb
```diff
@@ -7,6 +7,7 @@ class RolePermissionsController < ApiController
         before_action do
           ensure_authorized('ManageRoles')
         end
+        before_action :default_room, only: %i[update]
```
farhatahmad, line 10: I wouldn't do this in an after action, I would directly call it in the controller. What if the update fails?

app/controllers/api/v1/admin/role_permissions_controller.rb
farhatahmad: We may also want to add a check to make sure we're only creating rooms for users with no room

hadicheaito1: Are you sure about the 1 create call per batch? I know you can `update` a batch in one call, but to create `Rooms` for `Users` with one call per batch i couldn't find anything. I can't find a way to pass a list of users into one create call. We also need to be able to access the user's `name` and `id` for each `room` being created.

There was already a check to make sure the user had no room in the  loop. I made it better and changed it so that the sql request only gets the users with no rooms and the matching role

PR #4219, opened 2022-11-29
User signup existing email

alihadi-mazeh:
Specified the error message when creating a user with an existing email

## Description

Accounted for the specific error in the users controller. Created a specific toast error in en.json. Then in useCreateUser.jsx added the corresponding toast to display the specified error

image

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -55,7 +55,7 @@ module Api
           render_data data: current_user, serializer: CurrentUserSerializer, status: :created
         else
           # TODO: amir - Improve logging.
-          render_error errors: user.errors.to_a, status: :bad_request, errors: 'UserAlreadyExists'
+          render_error errors: 'UserAlreadyExists', status: :bad_request
         end
       end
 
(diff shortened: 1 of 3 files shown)


hadicheaito1:
You created a branch off of the last branch you had.

I'd close this PR, checkout v3, fetch and pull, create a new branch from the v3 branch we made together, and then re-do the changes and push

app/controllers/api/v1/users_controller.rb
```diff
@@ -55,7 +55,7 @@ def create
           render_data data: current_user, serializer: CurrentUserSerializer, status: :created
         else
           # TODO: amir - Improve logging.
-          render_error errors: user.errors.to_a, status: :bad_request
+          render_error errors: 'UserAlreadyExists', status: :bad_request
```
farhatahmad, line 58: So what this will do is always throw this error regardless of why it fails. We need to check `user.errors` and see if a duplicate email is the reason they fail. If not, we can just return `render_error status: :bad_request`

alihadi-mazeh: I addressed the change in my latest commit :+1:

farhatahmad (approved):

PR #4220, opened 2022-11-30
Fixing server delete room toast

hadicheaito1:
## Description

- Fixing broken toast for server delete room


hadicheaito1:
ESLint error from Select component that was recently made, not sure if it's still being worked on

farhatahmad (approved):

PR #4222, opened 2022-11-30
Fixed the localization workflow 

farhatahmad:
## Description

Here is the new workflow:

- If the user's browser is set to `ab-CD`, the user will be shown the strings from this language when first accessing the app
  - if we haven't translated `ab-CD`, then it will fallback to `ab`
  - if we haven't translated `ab`, then it will fallback to `en`

- On signup, the user's language will be set to `ab-CD` and the same fallback mechanism will apply


app/javascript/components/users/user/forms/UpdateUserForm.jsx
```diff
@@ -34,12 +34,16 @@ export default function UpdateUserForm({ user }) {
...
+      language: i18n.resolvedLanguage, // Whatever language is currently rendering (needed to handle unsupported languages)
       role_id: user?.role?.id,
     },
     resolver: yupResolver(validationSchema),
   });
 
+  useEffect(() => {
+    methods.setValue('language', i18n.resolvedLanguage);
+  }, [i18n.resolvedLanguage]);
+
```
farhatahmad, line 46: `i18n.resolvedLanguage` initially renders as `en`, before switching to `de-DE` (or whatever language you use)

hadicheaito1 (approved):

PR #4252, opened 2022-12-01
Fix SiteSettings and RoomConfigs toggles

scouillard:
## Description

Put isLoading back on SiteSettings and RoomConfigs pages as the toggles button would not load properly.
Didn't test them all, need those for the conf tho.


farhatahmad:
Got too many commits in here @scouillard

PR #4261, opened 2022-12-13
Fixed issues with locales

farhatahmad:


hadicheaito1 (approved):
I'm a bit out of the loop with the locale stuff, but where and when would the `show` method get called for this with `params[:name]`

farhatahmad:
Theres a library called `react-i18n` on the frontend that will automatically make this call to the backend on every full page refresh and then cache the value automatically

PR #4268, opened 2022-12-15
Add extra measures to improve join robustness

farhatahmad:
## Description

I've made a few changes to the join logic that should, in theory, make some improvements to the robustness. See code for explanations

Thoughts?

### DIFF
--- a/app/javascript/components/rooms/room/join/RoomJoin.jsx
+++ b/app/javascript/components/rooms/room/join/RoomJoin.jsx
@@ -88,7 +88,7 @@ export default function RoomJoin() {
     if (hasStarted) {
       toast.success(t('toast.success.room.meeting_started'));
       console.info(`Attempting to join the room(friendly_id): ${friendlyId} meeting in 7s.`);
-      setTimeout(methods.handleSubmit(handleJoin), 7000); // TODO: Improve this race condition handling by the backend.
+      setTimeout(methods.handleSubmit(handleJoin), Math.floor(Math.random() * 4000) + 1000); // Offset join times to ease load on BigBlueButton server
       reset();// Resetting the Join component.
     }
   }, [hasStarted]);
--- a/app/services/meeting_starter.rb
+++ b/app/services/meeting_starter.rb
@@ -21,7 +21,13 @@ class MeetingStarter
 
       @room.update!(online: true, last_session: DateTime.strptime(meeting[:createTime].to_s, '%Q'))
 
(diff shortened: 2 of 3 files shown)


app/javascript/components/rooms/room/join/RoomJoin.jsx
```diff
@@ -88,7 +88,7 @@ export default function RoomJoin() {
     if (hasStarted) {
       toast.success(t('toast.success.room.meeting_started'));
       console.info(`Attempting to join the room(friendly_id): ${friendlyId} meeting in 7s.`);
-      setTimeout(methods.handleSubmit(handleJoin), 7000); // TODO: Improve this race condition handling by the backend.
+      setTimeout(methods.handleSubmit(handleJoin), Math.floor(Math.random() * 4000) + 1000); // Offset join times to ease load on BigBlueButton server
```
farhatahmad, line 91: To prevent all users from hitting the BigBlueButton server at the same time, we can offset the time that they join by making them way a random amount of time (between 1 and 5 seconds).

This is already after atleast 7 seconds of the meeting being started (see below)

app/services/meeting_starter.rb
```diff
@@ -21,7 +21,13 @@ def call
...
       @room.update!(online: true, last_session: DateTime.strptime(meeting[:createTime].to_s, '%Q'))
 
-      ActionCable.server.broadcast "#{@room.friendly_id}_rooms_channel", 'started'
+      # new thread to avoid blocking the other code from running
+      Thread.new do
+        sleep 7 # Sleep 7 seconds to allow enough time for room to start
+        ActionCable.server.broadcast "#{@room.friendly_id}_rooms_channel", 'started'
+        sleep 7 # Sleep and rebroadcast again incase anyone missed the first broadcast
+        ActionCable.server.broadcast "#{@room.friendly_id}_rooms_channel", 'started'
+      end
```
farhatahmad, line 30: Wait 7 seconds and then inform the waiting cables/sockets that the meeting has started, and the wait another 7 seconds and broadcast again (incase we missed someone)

This is running in a new thread to prevent the whole app from waiting (due to the sleep)

alihadi-mazeh (approved):

app/javascript/components/rooms/room/join/RoomJoin.jsx
scouillard: This was mentioned at the Berlin convention

PR #4282, opened 2022-12-16
Adding current provider check in permission checker

hadicheaito1:
## Description

- Adding a check to make sure that the user is accessing their own provider info

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -76,7 +76,7 @@ RSpec/AnyInstance:
   Enabled: false
 
 Metrics/CyclomaticComplexity:
-  Max: 14
+  Max: 15
 
 Metrics/PerceivedComplexity:
   Max: 13
--- a/app/services/permissions_checker.rb
+++ b/app/services/permissions_checker.rb
@@ -1,15 +1,19 @@
 # frozen_string_literal: true
 
 class PermissionsChecker
-  def initialize(current_user:, permission_names:, user_id:, friendly_id:, record_id:)
+  def initialize(current_user:, permission_names:, user_id:, friendly_id:, record_id:, current_provider:)
     @current_user = current_user
     @permission_names = permission_names
     @user_id = user_id
     @friendly_id = friendly_id
     @record_id = record_id
+    @current_provider = current_provider
   end
 
   def call
+    # check to see if current user is trying to access own provider info
+    return false if @current_user.provider != @current_provider
+
     return true if @current_user.role == Role.find_by(name: 'SuperAdmin', provider: 'bn')
 
     return true if RolePermission.joins(:permission).exists?(
--- a/spec/services/permissions_checker_spec.rb
+++ b/spec/services/permissions_checker_spec.rb
@@ -15,7 +15,8 @@ describe PermissionsChecker, type: :service do
         permission_names: permission.name,
         user_id: '',
         friendly_id: '',
-        record_id: ''
+        record_id: '',
+        current_provider: user.provider
       ).call).to be(true)
     end
 
@@ -26,7 +27,8 @@ describe PermissionsChecker, type: :service do
         permission_names: permission.name,
         user_id: '',
         friendly_id: '',
-        record_id: ''
+        record_id: '',
+        current_provider: user.provider
       ).call).to be(false)
     end
 
@@ -37,7 +39,8 @@ describe PermissionsChecker, type: :service do
         permission_names: 'ManageUsers',
         user_id: user.id,
         friendly_id: '',
-        record_id: ''
+        record_id: '',
+        current_provider: user.provider
       ).call).to be(true)
     end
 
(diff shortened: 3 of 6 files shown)


.rubocop.yml
```diff
@@ -76,8 +76,8 @@ RSpec/AnyInstance:
   Enabled: false
 
 Metrics/CyclomaticComplexity:
-  Max: 14
+  Max: 17
```
scouillard, line 79: What is causing this? The new logic in PermissionChecker I assume?

app/services/permissions_checker.rb
```diff
@@ -1,17 +1,25 @@
...
 
   def call
+    # check to see if current_user has SuperAdmin role
     return true if @current_user.role == Role.find_by(name: 'SuperAdmin', provider: 'bn')
 
+    # check to see if current user is trying to access own provider info
+    return false if @current_user.provider != @current_provider
+
+    # check to see if current_user is trying to access a user from another provider
+    return false if @user_id.present? && (@current_user.provider != User.find(@user_id.to_s).provider)
```
scouillard, line 21: Would putting those two checks in a separated method solve the Rubocop complexity error?

scouillard: Maybe extract line 15 as well

.rubocop.yml
hadicheaito1: Yeah, `CyclomaticComplexity` is the number of possible independent paths that the method could take

app/services/permissions_checker.rb
hadicheaito1: We could, but it would just lower the number of possible outcomes in the method. I feel we should either leave them all as is, or take lines 14-27 and put them into a function

scouillard: Everytime I had those Rubocop error, I found a way to rewrite the logic in a simpler/more satisfying way.
Could be a bit harder to do here as this is one of the most complex service we have.
I would definetly try to avoid lowering the Rubocop constraint.

app/services/permissions_checker.rb
```diff
@@ -1,17 +1,25 @@
...
     @record_id = record_id
+    @current_provider = current_provider
   end
 
   def call
+    # check to see if current_user has SuperAdmin role
     return true if @current_user.role == Role.find_by(name: 'SuperAdmin', provider: 'bn')
 
+    # check to see if current user is trying to access own provider info
+    return false if @current_user.provider != @current_provider
```
KH-Amir-TN, line 18: Good catch!
That saves us a security ticket!!

app/services/permissions_checker.rb
KH-Amir-TN: Don't we want to further check the rest of resources like rooms, recordings?
Like with now we're checking if a user is trying to manage a user from another organazation we want to do that also for the rest of resources.
Maybe extract that into a private method like a `check_provider` or something more relevant.
That should help reduce the complexity and improve maintainability and readibility.
That may be a bit challenging to make the method communicate with the rest of logic (if returns false that won't fail the check as is now).
Remember to always start first with `return false if @current_user.provider != @current_provider`.

app/services/permissions_checker.rb
KH-Amir-TN: That shouldn't be a problem,
However, it may be better tp have this check as the first one even before the super admin check.

spec/services/permissions_checker_spec.rb
```diff
@@ -75,7 +80,8 @@
           permission_names: [],
           user_id: super_admin_user.id,
           friendly_id: '',
-          record_id: ''
+          record_id: '',
+          current_provider: super_admin_user.provider
```
KH-Amir-TN, line 84: Maybe we can add another check to ensure that we will catch users trying to access resources on a different provider,

.rubocop.yml
KH-Amir-TN: Maybe that is too high,
That may have a downside on our code quality and make refactoring more challenging,
No strong opinion tho,

KH-Amir-TN (commented):
# :fire: :fire: :fire:

scouillard (approved):

app/services/permissions_checker.rb
```diff
@@ -74,4 +79,20 @@ def authorize_room_limit
...
+    return false if @user_id.present? && (@current_user.provider != User.find(@user_id.to_s).provider)
+
+    # check to see if current_user is trying to access a room from another provider
+    return false if @friendly_id.present? && (@current_user.provider != Room.find_by(friendly_id: @friendly_id.to_s).user.provider)
+
+    # check to see if current_user is trying to access a recording from another provider
+    return false if @record_id.present? && (@current_user.provider != Recording.find_by(record_id: @record_id.to_s).user.provider)
+
+    true
+  end
```
scouillard, line 97: I think it would be a bit easier to read if it was positive logic
`return true if ==`

farhatahmad (approved):

PR #4290, opened 2022-12-19
Fix Access Code edgecases

scouillard:
## Description

This PR aims to review/fix the Access Code process.

**Moderator Code is enabled**
When a Moderator Code is enable, the input bar will be visible to anyone on the RoomJoin page. 
Right now, if you enter an input,  if the input correspond to the Moderator Access Code, you will join in as a Moderator.
If the input does not correspond, you will join in as a Viewer.
Following this PR, if a user enters an input that does not correspond to the Moderator Access Code, it will be denied the access to the room. 
So, the input field needs to be let blank if you want to join as a Viewer.

**Anyone Join As Moderator**
When AnyoneJoinAsModerator is enabled, the Room owner can still enable the Viewer and Moderator Access code.
Before this PR, enabling AnyoneJoinAsModerator would bypass the Access Code checkup.
Following this PR, entering an Access Code that corresponds to either the Moderator or the Viewer Access Code will let the user in as a Moderator.

## Testing Steps

- Toggle the RoomSettings, especially AnyoneJoinAsModerator
- Toggle the Mod/Viewer Access Code
- Open an incognito browser and try to join in the Room with different scenarios

### DIFF
--- a/app/controllers/api/v1/meetings_controller.rb
+++ b/app/controllers/api/v1/meetings_controller.rb
@@ -34,12 +34,14 @@ module Api
           provider: current_provider,
           current_user:,
           show_codes: true,
-          settings: %w[glRequireAuthentication glViewerAccessCode glModeratorAccessCode glAnyoneCanStart]
+          settings: %w[glRequireAuthentication glViewerAccessCode glModeratorAccessCode glAnyoneCanStart glAnyoneJoinAsModerator]
         ).call
 
         return render_error status: :unauthorized if !current_user && settings['glRequireAuthentication'] == 'true'
 
-        bbb_role = infer_bbb_role(mod_code: settings['glModeratorAccessCode'], viewer_code: settings['glViewerAccessCode'])
+        bbb_role = infer_bbb_role(mod_code: settings['glModeratorAccessCode'],
+                                  viewer_code: settings['glViewerAccessCode'],
+                                  anyone_join_as_mod: settings['glAnyoneJoinAsModerator'])
 
         return render_error status: :forbidden if bbb_role.nil?
 
@@ -95,16 +97,16 @@ module Api
       # 4. a room that has the AnyoneJoinAsModerator setting toggled on and does not require an access code
       # 5. a room that has the AnyoneJoinAsModerator setting toggled on and requires a moderator or a viewer access code
       #    and the access code input correspond to either the moderator or the viewer access code
-      def authorized_as_moderator?(mod_code:, viewer_code:)
+      def authorized_as_moderator?(mod_code:, viewer_code:, anyone_join_as_mod:)
         @room.user_id == current_user&.id ||
           current_user&.shared_rooms&.include?(@room) ||
           (params[:access_code].present? && mod_code == params[:access_code]) ||
-          (@room.anyone_joins_as_moderator? && mod_code.blank? && viewer_code.blank?) ||
-          (@room.anyone_joins_as_moderator? && (mod_code == params[:access_code] || viewer_code == params[:access_code]))
+          (anyone_join_as_mod && mod_code.blank? && viewer_code.blank?) ||
+          (anyone_join_as_mod && (mod_code == params[:access_code] || viewer_code == params[:access_code]))
       end
 
-      def infer_bbb_role(mod_code:, viewer_code:)
-        if authorized_as_moderator?(mod_code:, viewer_code:)
+      def infer_bbb_role(mod_code:, viewer_code:, anyone_join_as_mod:)
+        if authorized_as_moderator?(mod_code:, viewer_code:, anyone_join_as_mod:)
           'Moderator'
         elsif authorized_as_viewer?(viewer_code:)
           'Viewer'
--- a/app/controllers/api/v1/meetings_controller.rb
+++ b/app/controllers/api/v1/meetings_controller.rb
@@ -81,25 +81,32 @@ module Api
         @room = Room.find_by!(friendly_id: params[:friendly_id])
       end
 
-      def authorized_as_viewer?(access_code:)
-        return true if access_code.blank?
+      def authorized_as_viewer?(viewer_code:)
+        return true if viewer_code.blank? && params[:access_code].blank?
 
-        access_code == params[:access_code]
+        viewer_code == params[:access_code]
       end
 
-      def authorized_as_moderator?(access_code:)
-        (params[:access_code].present? && access_code == params[:access_code]) ||
-          @room.anyone_joins_as_moderator? ||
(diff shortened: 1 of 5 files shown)


scouillard:
@KH-Amir-TN I removed the Room method `anyone_join_as_moderator?` as it was only used in the MeetingController#status.
We can leverage the RoomSettingGetter service that is already used there instead to know if the setting is enabled.
Do you think we might need the `anyone_join_as_moderator?` method somewhere else?

app/controllers/api/v1/meetings_controller.rb
```diff
@@ -81,25 +83,32 @@ def find_room
...
-        (params[:access_code].present? && access_code == params[:access_code]) ||
-          @room.anyone_joins_as_moderator? ||
+      # Five scenarios where a user is authorized to join a BBB meeting as a moderator
+      # The user joins
+      # 1. its own room
+      # 2. a shared room
+      # 3. a room that requires a moderator access code and the access code input is correct
+      # 4. a room that has the AnyoneJoinAsModerator setting toggled on and does not require an access code
+      # 5. a room that has the AnyoneJoinAsModerator setting toggled on and requires a moderator or a viewer access code
+      #    and the access code input correspond to either the moderator or the viewer access code
```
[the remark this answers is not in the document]
KH-Amir-TN: Umm, I thought about the scenarios from a different perspective:
      # The user joins
      # 1. its own room
      # 2. a shared room
      # 3. a room with AnyoneJoinAsModerator setting off but has a moderator access code that the user input matched. 
      # 4. a room that has the AnyoneJoinAsModerator setting toggled on and does not require an access code (Here we don't care about the moderator access code).
      # 5. a room that has the AnyoneJoinAsModerator setting toggled on and requires an access code
      #    which the user input correspond to.
      The moderator access code is optional, it has a meaning only if the AnyoneJoinAsModerator setting if off it enables users to join in as moderators.
      When the option is toggled on the optional code has no effects.
      The access code (viewer access code) is required and its the true access code that if enabled will define the eligibility of joining or not a meeting.
      We may update the UI so that the moderator access code will not be displayed when the setting is toggled on.
      We have some other mutual exclusive settings that we haven't touched yet.

app/controllers/api/v1/meetings_controller.rb
```diff
@@ -81,25 +83,32 @@ def find_room
...
+      def authorized_as_moderator?(mod_code:, viewer_code:, anyone_join_as_mod:)
+        @room.user_id == current_user&.id ||
           current_user&.shared_rooms&.include?(@room) ||
-          @room.user_id == current_user&.id
+          (params[:access_code].present? && mod_code == params[:access_code]) ||
```
KH-Amir-TN, line 103: This is a none simplified conditional branches of the scenarios I mentionned above that may help us when discussing this:
```
        @room.user_id == current_user&.id ||
          current_user&.shared_rooms&.include?(@room) ||
          (!anyone_join_as_mod && mod_code.present? && mod_code == params[:access_code]) ||
          (anyone_join_as_mod && viewer_code.blank?) ||
          (anyone_join_as_mod && viewer_code.present? && viewer_code == params[:access_code])
```

app/controllers/api/v1/meetings_controller.rb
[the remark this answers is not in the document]
KH-Amir-TN: We can discuss different persepctives and see the pros and cons of each from the users expectations,

KH-Amir-TN:
> @KH-Amir-TN I removed the Room method `anyone_join_as_moderator?` as it was only used in the MeetingController#status. We can leverage the RoomSettingGetter service that is already used there instead to know if the setting is enabled. Do you think we might need the `anyone_join_as_moderator?` method somewhere else?

Yeah, it makes sense.
There's similar model methods that may remove in favor of services,
Good catch!

app/controllers/api/v1/meetings_controller.rb
```diff
@@ -81,25 +83,32 @@ def find_room
...
+          (anyone_join_as_mod && mod_code.blank? && viewer_code.blank?) ||
+          (anyone_join_as_mod && (mod_code == params[:access_code] || viewer_code == params[:access_code]))
```
KH-Amir-TN, line 105: After making e2e tests,
The following scenario failed:
```
[When anyone join as mod is ON]
Access code only:
	+ If entered join as mod [PASSED]
	+ If not entered or entered incorrectly forbid [FAILED, was able to bypass the access code check by an empty code]
```
I was able to bypass the access code check when giving an empty value when the access code is present but the moderator code is blank.
That is due to this line checking values directly without asserting the presence of codes.
This creates a security vulnerability,

KH-Amir-TN: In addition, the `anyone joins as moderator` setting will always be treated as ON no matter what,
After isolating the cause, it's because of using `anyone_join_as_mod` as if it was a Boolean value,
The rooms meeting options value attribute contains strings not Booleans,
Boolean settings will have a stringified Boolean value like `'true'` but the type will be a string,
A string in ruby is truthy meaning that no matter what value is held it will be casted to a Boolean `true`,

KH-Amir-TN (changes requested):

app/controllers/api/v1/meetings_controller.rb
KH-Amir-TN: Actually after thinking about this, I started to see why you used the moderator code @scouillard,

app/controllers/api/v1/meetings_controller.rb
```diff
@@ -81,25 +83,32 @@ def find_room
         @room = Room.find_by!(friendly_id: params[:friendly_id])
       end
 
-      def authorized_as_viewer?(access_code:)
-        return true if access_code.blank?
+      def authorized_as_viewer?(viewer_code:)
+        return true if viewer_code.blank? && params[:access_code].blank?
```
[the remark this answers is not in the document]
scouillard: We could also let the user type in anything in the Access Code Input, and he will be joining as a regular Viewer if the Access Code does not match the Moderator Access Code. I changed it because I felt it was bad UI but maybe it saves us a bunch of headaches.

app/controllers/api/v1/meetings_controller.rb
scouillard: Ah, right.
Thank you for your time @KH-Amir-TN

scouillard: > 

Not sure of this, isn't it expected?
If Moderator Access Code is generated, and you don't provide the access code when joining, it is expected that you will join as a regular Viewer.

farhatahmad (approved):

PR #4299, opened 2022-12-20
Forms refactoring: Refactored `SignupForm`

KH-Amir-TN:
## Description

This PR introduces a refactoring to the forms framework we're using to provide simplicity and to enable forms logic to be encapsulated in one hook that is easily maintainable and extendable ( can use other hooks, so localization for example becomes possible).
+ Refactored the `useCreateUser` mutation hook and localized the sign up form.

> 1. Pull the code.


app/controllers/api/v1/users_controller.rb
```diff
@@ -33,7 +33,7 @@ def create
         user = UserCreator.new(user_params: user_params.except(:invite_token), provider: current_provider, role: default_role).call
 
         # TODO: Add proper error logging for non-verified token hcaptcha
-        return render_error errors: user.errors.to_a if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
+        return render_error errors: 'HCaptchaInvalid' if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
```
KH-Amir-TN, line 36: I noticed that when refactoring the form and testing HCaptcha,

app/javascript/components/users/authentication/forms/SignupForm.jsx
```diff
@@ -1,62 +1,71 @@
...
 
 export default function SignupForm() {
+  const captchaRef = useRef(null);
   const { t } = useTranslation();
-  const methods = useForm(signupFormConfig);
-  const [token, setToken] = useState('');
-  const { onSubmit: createUser } = useCreateUser(token);
-  const { isSubmitting } = methods.formState;
-  const fields = signupFormFields;
+  const { fields, methods } = useSignUpForm();
```
KH-Amir-TN, line 16: Instead of having to deal with multiple objects and to do multiple steps to set a form,
The idea is now similar to how react hook form was designed.
Any form low logic on the App is nothing but a hook,
The hook will configure the react hook form as intended,
Plus extend that with the fields API that the Form controls API will use seamlessly,

app/javascript/components/users/authentication/forms/SignupForm.jsx
```diff
@@ -1,62 +1,71 @@
...
 import Spinner from '../../../shared_components/utilities/Spinner';
 import useCreateUser from '../../../../hooks/mutations/users/useCreateUser';
 import useEnv from '../../../../hooks/queries/env/useEnv';
+import useSignUpForm from '../../../../hooks/forms/authentication/useSignUpForm';
 
 export default function SignupForm() {
+  const captchaRef = useRef(null);
   const { t } = useTranslation();
-  const methods = useForm(signupFormConfig);
-  const [token, setToken] = useState('');
```
KH-Amir-TN, line 17: This state was declared but not used,

app/javascript/hooks/forms/authentication/useSignUpForm.js
```diff
@@ -0,0 +1,86 @@
...
+        validations: {
+          deps: ['password'],
+        },
+      },
+    },
+  };
+
+  const validationSchema = yup.object({
+    name: yup.string().required('forms.user.signup.validations.full_name.required')
+      .min(2, 'forms.user.signup.validations.full_name.min')
```
KH-Amir-TN, line 52: We can localize validations from here now,
I just left as is for compatibility,
You can carry any WIP on localizing validations and we can make the change later on,

farhatahmad (approved):

PR #4303, opened 2022-12-20
Converted language list to dynamic

farhatahmad:
## Description

The list of languages is now being pulled dynamically based on the languages available in the drop down. 

The result is cached for 24 hours as the result will never change (until upgrade)


scouillard (dismissed):

hadicheaito1 (approved):

KH-Amir-TN (approved):
# 🔥 🔥 🔥 
# Love this!

PR #4311, opened 2022-12-21
Users Controller: Minor fixes and improvements to APIs and specs.

KH-Amir-TN:
## Description

Fixes:
- [x] Signup failing for unauthenticated requests when HCaptcha is disabled (bug introduced by #4299.
- [x] Creating users from admin panel fails when HCaptcha is enabled.
- [x] Deleting a user from admin panel fails.
- [x] Created users from admin panel were having the creator browser language set and not the admin language.
- [x] Duplicated email error will not be detected for servers configured with a locale other than `:en` (English).
 
Improvements:
- [x] Reduced literal usage.

## Testing Steps

> 1. Pull the code.

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -7,7 +7,7 @@ module Api
 
       skip_before_action :ensure_authenticated, only: %i[create]
 
-      before_action only: %i[show update destroy purge_avatar] do
+      before_action except: %i[create change_password] do
         ensure_authorized('ManageUsers', user_id: params[:id])
       end
 
@@ -17,23 +17,28 @@ module Api
         render_data data: user, status: :ok
       end
 
+      # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
       # POST /api/v1/users.json
       # Expects: { user: { :name, :email, :password} }
       # Returns: { data: Array[serializable objects] , errors: Array[String] }
       # Does: Creates and saves a new user record in the database with the provided parameters.
 
       def create
-        # TODO: amir - ensure accessibility for unauthenticated requests only.
-        params[:user][:language] = I18n.default_locale if params[:user][:language].blank?
+        # Users created by a user will have the creator language by default with a fallback to the server configured default_locale.
+        user_params[:language] = current_user&.language || I18n.default_locale if user_params[:language].blank?
 
         registration_method = SettingGetter.new(setting_name: 'RegistrationMethod', provider: current_provider).call
 
-        return render_error errors: 'InviteInvalid' if registration_method == SiteSetting::REGISTRATION_METHODS[:invite] && !valid_invite_token
+        if registration_method == SiteSetting::REGISTRATION_METHODS[:invite] && !valid_invite_token
+          return render_error errors: Rails.configuration.custom_error_msgs[:invite_token_invalid]
+        end
 
         user = UserCreator.new(user_params: user_params.except(:invite_token), provider: current_provider, role: default_role).call
 
         # TODO: Add proper error logging for non-verified token hcaptcha
-        return render_error errors: 'HCaptchaInvalid' if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
+        if !current_user && hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
+          return render_error errors: Rails.configuration.custom_error_msgs[:hcaptcha_invalid]
+        end
 
         # Set to pending if registration method is approval
         user.pending! if !current_user && registration_method == SiteSetting::REGISTRATION_METHODS[:approval]
@@ -55,14 +60,14 @@ module Api
           create_default_room(user)
 
           render_data data: current_user, serializer: CurrentUserSerializer, status: :created
-        elsif user.errors.to_a == ['Email has already been taken']
-          render_error errors: 'EmailAlreadyExists', status: :bad_request
+        elsif user.errors.size == 1 && user.errors.of_kind?(:email, :taken)
+          render_error errors: Rails.configuration.custom_error_msgs[:email_exists], status: :bad_request
         else
-          render_error errors: user.errors.to_a, status: :bad_request
+          render_error errors: Rails.configuration.custom_error_msgs[:record_invalid], status: :bad_request
         end
       end
+      # rubocop:enable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
 
-      # TODO: Add a before_action callback to update,destory and purge_avatar calling a find_user.
       def update
         user = User.find(params[:id])
 
@@ -70,7 +75,7 @@ module Api
           create_default_room(user)
           render_data  status: :ok
         else
-          render_error errors: user.errors.to_a
+          render_error errors: Rails.configuration.custom_error_msgs[:record_invalid]
         end
       end
 
@@ -79,7 +84,7 @@ module Api
         if user.destroy
           render_data  status: :ok
         else
-          render_error errors: user.errors.to_a
+          render_error errors: Rails.configuration.custom_error_msgs[:record_invalid]
         end
       end
 
@@ -96,8 +101,6 @@ module Api
       # Does: Validates and change the user password.
 
       def change_password
-        return render_error status: :unauthorized unless current_user # TODO: generalise this.
-
         return render_error status: :forbidden if current_user.external_id?
 
         old_password = change_password_params[:old_password]
@@ -114,7 +117,7 @@ module Api
       private
 
       def user_params
-        params.require(:user).permit(:name, :email, :password, :avatar, :language, :role_id, :invite_token)
+        @user_params ||= params.require(:user).permit(:name, :email, :password, :avatar, :language, :role_id, :invite_token)
       end
 
       def create_default_room(user)
--- a/app/javascript/components/users/authentication/forms/SignupForm.jsx
+++ b/app/javascript/components/users/authentication/forms/SignupForm.jsx
@@ -19,7 +19,9 @@ export default function SignupForm() {
 
   const handleSubmit = useCallback(async (user) => {
     const results = await captchaRef.current?.execute({ async: true });
-    return createUserAPI.mutate({ user, token: results.response });
+    const token = results?.response || '';
+
+    return createUserAPI.mutate({ user, token });
   }, [captchaRef.current, createUserAPI.mutate]);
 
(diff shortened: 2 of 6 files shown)


app/controllers/api/v1/users_controller.rb
```diff
@@ -7,7 +7,7 @@ class UsersController < ApiController
 
       skip_before_action :ensure_authenticated, only: %i[create]
 
-      before_action only: %i[show update destroy purge_avatar] do
+      before_action except: %i[create change_password] do
```
KH-Amir-TN, line 10: Whitelisting vs blacklisting strategies like with authentication authorization,
With the old approach adding a new API will risk exposing it unintentionally,
Now any API that is not explicitly declared will require the highest permissions to pass and can be spotted easily without introducing a security vulnerability,
If you're okay with this we may want to propagate it through the code base and make a de-facto,

app/controllers/api/v1/users_controller.rb
```diff
@@ -17,23 +17,28 @@ def show
...
       # POST /api/v1/users.json
       # Expects: { user: { :name, :email, :password} }
       # Returns: { data: Array[serializable objects] , errors: Array[String] }
       # Does: Creates and saves a new user record in the database with the provided parameters.
 
       def create
-        # TODO: amir - ensure accessibility for unauthenticated requests only.
-        params[:user][:language] = I18n.default_locale if params[:user][:language].blank?
+        # Users created by a user will have the creator language by default with a fallback to the server configured default_locale.
+        user_params[:language] = current_user&.language || I18n.default_locale if user_params[:language].blank?
```
KH-Amir-TN, line 28: This way if the user was created by an admin it will have the admin language by default and not the admin browser language,
A fallback will be the Greenlight server default locale taken from the `.env`,

app/controllers/api/v1/users_controller.rb
```diff
@@ -17,23 +17,28 @@ def show
...
 
         registration_method = SettingGetter.new(setting_name: 'RegistrationMethod', provider: current_provider).call
 
-        return render_error errors: 'InviteInvalid' if registration_method == SiteSetting::REGISTRATION_METHODS[:invite] && !valid_invite_token
+        if registration_method == SiteSetting::REGISTRATION_METHODS[:invite] && !valid_invite_token
+          return render_error errors: Rails.configuration.custom_error_msgs[:invite_token_invalid]
```
KH-Amir-TN, line 33: Removing literal usage to improve code quality and resiliency,

app/controllers/api/v1/users_controller.rb
```diff
@@ -17,23 +17,28 @@ def show
...
+        end
 
         user = UserCreator.new(user_params: user_params.except(:invite_token), provider: current_provider, role: default_role).call
 
         # TODO: Add proper error logging for non-verified token hcaptcha
-        return render_error errors: 'HCaptchaInvalid' if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
+        if !current_user && hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
+          return render_error errors: Rails.configuration.custom_error_msgs[:hcaptcha_invalid]
```
KH-Amir-TN, line 40: Same here,

app/controllers/api/v1/users_controller.rb
```diff
@@ -55,22 +60,22 @@ def create
           create_default_room(user)
 
           render_data data: current_user, serializer: CurrentUserSerializer, status: :created
-        elsif user.errors.to_a == ['Email has already been taken']
-          render_error errors: 'EmailAlreadyExists', status: :bad_request
+        elsif user.errors.size == 1 && user.errors.of_kind?(:email, :taken)
+          render_error errors: Rails.configuration.custom_error_msgs[:email_exists], status: :bad_request
         else
-          render_error errors: user.errors.to_a, status: :bad_request
+          render_error errors: Rails.configuration.custom_error_msgs[:record_invalid], status: :bad_request
```
KH-Amir-TN, line 66: Removing literal usage,

app/controllers/api/v1/users_controller.rb
```diff
@@ -55,22 +60,22 @@ def create
...
-      # TODO: Add a before_action callback to update,destory and purge_avatar calling a find_user.
       def update
         user = User.find(params[:id])
 
         if user.update(user_params)
           create_default_room(user)
           render_data  status: :ok
         else
-          render_error errors: user.errors.to_a
+          render_error errors: Rails.configuration.custom_error_msgs[:record_invalid]
```
KH-Amir-TN, line 78: Sending the plain active record errors is a bad practice,
We can log them once tackling logging.
However, we need to send an error message without details to the client app that it can leverage later on for feedbacking,

app/controllers/api/v1/users_controller.rb
```diff
@@ -79,7 +84,7 @@ def destroy
         if user.destroy
           render_data  status: :ok
         else
-          render_error errors: user.errors.to_a
+          render_error errors: Rails.configuration.custom_error_msgs[:record_invalid]
```
KH-Amir-TN, line 87: Same here,

app/controllers/api/v1/users_controller.rb
```diff
@@ -96,8 +101,6 @@ def purge_avatar
       # Does: Validates and change the user password.
 
       def change_password
-        return render_error status: :unauthorized unless current_user # TODO: generalise this.
```
KH-Amir-TN, line 99: That was removed in favor of the default `before_action :ensure_authenticated`,

app/controllers/api/v1/users_controller.rb
```diff
@@ -114,7 +117,9 @@ def change_password
       private
 
       def user_params
-        params.require(:user).permit(:name, :email, :password, :avatar, :language, :role_id, :invite_token)
+        return @user_params unless @user_params.nil?
```
KH-Amir-TN, line 120: With our current way of using the `ActionController::Parameters` API we're having two issues:
1. The filtering operation is required only once, for each subsequent call that is a waste of computation and not required delay.
2. Each call creates a new instance.
3. Because of 2. the params methods will give an immutable instance.
4. Because of 3. errors can be created easily without being caught (e.g. expecting that calling user_params[:language] = I18n.default_locale will change the user language which will not do).

KH-Amir-TN: If you're okay with this, then I believe we may want to generalize this throughout the codebase.

app/javascript/components/users/authentication/forms/SignupForm.jsx
```diff
@@ -19,7 +19,9 @@ export default function SignupForm() {
 
   const handleSubmit = useCallback(async (user) => {
     const results = await captchaRef.current?.execute({ async: true });
-    return createUserAPI.mutate({ user, token: results.response });
+    const token = results?.response || '';
```
KH-Amir-TN, line 22: Solved the signup issue introduced by #4299.

app/javascript/hooks/mutations/admin/manage_users/useDeleteUser.jsx
```diff
@@ -8,7 +8,7 @@ export default function useDeleteUser(userId) {
   const queryClient = useQueryClient();
 
   return useMutation(
-    (data) => axios.delete(`/admin/users/${userId}.json`, data),
+    (data) => axios.delete(`/users/${userId}.json`, data),
```
KH-Amir-TN, line 11: Solved the user deletion on the admin panel issue,

spec/controllers/users_controller_spec.rb
```diff
@@ -163,29 +188,29 @@
...
           allow(SettingGetter).to receive(:new).with(setting_name: 'RegistrationMethod', provider: 'greenlight').and_return(reg_method)
-          allow(reg_method).to receive(:call).and_return('approval')
+          allow(reg_method).to receive(:call).and_return(SiteSetting::REGISTRATION_METHODS[:approval])
         end
 
         it 'sets a user to pending when registering' do
           expect { post :create, params: user_params }.to change(User, :count).from(0).to(1)
 
-          expect(User.find_by(email: user_params[:user][:email]).status).to eq('pending')
+          expect(User.find_by(email: user_params[:user][:email])).to be_pending
```
KH-Amir-TN, line 213: User#status is an enum.
ActiveRecord provides useful helpers like (active?, pending?,...) for each enum.
Instead of working with literals we can leverage those helpers,

spec/controllers/users_controller_spec.rb
```diff
@@ -106,7 +116,22 @@
 
         expect(ActionMailer::MailDeliveryJob).not_to have_been_enqueued
         expect(response).to have_http_status(:bad_request)
-        expect(JSON.parse(response.body)['errors']).to be_present
+        expect(JSON.parse(response.body)['errors']).to eq(Rails.configuration.custom_error_msgs[:record_invalid])
+      end
+
+      context 'Duplicated email' do
```
KH-Amir-TN, line 122: Missing test,

app/controllers/api/v1/users_controller.rb
farhatahmad: Can we not jsut do `@user_params ||= params.require...`

app/controllers/api/v1/users_controller.rb
scouillard: I'm not following - isn't whitelisting more secure?

app/controllers/api/v1/users_controller.rb
scouillard: Is this common? I don't think I've seen this

app/controllers/api/v1/users_controller.rb
KH-Amir-TN: Yes, whitelisting is more secure.
The old behavior conceptually was a blocklist (all are unrestricted except those mentioned explicitly in the `:only` list).
Now it's more of a whitelist where by default any API in the users controller will be restricted unless those mentioned explicitly in the whitelist (the `except:` list).

app/controllers/api/v1/users_controller.rb
KH-Amir-TN: @farhatahmad sure,

KH-Amir-TN: @scouillard what exactly?

KH-Amir-TN: @scouillard we may want to generalize this throughout the codebase to reduce error surface and to improve performance even when we're not mutating the params since there's always a great chance of reusing filtered params in the same request response cycle.

farhatahmad (approved):

PR #4312, opened 2022-12-21
Solved Critical Issue Allowing Users to Change their Role 

alihadi-mazeh:
## Description

Added a Shared Access Serializer to prevent exposing the role id that was being exposed prior to this change. 
Updated the users controller to account to check if a user is editing themselves or tying to edit another user; same check goes for an admin as well.
Added corresponding rspec tests to test for these changes

## Testing Steps

Added 3 tests to check for 3 different conditions:
- Does not allow the user to change their own role 
    - sign in with user
    - pass in updated parameters including the role id in a PUT request 
    - expect the user id to  **not** be equal to the updated role id 
- An admin trying to edit themselves
    - sign in with admin
    - pass in updated parameters including the role id in a PUT request
    - expect the admin user's id to be equal to the updated role id
- An admin trying to edit another user
     - sign in with admin
     - pass in updated parameters including the role id in a PUT request
     - expect the changed user's id to be equal to the updated role id

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -70,6 +70,12 @@ module Api
 
       def update
         user = User.find(params[:id])
+        # user is updating themselves
+
+        if current_user.id == params[:id] && !PermissionsChecker.new(permission_names: 'ManageUsers', current_user:, user_id: nil, record_id: nil,
+                                                                     friendly_id: nil).call
+          params[:user].delete(:role_id)
+        end
 
         if user.update(user_params)
           create_default_room(user)
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -71,8 +71,7 @@ module Api
       def update
         user = User.find(params[:id])
         # user is updating themselves
-        if current_user.id == params[:id] && !PermissionsChecker.new(permission_names: 'ManageUsers', current_user:, user_id: nil, record_id: nil,
-                                                                     friendly_id: nil).call
+        if current_user.id == params[:id] && !PermissionsChecker.new(permission_names: 'ManageUsers', current_user:, current_provider:).call
           params[:user].delete(:role_id)
         end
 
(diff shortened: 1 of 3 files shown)


spec/controllers/users_controller_spec.rb
```diff
@@ -243,6 +241,41 @@
...
+        role_id: create(:role, name: 'New Role').id
+      }
+      patch :update, params: { id: user.id, user: updated_params }
+
+      user.reload
+
+      expect(user.role_id).not_to eq(updated_params[:role_id])
+    end
+
+    it 'allows an admin to edit themself' do
```
hadicheaito1, line 256: Instead I'd say, "allows a user with ManageUsers permission to edit their own role"

spec/controllers/users_controller_spec.rb
```diff
@@ -243,6 +241,41 @@
...
+      updated_params = {
+        role_id: create(:role, name: 'New Role').id
+      }
+      patch :update, params: { id: user_with_manage_users_permission.id, user: updated_params }
+
+      user_with_manage_users_permission.reload
+      expect(user_with_manage_users_permission.role_id).to eq(updated_params[:role_id])
+    end
+
+    it 'allows an admin to edit someone else' do
```
hadicheaito1, line 267: I'd say, "allows a user with ManageUsers permission to edit  another users role"

app/controllers/api/v1/users_controller.rb
```diff
@@ -65,6 +65,12 @@ def create
       # TODO: Add a before_action callback to update,destory and purge_avatar calling a find_user.
       def update
         user = User.find(params[:id])
+        # user is updating themselves
+
```
hadicheaito1, line 69: remove this extra empty line

hadicheaito1 (changes requested):

hadicheaito1:
@alihadi-mazeh You have a rubocop complaint. You also accidentally put the same description twice for two different tests

farhatahmad (approved):

PR #4318, opened 2022-12-22
Password management forms refactoring: Refactored `ChangePwdForm`

KH-Amir-TN:
## Description

Refactored ChangePwdForm.
  + Updated form hooks to introduce a reset API.
  + Fixed losing initial defaulValues when injecting some defaults externally because of shallow merging.

## TODO: 
  - [ ] Fix `_defaultValues` identity.

## Testing Steps

> 1. Pull the code.


KH-Amir-TN:
# DEPENDS ON #4317.

app/javascript/hooks/forms/users/authentication/useSignInForm.js
```diff
@@ -0,0 +1,68 @@
...
+
+export function useSignInFormValidation() {
+  return useMemo(() => (yup.object({
+    email: yup.string().required('forms.validations.email.required').email('forms.validations.email.email'),
+    password: yup.string().required('forms.validations.password.required'),
+    extend_session: yup.bool(''),
+  })), []);
+}
+
+export default function useSignInForm({ _defaultValues, ..._config } = {}) {
```
KH-Amir-TN, line 15: That's part of the config merging issue mentioned below.
By default the configuration will be an empty object and the `defaultValues` will be extracted into a separate variable when the rest of the config props will be collected in one level deep object `_config`.

app/javascript/hooks/forms/users/authentication/useSignInForm.js
```diff
@@ -0,0 +1,68 @@
...
+        ...{
+          email: '',
+          password: '',
+          extend_session: false,
+        },
+        ..._defaultValues,
+      },
+      resolver: yupResolver(validationSchema),
+    },
+    ..._config,
```
KH-Amir-TN, line 60: When configuring the form and providing the default values.
Sometimes we may need to inject more configuration externally.
The idea was to merge the local config with the injected config prioritizing the latter.
However, that will not work since the react-hook-form config object is multi dimensional (or multi level).
The defaultValues prop is an object with another level deep.
So shallow merging the defaultValues will overwrite it meaning that to extend one default value, one is required to redefine all of the others (which means being aware of the underlying implementation of the form hook that removes its usage).
The solution is to deeply merge the objects which is complicated to do or to use through a lib,
Since the react-hook-form API confirms that only the defaultValues prop is an object.
The idea was to extract it and merge it independently therefore making a deep merge with two levels of shallow merge.

app/javascript/components/users/password_management/forms/ChangePwdForm.jsx
```diff
@@ -1,41 +1,26 @@
...
-            old_password: '',
-            new_password: '',
-            password_confirmation: '',
-          })}
-        >
-          { t('cancel') }
-        </Button>
-        <Button variant="brand" type="submit" disabled={changePwd.isLoading}>
-          {changePwd.isLoading && <Spinner className="me-2" />}
+        <Button variant="neutral" onClick={reset}> { t('cancel') } </Button>
```
KH-Amir-TN, line 21: Using the reset API that is introduced below,

farhatahmad (approved):

Issue #4085, opened 2022-11-08
V3: Migrate room settings, presentations, etc


debuglevel:
**Describe the bug**
https://docs.bigbluebutton.org/greenlight_v3/gl3-migration.html states that "Rooms settings, presentations and shared access will not propagate from v2 to v3, so migrated rooms will have v3 the default configurations assigned."

**Expected behavior**
A migration should migrate data from one system to another as good as possible. If features between versions do not match perfectly, the migration should derive/guess information.

I as well as our users would not expect data loss if we update a software.

**Actual behavior**
It only migrates a subset of the original data.

farhatahmad:
Due to the differences in the the data structures, moving over shared access and presentation is very difficult. Room settings are also tricky because there are some changed and some new settings, so guessing which value they would want isn't really something we want to do.

During the community calls that were done during v3, the majority of those attending were okay with resetting their room settings

farhatahmad:
Thanks for the feedback - we'll attempt to tackle this before heading to production

PR #4339, opened 2022-12-28
Migration upgrade (V2): addition of RolePermissions, RoomSettings, SharedAccess, SiteSettings, RoomConfiguration data

scouillard:
## Description

Regarding https://github.com/bigbluebutton/greenlight/issues/4085

The existing migrations tasks (Roles, Users, Rooms) have been upgraded to include the subsequent data (Role Permissions, Room Settings, Shared Access).

Also, the SiteSetting migration task has been added and will migrate the SiteSettings and the RoomsConfiguration data.

## Testing Steps

Duplicate GLv2 and GLv3 so that the apps are located in a different directory and open a terminal for both directory.

**For GLv1**
1. Launch GLv3 with the following commands
```
rails db:create
rails db:migrate:with_data
./bin/dev
```

**For GLv2**
1. Launch GLv2 with the following commands
```
rails db:create
rails db:migrate
rails db:seed
rails s -p 3001
```

3. Populate GLv2 database

- Access GLV2 via your browser [(localhost:3001](http://localhost:3001/) so that the default settings are loaded into the database.
- Log in via the default Admin account (admin@example.com // Administrator1!) 
- Create new Role(s), edit the role settings
- Create new User(s), assign the users different roles
- Create new Room(s), edit the room settings, share the room
- Edit the SiteSettings and the RoomsConfigurations

4. Migrate the data to GLv3

- Make sure that GLv3 server is up and running
- In the terminal, run the following tasks in that order

```
rake migrations:roles
rake migrations:users
rake migrations:rooms
rake migrations:settings
```

## TODO
Improve error-logging

### DIFF
--- a/lib/tasks/migrations/migrations.rake
+++ b/lib/tasks/migrations/migrations.rake
@@ -22,14 +22,14 @@ namespace :migrations do
       role_permissions_hash = RolePermission.where(role_id: r.id).pluck(:name, :value).to_h
       # Returns nil if the Role Permission value is the same as the corresponding default value in V3
       role_permissions = {
-        CreateRoom: role_permissions_hash['can_create_rooms'] == true ? nil : "false",
-        CanRecord: role_permissions_hash['can_launch_recording'] == true ? nil : "false",
-        ManageUsers: role_permissions_hash['can_manage_users'],
-        ManageRoles: role_permissions_hash['can_edit_roles'],
+        CreateRoom: role_permissions_hash['can_create_rooms'] == "true" ? nil : "false",
+        CanRecord: role_permissions_hash['can_launch_recording'] == "true" ? nil : "false",
+        ManageUsers: role_permissions_hash['can_manage_users'] == "false" ? nil : "true",
+        ManageRoles: role_permissions_hash['can_edit_roles'] == "false" ? nil : "true",
         # In V3, can_manage_room_recordings is split into two distinct permissions: ManageRooms and ManageRecordings
-        ManageRooms: role_permissions_hash['can_manage_room_recordings'],
-        ManageRecordings: role_permissions_hash['can_manage_room_recordings'],
-        ManageSiteSettings: role_permissions_hash['edit_site_settings']
+        ManageRooms: role_permissions_hash['can_manage_rooms_recordings'] == "false" ? nil : "true",
+        ManageRecordings: role_permissions_hash['can_manage_room_recordings'] == "false" ? nil : "true",
+        ManageSiteSettings: role_permissions_hash['can_edit_site_settings'] == "false" ? nil : "true"
       }.compact
 
       params = { role: { name: r.name.capitalize,
--- a/lib/tasks/migrations/migrations.rake
+++ b/lib/tasks/migrations/migrations.rake
@@ -139,7 +139,8 @@ namespace :migrations do
     start, stop = range(args)
     has_encountred_issue = 0
 
-    User.joins(shared_access: [:rooms])
+    User.unscoped
+        .joins(shared_access: [:rooms])
         .select(:id, :'rooms.uid', :email)
         .find_each(start: start, finish: stop, batch_size: COMMON[:batch_size]) do |sa|
             params = { shared_access: { friendly_id: sa.uid,
@@ -149,18 +150,19 @@ namespace :migrations do
 
             case response
             when Net::HTTPCreated
-              puts green "Successfully migrated Site Settings"
+              puts green "Successfully migrated Shared Access"
             else
-              puts red "Unable to migrate Site Settings"
+              puts red "Unable to migrate Shared Access"
               has_encountred_issue = 1 # At least one of the migrations failed.
             end
 
             puts
-            puts green "Site Settings migration completed."
+            puts green "Shared Access migration completed."
 
             unless has_encountred_issue.zero?
               puts yellow "In case of an error please retry the process to resolve."
-              puts yellow "If you have not migrated your users, kindly run 'rake migrations:site_settings' first and then retry."
+              puts yellow "If you have not migrated your users, kindly run 'rake migrations:users' first and then retry."
+              puts yellow "If you have not migrated your rooms, kindly run 'rake migrations:rooms' first and then retry."
             end
 
             exit has_encountred_issue
@@ -195,7 +197,6 @@ namespace :migrations do
 
     unless has_encountred_issue.zero?
       puts yellow "In case of an error please retry the process to resolve."
-      puts yellow "If you have not migrated your users, kindly run 'rake migrations:site_settings' first and then retry."
     end
 
     exit has_encountred_issue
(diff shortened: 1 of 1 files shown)


scouillard:
Decided to add a standalone task for RoomSettings instead of adding it to the Room task.

It is less efficient but it feels like the right thing to do:

- the endpoint in V3 for rooms would be very crowded
- it is safer, in case of an error you don't have to migrate all the rooms again

lib/tasks/migrations/migrations.rake
```diff
@@ -121,6 +121,91 @@ namespace :migrations do
...
+      parsed_room_settings = JSON.parse(r.room_settings)
+      room_settings = {
+        record: parsed_room_settings["recording"],
+        muteOnStart: parsed_room_settings["muteOnStart"],
+        guestPolicy: parsed_room_settings["requireModeratorApproval"],
+        glAnyoneCanStart: parsed_room_settings["anyoneCanStart"],
+        glAnyoneJoinAsModerator: parsed_room_settings["joinModerator"],
+        glViewerAccessCode: r.access_code,
+        glModeratorAccessCode: r.moderator_access_code
+      }
```
KH-Amir-TN, line 147: I believe the logic here is solid and will answer the feature requests,
And the keys mapping is correct, well done!
However, there's few things to note, so that not all the values in v2 room settings are directly migratable to v3, :
1. Some values needs to be mapped for compatibility:
    - All values on v2 room settings are Booleans verses being strings on v3 (rails should be able to cast that automatically on v3 but we may need more control over this to reduce error surface.).
  2. Room meetings options values on v3 aren't binary meaning there's more then 2 possible values unlike on v2, e.g. `"requireModeratorApproval" => true` on v2 maps to `"guestPolicy" => "ALWAYS_ACCEPT"` and not "guestPolicy" => "true"`.
 3. IMO Access codes may be better to be left to be generated by v3, on v2 they were generated insecurely which was patched on v3.
 Migrating insecure codes to v3 will break the security mechanism it introduced for generating access codes until someone decides to regenerate codes (which since we allow them to migrate old used codes they won't bather changing it).

KH-Amir-TN: Here's an example of a room settings Json object on v2:
```json
{"muteOnStart":true,"requireModeratorApproval":true,"anyoneCanStart":true,"joinModerator":true,"recording":false}
```

KH-Amir-TN (commented):
Why don't we extend the `migrations:rooms` task to migrate a room along with its formatted settings?
If a room succeeds to migrate that means that its settings migrated, we gain more performance (same API call instead of two), faster complete system migrations and less overhead (how can we effectively determine that a room settings were already been migrated successfully if the room settings does not contain an identifier?).

lib/tasks/migrations/migrations.rake
scouillard: 1 is in progress
2 Great call. will purge Access Codes. Users will have to re-generate an Access Code if needed.

lib/tasks/migrations/migrations.rake
```diff
@@ -121,6 +121,91 @@ namespace :migrations do
...
+    has_encountred_issue = 0
+
+    filtered_roles_ids = Role.unscoped
+                             .select(:id, :name)
+                             .where(name: COMMON[:filtered_user_roles])
+                             .pluck(:id)
+
+    Room.unscoped
+        .select(:uid, :room_settings)
+        .where.not(users: { role_id: filtered_roles_ids, deleted: true }, deleted: true)
```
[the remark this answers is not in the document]
scouillard: I am not sure about splitting Room and Room Settings, I can bundle the logic instead if it gets too complicated/inefficient

lib/tasks/migrations/migrations.rake
```diff
@@ -121,6 +121,91 @@ namespace :migrations do
...
+      # Returns nil if the Room Setting value is the same as the corresponding default value in V3
+      room_settings = {
+        record: parsed_room_settings["recording"] == false ? nil : "true",
+        muteOnStart: parsed_room_settings["muteOnStart"] == false ? nil : "true",
+        glAnyoneCanStart: parsed_room_settings["anyoneCanStart"] == false ? nil : "true",
+        glAnyoneJoinAsModerator: parsed_room_settings["joinModerator"] == false ? nil : "true",
+        guestPolicy: parsed_room_settings["requireModeratorApproval"] == false ? nil : "ASK_MODERATOR",
+      }
+
+      params = { room: { friendly_id: r.uid, room_settings: room_settings.compact } }
```
scouillard, line 149: Made it so only RoomSettings values that differs from the default values of V3 are sent

scouillard: `room_settings.compact` on line 149 will remove the keys with nil value

lib/tasks/migrations/migrations.rake
```diff
@@ -121,6 +121,91 @@ namespace :migrations do
...
+
+    params = { settings: { PrimaryColor: Rails.configuration.primary_color_default,
+                                PrimaryColorLight: Rails.configuration.primary_color_lighten_default,
+                                PrimaryColorDark: Rails.configuration.primary_color_darken_default,
+                                # Terms: ,
+                                # BrandingImage: ,
+                                # PrivacyPolicy: ,
+                                RegistrationMethod: Rails.configuration.registration_method_default,
+                                ShareRooms: Rails.configuration.shared_access_default,
+                                PreuploadPresentation: Rails.configuration.preupload_presentation_default } }
```
scouillard, line 186: Still got to research the following settings: Terms, BrandingImage and PrivacyPolicy

lib/tasks/migrations/migrations.rake
scouillard: v2 seems to support only "ASK_MODERATOR" (and not "ALWAYS_ACCEPT") according to line 78 in bbb_server.rb

scouillard:
I've been adamant about not sending any data that correspond to a V3 default value.

For example, in a room settings, all the settings values are set to "false" by default.
So, instead of sending { room_settings: { record: 'false', muteOnStart: 'true' } },
V2 would send { room_settings: { muteOnStart: 'true' } }

This saves many update operations on the database but it is adding some ugly logic.

lib/tasks/migrations/migrations.rake
```diff
@@ -121,6 +157,53 @@ namespace :migrations do
...
+  task settings: :environment do |_task|
+    has_encountred_issue = 0
+
+    settings_hash = Setting.find_by(provider: 'greenlight').features.pluck(:name, :value).to_h
+
+    # SiteSettings
+    site_settings = {
+      PrimaryColor: settings_hash['Primary Color'],
+      PrimaryColorLight: settings_hash['Primary Color Lighten'],
+      PrimaryColorDark: settings_hash['Primary Color Lighten'],
```
farhatahmad, line 169: This setting doesn't exist in v3 - no need to send it over

farhatahmad:
Very nicely done @scouillard

PR #4378, opened 2022-12-29
Migration upgrade (V3): addition of RolePermissions, RoomSettings, SharedAccess, SiteSettings, RoomConfiguration data

scouillard:
## Description

Regarding https://github.com/bigbluebutton/greenlight/issues/4085

The existing migrations endpoints (Roles, Users, Rooms) have been upgraded to include the subsequent data (Role Permissions, Room Settings, Shared Access).

Also, the SiteSetting migration endpoint has been added and will migrate the SiteSettings and the RoomsConfiguration data.

## Testing Steps

Duplicate GLv2 and GLv3 so that the apps are located in a different directory and open a terminal for both directory.

**For GLv1**
1. Launch GLv3 with the following commands
```
rails db:create
rails db:migrate:with_data
./bin/dev
```

**For GLv2**
1. Launch GLv2 with the following commands
```
rails db:create
rails db:migrate
rails db:seed
rails s -p 3001
```

3. Populate GLv2 database

- Access GLV2 via your browser [(localhost:3001](http://localhost:3001/) so that the default settings are loaded into the database.
- Log in via the default Admin account (admin@example.com // Administrator1!) 
- Create new Role(s), edit the role settings
- Create new User(s), assign the users different roles
- Create new Room(s), edit the room settings, share the room
- Edit the SiteSettings and the RoomsConfigurations

4. Migrate the data to GLv3

- Make sure that GLv3 server is up and running
- In the terminal, run the following tasks in that order

```
rake migrations:roles
rake migrations:users
rake migrations:rooms
rake migrations:settings
```

## TODO
Improve error-logging

### DIFF
--- a/app/controllers/api/v1/migrations/external_controller.rb
+++ b/app/controllers/api/v1/migrations/external_controller.rb
@@ -69,7 +69,7 @@ module Api
         # POST /api/v1/migrations/rooms.json
         # Expects: { room: { :name, :friendly_id, :meeting_id, :last_session } }
         # Returns: { data: Array[serializable objects] , errors: Array[String] }
-        # Does: Creates a room.
+        # Does: Creates a Room and its RoomMeetingOptions.
         def create_room
           room_hash = room_params.to_h
 
@@ -89,8 +89,36 @@ module Api
 
           return render_error status: :bad_request unless room.save
 
-          # RoomMeetingOption
-          update_room_meeting_options(room_settings: room_hash[:room_settings], room_meeting_options: room.room_meeting_options)
+          # As per the task file in V2, V3 should receive and update ONLY the RoomMeetingOptions that differs from V3 default values
+          # In V3, RoomMeetingOptions default values are set to either "false" or "ALWAYS_ACCEPT" for guestPolicy
+          # Hence, the following is an example of an expected room_settings hash from V2: { "record": "true", glAnyoneCanStart": "true" }
+          room_hash[:room_settings].each do |name, value|
+            # Finds the RoomMeetingOption (V3) corresponding to the given RoomSetting (V2)
+            room_meeting_option = room.room_meeting_options.includes(:meeting_option).find_by(meeting_option: { name: })
+            return render_error status: :bad_request unless room_meeting_option
+
+            # Updates the RoomMeetingOption value
+            room_meeting_option.update!(value:) if room_meeting_option.value != value
+            return render_error status: :bad_request unless room_meeting_option.save
+          end
+
+          render_data status: :created
+        end
+
+        # POST /api/v1/migrations/shared_access.json
+        # Expects: { shared_access: { :friendly_id, :user_email } }
+        # Returns: { data: Array[serializable objects] , errors: Array[String] }
+        # Does: Creates a SharedAccess.
+        def create_shared_access
+          shared_access_hash = shared_access_params.to_h
+
+          room = Room.find_by(friendly_id: shared_access_hash[:friendly_id])
+          return render_error status: :bad_request unless room
+
+          user = User.find_by(email: shared_access_hash[:user_email])
+          return render_error status: :bad_request unless user
+
+          SharedAccess.create!(room:, user:)
 
           render_data status: :created
         end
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -90,21 +110,78 @@ def create_room
...
 
         def user_params
           decrypted_params.require(:user).permit(:name, :email, :external_id, :language, :role)
         end
 
         def room_params
-          decrypted_params.require(:room).permit(:name, :friendly_id, :meeting_id, :last_session, :owner_email)
+          decrypted_params.require(:room).permit(:name, :friendly_id, :meeting_id, :last_session, :owner_email, room_settings: {},
+                                                                                                                shared_users_emails: [])
+        end
```
scouillard, line 181: Having an empty hash or and empty array in permit is not optimal but I believe that specifying every nested element would require a refactoring of the data structure.

app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -21,26 +21,41 @@ class ExternalController < ApiController
...
 
           return render_error status: :bad_request unless role.save
 
+          # Returns unless the Role has a RolePermission that differs from V3 default RolePermissions values
+          return render_data status: :created unless role_hash[:role_permissions].any?
+
+          # Finds & Updates the associated RolePermissions associated with the Role
+          role_permissions = RolePermission.where(role_id: role.id).includes(:permission)
+          # The role_permissions hash contains only the permissions with value that differs from V3 default values
+          role_hash[:role_permissions].any? && role_hash[:role_permissions].each do |name, value|
```
farhatahmad, line 44: Is the `.any?` necessary here? Can we not just do `.each`? I believe the `.each` can handle `nil`

app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -21,26 +21,41 @@ class ExternalController < ApiController
...
+            role_permission = role_permissions.find_by('permissions.name': name)
+            return render_error status: :bad_request unless role_permission
+
+            role_permission.update!(value:)
+            return render_error status: :bad_request unless role_permission.save
```
farhatahmad, line 49: This will result in a ton of queries. Can we not do 1 find and 1 update_all?  See: https://api.rubyonrails.org/v3.2.0/classes/ActiveRecord/Relation.html#method-i-update

farhatahmad: You're also doing an `update` then a `save` which doesn't really make sense

app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -90,21 +107,76 @@ def create_room
 
           return render_error status: :bad_request unless room.save
 
+          # Finds & Updates the RoomMeetingOptions associated with the Room
+          room_meeting_options = RoomMeetingOption.where(room:).includes(:meeting_option)
+          # The room_settings hash contains only the MeetingOption with value that differs from V3 rooms default values
+          room_hash[:room_settings].any? && room_hash[:room_settings].each do |name, value|
```
farhatahmad, line 113: Same comments as above

app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -90,21 +107,76 @@ def create_room
...
+        #                        room_configurations: { :record, :muteOnStart, :guestPolicy, :glAnyoneCanStart,
+        #                          :glAnyoneJoinAsModerator, :glRequireAuthentication } } }
+        # Returns: { data: Array[serializable objects] , errors: Array[String] }
+        # Does: Creates a SiteSettings or a RoomsConfiguration.
+        def create_settings
+          settings_hash = settings_params.to_h
+
+          render_data status: :created unless settings_hash.any?
+
+          settings_hash[:site_settings].any? && settings_hash[:site_settings].each do |name, value|
```
farhatahmad, line 144: Same as above

app/controllers/api/v1/migrations/external_controller.rb
scouillard: `role_hash[:role_permissions]` will return an empty hash, not `nil`
The same logic is repeated:
The hash will be populated with RolePermissions, RoomSettings that differs from V3 default values.
If no RolePermissions, RoomSettings differ from default values, `params_hash[:role_permissions]` and `params_hash[:room_settings] `will return an empty hash.

app/controllers/api/v1/migrations/external_controller.rb
scouillard: Right, using `update_all` would be great.

scouillard:
Rework so it updates the join table records in one query.
Ready for review BUT has not been tested in production yet.

app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -90,21 +105,76 @@ def create_room
...
+          room_meeting_options = room_meeting_options_temp.map { |k, v| [k, { value: room_hash[:room_settings][v.to_sym] }] }
+          RoomMeetingOption.update!(room_meeting_options)
+
+          # Creates the associated SharedAccess
+          room_hash[:shared_users_emails].any? && room_hash[:shared_users_emails].each do |email|
+            user = User.find_by(email:)
+            return render_error status: :bad_request unless user
+
+            SharedAccess.create!(room_id: room.id, user_id: user.id)
+          end
```
scouillard, line 123: Should be able to create all the related SharedAccess in a single query as well

farhatahmad (approved):

PR #4410, opened 2023-01-04
Complete invitation reg method

farhatahmad:
## Description

Still need to test this properly to make sure it works

### DIFF
--- a/app/controllers/external_controller.rb
+++ b/app/controllers/external_controller.rb
@@ -15,6 +15,13 @@ class ExternalController < ApplicationController
       verified: true
     }
 
+    registration_method = SettingGetter.new(setting_name: 'RegistrationMethod', provider: current_provider).call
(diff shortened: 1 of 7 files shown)


app/controllers/external_controller.rb
```diff
@@ -15,6 +15,13 @@ def create_user
       verified: true
     }
 
+    registration_method = SettingGetter.new(setting_name: 'RegistrationMethod', provider: current_provider).call
+
+    # Check if they have a valid token
+    if registration_method == SiteSetting::REGISTRATION_METHODS[:invite] && !valid_invite_token(email: user_info[:email])
+      raise StandardError, Rails.configuration.custom_error_msgs[:invite_token_invalid]
```
farhatahmad, line 22: Temporary - need to see what this means for the app

PR #4417, opened 2023-01-06
Server Rooms fix + sorting order

hadicheaito1:
## Description

- Fixing the bug for ServerRooms owner sorting
- Changed sorting order

### DIFF
--- a/app/controllers/api/v1/admin/server_rooms_controller.rb
+++ b/app/controllers/api/v1/admin/server_rooms_controller.rb
@@ -12,10 +12,10 @@ module Api
 
         # GET /api/v1/admin/server_rooms.json
         def index
-          sort_config = config_sorting(allowed_columns: %w[name user.name])
+          sort_config = config_sorting(allowed_columns: %w[name users.name])
 
-          rooms = Room.includes(:user).joins(:user).where(users: { provider: current_provider }).order(sort_config,
-                                                                                                       online: :desc)&.admin_search(params[:search])
+          rooms = Room.includes(:user).joins(:user).where(users: { provider: current_provider }).order(sort_config, online: :desc)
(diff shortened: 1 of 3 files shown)


alihadi-mazeh (dismissed):

app/controllers/api/v1/admin/server_rooms_controller.rb
```diff
@@ -12,10 +12,10 @@ class ServerRoomsController < ApiController
 
         # GET /api/v1/admin/server_rooms.json
         def index
-          sort_config = config_sorting(allowed_columns: %w[name user.name])
+          sort_config = config_sorting(allowed_columns: %w[name users.name])
 
-          rooms = Room.includes(:user).joins(:user).where(users: { provider: current_provider }).order(sort_config,
-                                                                                                       online: :desc)&.admin_search(params[:search])
+          rooms = Room.includes(:user).joins(:user).where(users: { provider: current_provider }).order(sort_config, online: :desc)
+                      .order(Room.arel_table[:last_session].desc.nulls_last)&.admin_search(params[:search])
```
farhatahmad, line 18: I think `.order('last_session DESC NULLS LAST')` might be a bit cleaner

farhatahmad: Can you try doing it in one call? `.order(sort_config, online: :desc, 'last_session DESC NULLS LAST')` Not sure if that'll work

hadicheaito1: `.order('last_session DESC NULLS LAST')` works. The arel table stuff is also done in server rooms. I will change it there too since you find that cleaner

farhatahmad (approved):

PR #4425, opened 2023-01-09
Fix visual issue with avatar load

farhatahmad:


app/javascript/components/users/user/Avatar.jsx
```diff
@@ -19,9 +19,10 @@ export default function Avatar({ avatar, radius, className }) {
         width={radius}
         height={radius}
         className={className}
+        style={{ display: isLoaded ? '' : 'none' }}
```
farhatahmad, line 22: Essentially, the avatar was still taking up space while it was loading, which was causing the visual glitch. This change will make the avatar invisible until it loads in (replacing the placeholder)

scouillard: Interesting

scouillard (approved):

PR #4428, opened 2023-01-10
Add Empty Rooms List layout

scouillard:
## Description

Add EmptyRoomsList layout so that user are prompted to create a Room after creating an account.
Showed our demo to a friend, and he was confused because the landing page was empty - this should help guide a new user.
Not sure about the text, please comment if you have a better text.

image

### DIFF
--- a/app/assets/locales/en.json
+++ b/app/assets/locales/en.json
@@ -92,6 +92,8 @@
     "last_session": "Last Session: {{ room.last_session }}",
     "no_last_session": "No previous session created",
     "no_rooms_found": "No rooms found",
+    "rooms_list_is_empty": "Your rooms list is empty!",
+    "rooms_list_empty_create_room": "Create your first room by clicking on the button below and entering a room name.",
     "meeting": {
       "start_meeting": "Start Meeting",
       "join_meeting": "Join Meeting",
--- a/app/assets/locales/en.json
+++ b/app/assets/locales/en.json
@@ -92,7 +92,7 @@
     "last_session": "Last Session: {{ room.last_session }}",
     "no_last_session": "No previous session created",
     "no_rooms_found": "No rooms found",
-    "rooms_list_is_empty": "Your rooms list is empty!",
+    "rooms_list_is_empty": "You don't have any room yet!",
     "rooms_list_empty_create_room": "Create your first room by clicking on the button below and entering a room name.",
     "meeting": {
       "start_meeting": "Start Meeting",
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/rooms.scss
```diff
@@ -108,3 +108,9 @@
     opacity: 0.6;
   }
 }
+
+#rooms-list-empty {
+  svg {
+    padding-top: 2rem;
+  }
+}
```
farhatahmad, line 116: This can be done in bootstrap

app/assets/locales/en.json
```diff
@@ -92,6 +92,8 @@
     "last_session": "Last Session: {{ room.last_session }}",
     "no_last_session": "No previous session created",
     "no_rooms_found": "No rooms found",
+    "rooms_list_is_empty": "You don't have any room yet!",
```
farhatahmad, line 95: "any rooms" is more correct I believe

app/assets/stylesheets/rooms.scss
scouillard: It can't, pt-4 is 1.5rem and pt-5 is 3rem.

app/javascript/components/rooms/EmptyRoomsList.jsx
```diff
@@ -0,0 +1,36 @@
...
+import UserBoardIcon from './UserBoardIcon';
+
+export default function EmptyRoomsList() {
+  const { t } = useTranslation();
+  const currentUser = useAuth();
+  const mutationWrapper = (args) => useCreateRoom({ userId: currentUser.id, ...args });
+
+  return (
+    <div id="rooms-list-empty">
+      <Card className="border-0 shadow-sm mt-5 text-center">
```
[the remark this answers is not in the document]
scouillard: I think it looks better full width.
image

And it fits with the rest of the cards layout
image

hadicheaito1:
When a user signs in for the first time they should always have a default room (assuming they have the create room role permission), i guess this is for the case that they delete all their rooms only?

Visually I like the full width, it's also consistent with the other tabs. Not a strong opinion on either tho

scouillard:
> When a user signs in for the first time they should always have a default room (assuming they have the create room role permission), i guess this is for the case that they delete all their rooms only?
> 
> Visually I like the full width, it's also consistent with the other tabs. Not a strong opinion on either tho

I wasn't sure - didn't we got rid of the Home Room?

It doesn't seem to create a Home Room when I make an account, and I don't see it in the issues/cards.

But yeah indeed, either way it is proper display if there is no room, maybe change the wording a bit.

farhatahmad (approved):

farhatahmad:
Based on this PR: https://github.com/bigbluebutton/greenlight/pull/4217/files it was only done for local accounts - not external

PR #4501, opened 2023-01-12
Rework Homepage Layout

scouillard:
## Description

Added some icons/descriptions on what Greenlight/BBB does.
Texts are to be improved.

I think we should blur the line between "BBB" and "Greenlight" for the descriptions on the homepage as most of GL users don't make the distinction.

image


scouillard:
The footer is still broken. If you play with the browser width, it will overlay or disappear. Should be fixed in another PR.

farhatahmad (approved):

PR #4503, opened 2023-01-12
Remove Brand Color on Input hover and focus

scouillard:
## Description

Removed the hover and focus effect on the form inputs that would change the border color to brand color.
Re-added the default bootstrap box-shadow but changed the width so it's a bit smaller.

It was doing some free games with the SearchBar and the Select fields and it can be really ugly/weird depending in the brand color.
Having the default bootstrap blue box-shadow on input select is widely accepted and it was (my) bad idea to have it another way (the effect is actually a box-shadow, not a border-color).

image

### DIFF
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -120,6 +120,10 @@ input.search-bar {
   @include media-breakpoint-down(sm) {
     min-width: $button-min-width;
   }
+
+  &:focus {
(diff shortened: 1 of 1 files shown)


farhatahmad:
Hmm I'm not sure I agree with this. v2 had the brand colored inputs as well and I find them nicer when it's 1 uniform color

scouillard:
Right way to do would be to add the box-shadow effect instead of changing the border color.

PR #4508, opened 2023-01-13
Fixing room limit input glitch

hadicheaito1:
## Description

- Fixing room limit input glitch when validation occurs

### DIFF
--- a/app/assets/stylesheets/admin_panel.scss
+++ b/app/assets/stylesheets/admin_panel.scss
@@ -162,6 +162,10 @@
   width: 100px;
 }
 
+.invalid-feedback{
+  width: 200px;
+}
+
 .role-badge {
   background-color: whitesmoke !important;
   height: 24px;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/admin_panel.scss
```diff
@@ -162,6 +162,10 @@
   width: 100px;
 }
 
+.invalid-feedback{
+  width: 200px;
+}
```
farhatahmad, line 167: This will apply to all validation errors on all the admin pages - is that what you want to achieve?

hadicheaito1: Yeah I know, I tried avoiding it but it's the only thing I could come up with right now. It's fine for now since it won't affect anything else since there are no other yup validations on the admin pages

KH-Amir-TN: This affecting the entire application,
image
image

PR #4516, opened 2023-01-16
Production improvements

farhatahmad:


Dockerfile
```diff
@@ -39,8 +39,7 @@ RUN apk add --no-cache \
     && ( echo 'install: --no-document' ; echo 'update: --no-document' ) >>/etc/gemrc
 COPY . ./
 RUN bundle install -j4 \
-    && yarn install \
-    && ./node_modules/.bin/esbuild app/javascript/*.* --bundle --sourcemap --outdir=app/assets/builds
```
farhatahmad, line 43: Already done as part of `rake assets:precompile`

Dockerfile
```diff
@@ -51,6 +50,9 @@ ENV RAILS_SERVE_STATIC_FILES=${RAILS_SERVE_STATIC_FILES:-true}
 ARG PORT
 ENV PORT=${PORT:-3000}
 
+ARG NODE_ENV
+ENV NODE_ENV=${NODE_ENV:-production}
```
farhatahmad, line 54: Production mode :rocket:

package.json
```diff
@@ -39,8 +39,8 @@
     "yup": "^0.32.11"
   },
   "scripts": {
-    "build": "esbuild app/javascript/*.* --bundle --sourcemap --outdir=app/assets/builds --loader:.png=dataurl",
-    "build:css": "sass ./app/assets/stylesheets/application.bootstrap.scss ./app/assets/builds/application.css --no-source-map --load-path=node_modules"
+    "build": "esbuild app/javascript/*.* --bundle --sourcemap --outdir=app/assets/builds --loader:.png=dataurl --minify",
```
KH-Amir-TN, line 42: The `--minify` will cause esbuild to make a complete minification to the js bundle(s) and furthermore set the `process.env.NODE_END` to `production` which what we need on a production environment.
However this will cause the following issues for development:
1. Dev js assets will be minified (no practical debugging is possible).
2. Any js package that is conform with the standard will then run its production version which will not usually include any verbose errors or logging.

farhatahmad: Let me see if there's a way to run 2 different scripts based on env variable

Dockerfile
farhatahmad: It's what tells npm to run in production - meaning React will run in production mode

PR #4526, opened 2023-01-17
Added comments to all methods

farhatahmad:
## Description

Added basic descriptions and missing routes to all methods

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -12,7 +12,7 @@ Rails.application.routes.draw do
   # All the Api endpoints must be under /api/v1 and must have an extension .json.
   namespace :api do
     namespace :v1 do
-      resources :sessions, only: %i[index create] do
+      resources :sessions, only: %i[index create destroy] do
         collection do
           delete 'signout', to: 'sessions#destroy'
         end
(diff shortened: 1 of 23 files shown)


config/routes.rb
```diff
@@ -12,7 +12,7 @@
   # All the Api endpoints must be under /api/v1 and must have an extension .json.
   namespace :api do
     namespace :v1 do
-      resources :sessions, only: %i[index create] do
+      resources :sessions, only: %i[index create destroy] do
         collection do
           delete 'signout', to: 'sessions#destroy'
```
KH-Amir-TN, line 17: I believe that we need to remove the `DELETE /sessions/signout.json` collection route L16-L18 and update the `useDeleteSession` hook to use the `DELETE /sessions.json` API instead.
Because this way the `SessionsController#destroy` is not accessed through the destroy resource API.

KH-Amir-TN (commented):
I believe this PR provides a basic documentation to controllers action methods only?
We may want to further document the controllers private methods especially the ones that maintainers should be advertised on how to use (like `UsersController#valid_invite_token`) and also the rest of the backend resources (models, services, ...).
However, that can be done in subsequent PRs,

config/routes.rb
farhatahmad: That was my first though as well - but then I realized `destroy` requires an `id` which doesn't really exist in this context. Looks like I forgot to remove this

hadicheaito1 (approved):

config/routes.rb
KH-Amir-TN: Gotcha,

PR #4537, opened 2023-01-18
Fix the Logo jumping on page loading

scouillard:
## Description

- The work is in `Logo.jsx`
- Put the Logo in a wrapper with a fixed width and height. If the Logo is loading, only the wrapper is returned. This prevents the Logo "jumping" on load.
- Made "regular" logo the default instead of "small" for the `Logo` component which caused a bunch of changed files
- Size down the "regular" logo from `max 300px by 150px` to `max 250px by 125px`

### DIFF
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -398,14 +398,14 @@ input[type='checkbox'] {
   max-height: $header-height;
 }
 
-.medium-logo {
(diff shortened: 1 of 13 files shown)


scouillard:
There is some discrimination against Tall/Vertical logos. 
Could make the wrapper 250 by 250 instead of 250 by 125 to allow taller logos but it would need some work on the `vertical-center` class.

farhatahmad (approved):

PR #4546, opened 2023-01-19
Fix Logo false class console warning

scouillard:
## Description

Would return className:"false" which would render a warning in the console


scouillard:
Logo component is kind of dumb and primitive but doesn't really need refactoring for now since we are using it only in the header and on top of the forms/cards

PR #4616, opened 2023-01-20
Removed email from being a permitted parameter in the users controller,

alihadi-mazeh:
changed the email field on the user update form on the front end to being read-only. Completely disabling email changing for all types of users.

Disabled the ability to change email
## Description

- Removed the email parameter from being permitted when doing an update in the users controller
- Changed the email field in the User Update Form to being read-only

image


hadicheaito1:
RSpec @alihadi-mazeh

farhatahmad (approved):

farhatahmad (approved):

Issue #4376, opened 2022-12-29
(v3) cannot remove custom brand image


hiroshisuga:
**Describe the bug**
I don't find a way to remove the brand image that I uploaded.

**To Reproduce**
1. upload a image file as a brand image
2. you cannot remove it to revert to the original BBB logo

**Expected behavior**
One can remove the uploaded brand image, which was possible on v2 by putting a blank text. An addition of "Delete uploaded brand image" button would be ideal.

**Actual behavior**
Cannot remove, as far as I tried...

hadicheaito1:
Agreed, will add this ability

PR #4620, opened 2023-01-21
Adding ability to remove custom branding image

hadicheaito1:
## Description

- Added ability to remove the custom branding image and make it return to the bbb_logo.png default
- Solves #4376 
## Testing Steps

- Add custom image
- Click on remove custom branding image button
- Make sure it goes back to default

image

### DIFF
--- a/app/controllers/api/v1/admin/site_settings_controller.rb
+++ b/app/controllers/api/v1/admin/site_settings_controller.rb
@@ -29,17 +29,17 @@ module Api
                                     )
           return render_error status: :not_found unless site_setting
 
-          # Case where user is removing the custom branding image
-          if params[:site_setting][:value].nil?
-            site_setting.image.purge
-            update = true
-          else
-            update = if params[:name] == 'BrandingImage'
-                       site_setting.image.attach params[:site_setting][:value]
+          update = if params[:name] == 'BrandingImage'
+                     # Case where user is removing the custom branding image
+                     if params[:site_setting][:value].nil?
+                       site_setting.image.purge
+                       return true
                      else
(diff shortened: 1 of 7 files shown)


hadicheaito1:
I could keep it like this, or create a separate endpoint. Let me know what you guys think.

app/controllers/api/v1/admin/site_settings_controller.rb
```diff
@@ -30,7 +30,13 @@ def update
           return render_error status: :not_found unless site_setting
 
           update = if params[:name] == 'BrandingImage'
-                     site_setting.image.attach params[:site_setting][:value]
+                     # Case where user is removing the custom branding image
+                     if params[:site_setting][:value].nil?
+                       site_setting.image.purge
```
farhatahmad, line 35: Can you try replacing this with `site_setting.image.attach ActionController::Base.helpers.image_path('bbb_logo.png')`

hadicheaito1: Yeah this is what I initially tried to do but it wasn't working due to the image_path not getting the correct path. I tried other helper methods to get the correct path, but the only alternative to what I did that worked was this:
`site_setting.image.attach(io: File.open('/home/blindsidenetworks/greenlight/app/assets/images/bbb_logo.png'), filename: 'bbb_logo.png')`, which I'm not sure is better. Let me know what you think, also let me know what you think about creating it's own endpoint for this if I keep the `purge`

farhatahmad: Yeah I think creating a separate endpoint then would make more sense. The logic here is becoming a bit much

farhatahmad (approved):

PR #4655, opened 2023-01-26
No search results page

hadicheaito1:
## Description

- Adding a no search result page to Server rooms, Manage Users (Active, Pending, Banned, Invited), and Roles
- Refactored existing `NoRoomsFound` and `NoRecordingsFound` components into one component that could be used for all called `NoSearchResults`
## Testing Steps

- Go to those pages and search for something that doesn't exist and make sure no results page shows nicely


app/assets/locales/en.json
```diff
@@ -23,6 +23,7 @@
   "are_you_sure": "Are you sure?",
   "return_home": "Return Home",
   "created_at": "Created at",
+  "search_not_found": "No {{ name }} found",
```
farhatahmad, line 26: It's normally not a good idea to pass in the text here because this becomes unlocalizable. In english it might be "no rooms found" but in other languages it might be something like "not found room" (just and example)

farhatahmad: We're better off precreating all of the different options

app/javascript/components/admin/manage_users/InvitedUsers.jsx
```diff
@@ -6,53 +6,66 @@ import { CheckIcon, XMarkIcon } from '@heroicons/react/24/solid';
...
+                      </tr>
+                    ))
+                  )
+                  : (
+                    <tr>
+                      <td className="fw-bold">
+                        { t('user.no_user_found') }
+                      </td>
+                    </tr>
+                  )}
```
scouillard, line 53: We should replace this for a "EmptyResource" card eventually. Like we have in Rooms, Recordings, SharedAccess, Presentation, when there is no resource yet.

hadicheaito1: Yeah i'll do that in another upcoming PR

farhatahmad (approved):

PR #4677, opened 2023-01-30
Date Time Localization

alihadi-mazeh:
## Description

- Installed NPM package strftime to allow for date_localization
- Changes were done in:
  - ManageUserRow.jsx
  - ServerRoomRow.jsx
  - RecordingRow.jsx
  - RoomCard.jsx
  - Room.jsx
- Removed corresponding strftimes in serializers to allow for change
- updated the en.json files for the localization

image
image
image
image

### DIFF
--- a/app/assets/locales/en.json
+++ b/app/assets/locales/en.json
@@ -23,6 +23,8 @@
   "are_you_sure": "Are you sure?",
   "return_home": "Return Home",
   "created_at": "Created at",
+  "date_time": "%B %e, %Y %l:%M%P",
+  "day_date_time": "%A %B %e, %Y %l:%M%P",
   "homepage": {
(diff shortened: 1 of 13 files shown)


alihadi-mazeh:
Just an FYI: eslint is complaining on an error: `error  Unexpected require()  global-require`. Should I just do: `/* eslint-disable global-require */` above the require line and re-enable for everything below it like this:
```
/* eslint-disable global-require */
require ...
/* eslint-enable global-require */
```

scouillard:
I think [javascripts toLocaleDateString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString) could be used for the date localization.

Also, see https://stackoverflow.com/questions/3552461/how-do-i-format-a-date-in-javascript

Issue #4208, opened 2022-11-24
V2/V3: Migration should tell why it could not migrate an object


debuglevel:
Right now, the migration script just tells me that something did not work:
```
Unable to migrate Role:
  ID: 6
  Name: Freischalter

Roles migration complete.
In case of an error please retry the process to resolve.
```

Greenlight 3 logs do not tell much either:

```
I, [2022-11-24T12:21:55.490897 #124]  INFO -- : [194c7872-6192-4a3b-bc5b-e19e54eab06c] Started POST "/api/v1/migrations/roles.json" for 168.119.44.222 at 2022-11-24 12:21:55 +0100
I, [2022-11-24T12:21:55.491772 #124]  INFO -- : [194c7872-6192-4a3b-bc5b-e19e54eab06c] Processing by Api::V1::Migrations::ExternalController#create_role as JSON
I, [2022-11-24T12:21:55.492114 #124]  INFO -- : [194c7872-6192-4a3b-bc5b-e19e54eab06c]   Parameters: {"v2"=>{"encrypted_params"=>"[FILTERED]"}, "external"=>{"v2"=>{"encrypted_params"=>"[FILTERED]"}}}
I, [2022-11-24T12:21:55.492616 #124]  INFO -- : [194c7872-6192-4a3b-bc5b-e19e54eab06c]   Rendered components/index.html.erb within layouts/application (Duration: 0.0ms | Allocations: 4)
I, [2022-11-24T12:21:55.493222 #124]  INFO -- : [194c7872-6192-4a3b-bc5b-e19e54eab06c]   Rendered layout layouts/application.html.erb (Duration: 0.7ms | Allocations: 388)
I, [2022-11-24T12:21:55.493328 #124]  INFO -- : [194c7872-6192-4a3b-bc5b-e19e54eab06c] Filter chain halted as :ensure_valid_request rendered or redirected
I, [2022-11-24T12:21:55.493397 #124]  INFO -- : [194c7872-6192-4a3b-bc5b-e19e54eab06c] Completed 200 OK in 1ms (Views: 1.0ms | ActiveRecord: 0.0ms | Allocations: 609)
```

PR #4678, opened 2023-01-30
Add Errors Logs to Migration (v2)

scouillard:
## Description

To solve https://github.com/bigbluebutton/greenlight/issues/4208

V3 will return the ActiveRecord `resource.errors` if a `resource.save` does not work or a string `resource does not exist` if a depending resource has not been properly migrated.

### DIFF
--- a/lib/tasks/migrations/migrations.rake
+++ b/lib/tasks/migrations/migrations.rake
@@ -259,6 +259,8 @@ namespace :migrations do
       "invite"
     when "2"
       "approval"
+    else
+      "open"
     end
   end
 end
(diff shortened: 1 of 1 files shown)


lib/tasks/migrations/migrations.rake
```diff
@@ -255,8 +259,6 @@ namespace :migrations do
       "invite"
     when "2"
       "approval"
-    else
-      "open"
```
farhatahmad, line 259: Why are we dropping this?

scouillard: Good catch, this is a copy/paste mistake.

farhatahmad (approved):

PR #4679, opened 2023-01-30
Add Errors Logs to Migration (v3)

scouillard:
## Description

To solve https://github.com/bigbluebutton/greenlight/issues/4208

V3 will return the ActiveRecord `resource.errors` if a `resource.save` does not work or a string `resource does not exist` if a depending resource has not been properly migrated.

### DIFF
--- a/app/controllers/api/v1/migrations/external_controller.rb
+++ b/app/controllers/api/v1/migrations/external_controller.rb
@@ -33,7 +33,7 @@ module Api
 
           role = Role.new(name: role_hash[:name], provider: 'greenlight')
 
-          return render_error(status: :bad_request, errors: role.errors.to_a) unless role.save!
+          return render_error(status: :bad_request, errors: role&.errors&.to_a) unless role.save
 
           # Returns unless the Role has a RolePermission that differs from V3 default RolePermissions values
           return render_data status: :created unless role_hash[:role_permissions].any?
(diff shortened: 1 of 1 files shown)


app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -33,7 +33,7 @@ def create_role
 
           role = Role.new(name: role_hash[:name], provider: 'greenlight')
 
-          return render_error status: :bad_request unless role.save
+          return render_error(status: :bad_request, errors: role&.errors&.to_a) unless role.save
```
KH-Amir-TN, line 36: Is safe navigation operator  `&.` required here?
This logic should be run after ensuring that the `role` exists.

app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -93,7 +94,7 @@ def create_room
 
           user = User.find_by(email: room_hash[:owner_email], provider: 'greenlight')
 
-          return render_error status: :bad_request unless user
+          return render_error(status: :bad_request, errors: 'The room owner does not exist.') unless user
```
KH-Amir-TN, line 97: We can shorten that to an identifier like `RecordNotFound` or even send a status 404 and handle that on v2.

farhatahmad (approved):

app/controllers/api/v1/migrations/external_controller.rb
KH-Amir-TN: Is it adequate to send raw Activerecord errors?
It can be excessive,

The Activerecord errors can be logged on v3 and they are useful to maintainers not to admins,
There's nothing an admin can do when facing a `Password can't be blank` and that indicates that maintainers made a bug,
Admins can access v3 logs when opening an issue,
I think they need some finer grained specificity for errors not the generic case they have now,

So we can check for unfound records case, invalid data (wrong encryption) and maybe an indication that something went wrong if both cases didn't match and error happened,

PR #4697, opened 2023-01-31
Improve User Dropdown

scouillard:
## Description

Small improvements to Navbar User Dropdown

- Remove the user name on the Navbar dropdown
- The user name does not bring much value - the user should not have many GL accounts.
- Looks much cleaner IMO

Tried to refactor as the Mobile/Desktop Navbar logic is somewhat redundant and complicated without much success. The problem is that the Navbar does not play nice with Dropdown, which is our only Navbar element.

image
image

### DIFF
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -76,9 +76,13 @@ body {
   cursor: pointer;
   outline: inherit;
   min-width: auto;
+  background-color: transparent !important;
 
   &:focus {
     box-shadow: none;
+  }
+  &:active {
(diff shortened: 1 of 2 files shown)


hadicheaito1 (approved):
I like it :+1:

farhatahmad:
Honestly, still not quite sure about this. We have all of that space on top, I think displaying the user name is nice. i normally see this format if there are other options in the header and it's become too crowded

scouillard:
> Honestly, still not quite sure about this. We have all of that space on top, I think displaying the user name is nice. i normally see this format if there are other options in the header and it's become too crowded

I'll give it another shot with the Name.

PR #4716, opened 2023-02-01
Fix Moderator Access Code

scouillard:
## Description

Solution for the case when:
1. AnyoneJoinAsModerator is enabled
2. The meeting requires a Moderator Access Code - but does not require a Viewer Access Code.

There could be confusion as the Moderator Access Code is labelled as (Optional).

Solution is to add AnyoneJoinAsModerator as a Room public information and remove the (Optional) from the string in that case.


farhatahmad (dismissed):

farhatahmad:
@scouillard Rspec is failing

farhatahmad (approved):

PR #4717, opened 2023-02-01
Fixed multiple issues with locale

farhatahmad:
## Description

- Added proper fallback logic
- Fixed edit user showing the wrong language
- Fixed non-existent language being set for users on signup

### DIFF
--- a/app/javascript/i18n.jsx
+++ b/app/javascript/i18n.jsx
@@ -6,15 +6,17 @@ i18next
   .use(initReactI18next)
   .use(HttpApi)
   .init({
-    fallbackLng: 'en',
     backend: {
       loadPath: '/api/v1/locales/{{lng}}.json',
-      requestOptions: {
-        cache: 'no-store', // TODO - samuel: i18n will sometime use the cache translation
-      },
     },
-    interpolation: {
-      escapeValue: false,
+    load: 'currentOnly',
+    fallbackLng: (locale) => {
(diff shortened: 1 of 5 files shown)


app/javascript/i18n.jsx
```diff
@@ -6,15 +6,17 @@ i18next
   .use(initReactI18next)
   .use(HttpApi)
   .init({
-    fallbackLng: 'en',
     backend: {
       loadPath: '/api/v1/locales/{{lng}}.json',
-      requestOptions: {
-        cache: 'no-store', // TODO - samuel: i18n will sometime use the cache translation
-      },
```
farhatahmad, line 14: I think we can remove this now. It seems to be caching correctly

app/javascript/i18n.jsx
```diff
@@ -6,15 +6,17 @@ i18next
...
-    interpolation: {
-      escapeValue: false,
+    load: 'currentOnly',
+    fallbackLng: (locale) => {
+      const fallbacks = [];
+      if (locale?.indexOf('-') > -1) {
+        fallbacks.push(locale.split('-')[0]);
+      }
+      fallbacks.push('en');
+      return fallbacks;
```
farhatahmad, line 19: returns an array of fallbacks ordered by priority.
If langauge is 'fa-IR', the language will be decided by this array ['fa-IR', 'fa', 'en']

PR #4719, opened 2023-02-01
V3 Migrations: Minor fixes and optimizations to v2 rake tasks.

KH-Amir-TN:
## Description

Minor fixes and optimized to v2 rake tasks.

### DIFF
--- a/lib/tasks/migrations/migrations.rake
+++ b/lib/tasks/migrations/migrations.rake
@@ -15,21 +15,22 @@ namespace :migrations do
 
     Role.unscoped
         .select(:id, :name)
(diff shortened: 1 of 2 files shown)


lib/tasks/migrations/migrations.rake
```diff
@@ -15,21 +16,22 @@ namespace :migrations do
 
     Role.unscoped
         .select(:id, :name)
+        .includes(:role_permissions)
```
KH-Amir-TN, line 19: N+1

farhatahmad (approved):

PR #4761, opened 2023-02-06
Tasks: Initial `user:create` task.

KH-Amir-TN:
## Description

- [x] Initial skeleton.
- [x] Create a user with some basic defaults and fallbacks in place. 
- [x] Catch inexistent role case.
- [x] Basic errors.
- [x] Support streams redirection and exit code.
- [x] Logs must be order independent. 
- [ ] Improve errors degree of information.
- [ ] Improve task resiliency (cast/map user values to reduce error surface).
- [ ] Add color codes to different logging levels.
- [ ] Enable quite mode.


.rubocop.yml
```diff
@@ -81,3 +81,11 @@ Metrics/CyclomaticComplexity:
 Metrics/PerceivedComplexity:
   Max: 13
 
+Rails/Exit:
+  Exclude:
+    - 'lib/tasks/task_helpers.rb'
```
KH-Amir-TN, line 86: The file shouldn't be a rake task (otherwise, it will be picked by the Rakefile) but instead a regular ruby file.
Rubocop then lint it as if it was regular rails code therefore complaining about the usage of `exit` and direct usage of standard streams.

lib/tasks/user.rake
```diff
@@ -0,0 +1,50 @@
...
+    display_user_errors(user:) unless user.save
+
+    success 'User account was created successfully!'
+    info "  Name: #{user.name}"
+    info "  Email: #{user.email}"
+    info "  Password: #{user.password}"
+    info "  Role: #{user.role.name}"
+    info "  Verified: #{user.verified}"
+    info "  Status: #{user.status}"
+    info "  Language: #{user.language}"
```
farhatahmad, line 30: I would drop these from the output

lib/tasks/user.rake
```diff
@@ -0,0 +1,50 @@
...
+
+    exit 0
+  end
+
+  private
+
+  def check_role!(user:)
+    role_name = user[:role]
+    user[:role] = Role.find_by(name: role_name, provider: 'greenlight')
+    return if user[:role]
```
farhatahmad, line 40: Can we simplify this to `return if Role.exists?(...)`

KH-Amir-TN: We are setting the role as `user[:role]` if found so that query is used for both setting the role and also to enable us to catch inexistent roles.

The `Role.exists` will catch the latter case but will require us to make an additional find operation to create the user.

Is that what you're suggesting Ahmad?

lib/tasks/user.rake
KH-Amir-TN: Good catch!

lib/tasks/user.rake
```diff
@@ -0,0 +1,50 @@
...
+  task :create, %i[name email password role] => :environment do |_task, args|
+    # Default values.
+    user = {
+      provider: 'greenlight',
+      verified: true,
+      status: :active,
+      language: I18n.default_locale
+    }.merge(args)
+
+    user[:role] = SettingGetter.new(setting_name: 'DefaultRole', provider: 'greenlight').call if user[:role].blank?
```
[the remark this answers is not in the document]
KH-Amir-TN: Will do,

lib/tasks/user.rake
farhatahmad: Ah I see what you're doing. No this is fine

farhatahmad (approved):

PR #4770, opened 2023-02-07
Fix start script not accepting non-localhost params

farhatahmad:
## Description

The host and port variables were hardcoded to localhost, meaning external PG and RD was resulting in never starting. 

As far as I'm aware, this is the cleanest way to parse URLs in bash without any external libraries. We should revisit this to simplify it as some point

### DIFF
--- a/bin/start
+++ b/bin/start
@@ -1,32 +1,42 @@
 #!/usr/bin/env bash
 
+if [ -f .env ]; then
+  export $(echo $(cat .env | sed 's/#.*//g'| xargs) | envsubst)
+fi
+
 PORT="${PORT:=3000}"
 
 # Parse Rails DATABASE and REDIS urls to get host and port
-pgProtocol="$(echo $DATABASE_URL | grep :// | sed -e's,^\(.*://\).*,\1,g')" # extract the protocol
-pgUrl="${DATABASE_URL/$pgProtocol/}" # remove the protocol
-pgUser="$(echo $pgUrl | grep @ | cut -d@ -f1)" # extract the user (if any)
-pgHostport="$(echo ${pgUrl/$pgUser@/} | cut -d/ -f1)" # extract the host and port
-pgHost="$(echo $pgHostport | cut -d: -f1)" # by request host without port
-pgPort="$(echo $pgHostport | cut -d: -f2)" # by request - try to extract the port
+TXADDR=${DATABASE_URL/*:\/\/}
+TXADDR=${TXADDR/*@/}
+TXADDR=${TXADDR/\/*/}
+IFS=: TXADDR=($TXADDR) IFS=' '
+PGHOST=${TXADDR[0]}
+PGPORT=${TXADDR[1]:-5432}
 
-rdProtocol="$(echo $REDIS_URL | grep :// | sed -e's,^\(.*://\).*,\1,g')" # extract the protocol
-rdUrl="${REDIS_URL/$rdProtocol/}" # remove the protocol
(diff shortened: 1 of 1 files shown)


bin/start
```diff
@@ -1,22 +1,36 @@
 #!/usr/bin/env bash
 
+if [ -f .env ]; then
+  export $(echo $(cat .env | sed 's/#.*//g'| xargs) | envsubst)
```
KH-Amir-TN, line 4: Is this required?
 I don't think it is required for production workloads, the container itself will already have all ENV VARs on the `.env` file parsed and loaded by the composer.
This IMO will only add more delay for the start of the service,

bin/start
```diff
@@ -1,22 +1,36 @@
...
-RSHOST=redis
-RSPORT=6379
 
-echo "Greenlight-v3 starting on port: $PORT"
+# Parse Rails DATABASE and REDIS urls to get host and port
+pgProtocol="$(echo $DATABASE_URL | grep :// | sed -e's,^\(.*://\).*,\1,g')" # extract the protocol
+pgUrl="$(echo ${DATABASE_URL/$pgProtocol/})" # remove the protocol
+pgUser="$(echo $pgUrl | grep @ | cut -d@ -f1)" # extract the user (if any)
+pgHostport="$(echo ${pgUrl/$pgUser@/} | cut -d/ -f1)" # extract the host and port
+pgHost="$(echo $pgHostport | sed -e 's,:.*,,g')" # by request host without port
```
KH-Amir-TN, line 14: I think all of this logic is optimistic,
Therefore, this can be simplified to something like `pgHost="$(echo $pgHostport | cut -d: -f1)" # by request host without port`

bin/start
```diff
@@ -1,22 +1,36 @@
...
+pgPort="$(echo $pgHostport | sed -e 's,^.*:,:,g' -e 's,.*:\([0-9]*\).*,\1,g' -e 's,[^0-9],,g')" # by request - try to extract the port
```
KH-Amir-TN, line 15:
```bash
pgPort="$(echo $pgHostport | cut -d: -f2)" # by request - try to extract the port
```

bin/start
```diff
@@ -1,22 +1,36 @@
...
 PORT="${PORT:=3000}"
-PGHOST=postgres
-PGPORT=5432
-RSHOST=redis
-RSPORT=6379
 
-echo "Greenlight-v3 starting on port: $PORT"
+# Parse Rails DATABASE and REDIS urls to get host and port
+pgProtocol="$(echo $DATABASE_URL | grep :// | sed -e's,^\(.*://\).*,\1,g')" # extract the protocol
+pgUrl="$(echo ${DATABASE_URL/$pgProtocol/})" # remove the protocol
```
KH-Amir-TN, line 11: I think using echo here isn't required since this is a bash operation,
`pgUrl=${DATABASE_URL/$pgProtocol/}`

bin/start
farhatahmad: Yep - just forgot to remove it

bin/start
```diff
@@ -1,13 +1,29 @@
...
+RDHOST=${TXADDR[0]}
+RDPORT=${TXADDR[1]:-6379}
 
 echo "Greenlight-v3 starting on port: $PORT"
 
+echo $PGHOST
+echo $PGPORT
+
+echo $RDHOST
+echo $RDPORT
```
KH-Amir-TN, line 26: Is this for debugging?
Otherwise we can have a better message like `Checking $PGHOST $PGPORT Postgres service `...

KH-Amir-TN (approved):
# 🔥 🔥 🔥

PR #4776, opened 2023-02-08
More localesssss

farhatahmad:

### DIFF
--- a/app/javascript/components/users/user/forms/UpdateUserForm.jsx
+++ b/app/javascript/components/users/user/forms/UpdateUserForm.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { Button, Stack } from 'react-bootstrap';
 import PropTypes from 'prop-types';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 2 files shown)


app/javascript/components/users/user/forms/UpdateUserForm.jsx
```diff
@@ -15,36 +15,52 @@ import useUpdateUserForm from '../../../../hooks/forms/users/user/useUpdateUserF
...
+    const language = user?.language;
+    const noDialect = language.substring(0, language.indexOf('-'));
+
+    if (locales?.[language] !== undefined) {
+      return language;
+    } if (noDialect !== '' && locales?.[noDialect] !== undefined) {
+      return noDialect;
+    }
+    return 'en';
+  }
```
farhatahmad, line 36: Shows the right language in the dropdown. Basically 
if users language is supported -> return that
elseif the users language is a dialect and we support the non-dialect version -> return that
fallback to english

alihadi-mazeh (approved):

farhatahmad:
@hiroshisuga This should be the final locales PR hopefully forever :)

PR #4778, opened 2023-02-08
Add Remove Shared Access feature

scouillard:
## Description

Users should be able to delete their own Shared Access

## Testing Steps

Some things to consider & test:

- A user can delete its own share access. In that case, the user is redirected to /rooms
- A user can delete its own share access AND be an admin with ManageUsers. In that case, the admin is not redirected to /rooms

image

### DIFF
--- a/app/controllers/api/v1/shared_accesses_controller.rb
+++ b/app/controllers/api/v1/shared_accesses_controller.rb
@@ -8,7 +8,7 @@ module Api
       before_action only: %i[create shareable_users] do
         ensure_authorized('ManageRooms', friendly_id: params[:friendly_id])
       end
-      before_action only: %i[show] do
+      before_action only: %i[show destroy] do
         ensure_authorized(%w[ManageRooms SharedRoom], friendly_id: params[:friendly_id])
       end
 
--- a/app/controllers/api/v1/shared_accesses_controller.rb
+++ b/app/controllers/api/v1/shared_accesses_controller.rb
@@ -5,10 +5,10 @@ module Api
     class SharedAccessesController < ApiController
       before_action :find_room
 
-      before_action only: %i[create shareable_users] do
+      before_action only: %i[create destroy shareable_users] do
         ensure_authorized('ManageRooms', friendly_id: params[:friendly_id])
       end
-      before_action only: %i[show destroy] do
+      before_action only: %i[show unshare_room] do
         ensure_authorized(%w[ManageRooms SharedRoom], friendly_id: params[:friendly_id])
       end
 
@@ -34,6 +34,15 @@ module Api
       # Unshares a room with the specified user
       def destroy
         SharedAccess.delete_by(user_id: params[:user_id], room_id: @room.id)
+
+        render_data status: :ok
+      end
+
+      # DELETE /api/v1/shared_accesses/friendly_id.json
+      # Similar to the destroy action
+      # Unshares a room that is shared with the current user as a the current user
+      def unshare_room
+        SharedAccess.delete_by(user_id: current_user.id, room_id: @room.id)
 
         render_data status: :ok
       end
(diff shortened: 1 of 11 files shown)


app/services/permissions_checker.rb
```diff
@@ -61,7 +62,8 @@ def authorize_manage_rooms
   def authorize_shared_room
     return false if @friendly_id.blank? && @record_id.blank?
 
-    @current_user.shared_rooms.exists?(friendly_id: @friendly_id) ||
+    SharedAccess.joins(:shared_room).find_by(user_id: @current_user.id, 'shared_room.friendly_id': @friendly_id).present? ||
```
farhatahmad, line 65: Why the change here?

farhatahmad:
Question - wouldn't this be much simpler if we just created another endpoint?

app/services/permissions_checker.rb
scouillard: Doesn't this solve the "delete another shared access" issue actually?
It checks if the user is trying to do an action on the SharedAccess specifically, instead of just checking if the room is a shared room of the user.

app/javascript/components/rooms/room/FeatureTabs.jsx
```diff
@@ -1,19 +1,23 @@
...
+import useRoom from '../../../hooks/queries/rooms/useRoom';
 
-export default function FeatureTabs({ shared }) {
+export default function FeatureTabs() {
   const { t } = useTranslation();
   const { isLoading, data: settings } = useSiteSetting(['PreuploadPresentation', 'ShareRooms']);
   const currentUser = useAuth();
 
+  const { friendlyId } = useParams();
+  const { data: room } = useRoom(friendlyId);
```
scouillard, line 19: This feels not optimised.
We are querying the Room inside every Feature tab - it could be passed down from the Room component.
Not sure how React Query works - maybe it is not too bad.
But it is either that or a (somewhat) major refactoring with placeholders etc.

app/javascript/components/rooms/room/FeatureTabs.jsx
```diff
@@ -27,17 +31,6 @@ export default function FeatureTabs({ shared }) {
...
-  // Returns only the Recording tab if the room is a Shared Room and the User does not have the ManageRoom permission
-  if (shared && !currentUser?.permissions?.ManageRooms) {
-    return (
-      <Tabs className="wide-white pt-4 mx-0" defaultActiveKey="recordings" unmountOnExit>
-        <Tab className="background-whitesmoke" eventKey="recordings" title={t('recording.recording')}>
-          <RoomRecordings />
-        </Tab>
-      </Tabs>
-    );
-  }
```
scouillard, line 39: This has never worked, related to the comment above.
Passed down props was not properly dealt with - it needs a placeholder/isLoading.
It was set by default as 'false' and would not re-render for some reason.

app/javascript/components/rooms/room/FeatureTabs.jsx
```diff
@@ -61,11 +54,3 @@ export default function FeatureTabs({ shared }) {
...
   );
 }
-
-FeatureTabs.defaultProps = {
-  shared: false,
-};
-
-FeatureTabs.propTypes = {
-  shared: PropTypes.bool,
-};
```
scouillard, line 71: Related to the other comments

farhatahmad (dismissed):

app/controllers/api/v1/shared_accesses_controller.rb
```diff
@@ -38,6 +38,21 @@ def destroy
...
+
+        render_data status: :ok
+      end
+
+      # GET /api/v1/shared_accesses/:friendly_id.json
+      # Returns a list of all of the room's shared users
+      def show
+        render_data data: @room.shared_users.search(params[:search]), serializer: SharedAccessSerializer, status: :ok
+      end
+
```
farhatahmad, line 55: Duplicated code?

farhatahmad:
@scouillard Failing rubocop

farhatahmad (approved):

PR #4785, opened 2023-02-10
Added configuration:check rake task that checks a configuration

farhatahmad:
## Description

This task makes the following checks:
1. Confirms all required env variables are set
2. Checks that the PG db is connected
3. Checks that Redis is connected
4. Checks that it can connect to BBB
5. (if enabled) Checks that the SMTP server can send an email

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -48,9 +48,9 @@ Rails.application.configure do
       user_name: ENV.fetch('SMTP_USERNAME', nil),
       password: ENV.fetch('SMTP_PASSWORD', nil),
       authentication: ENV.fetch('SMTP_AUTH', nil),
-      enable_starttls_auto: ENV.fetch('SMTP_STARTTLS_AUTO', true),
-      enable_starttls: ENV.fetch('SMTP_STARTTLS', false),
-      tls: ENV.fetch('SMTP_TLS', 'false') != 'false',
+      enable_starttls_auto: ActiveModel::Type::Boolean.new.cast(ENV.fetch('SMTP_STARTTLS_AUTO', 'true')),
+      enable_starttls: ActiveModel::Type::Boolean.new.cast(ENV.fetch('SMTP_STARTTLS', 'false')),
+      tls: ActiveModel::Type::Boolean.new.cast(ENV.fetch('SMTP_TLS', 'false')),
       openssl_verify_mode: ENV.fetch('SMTP_SSL_VERIFY', 'true') == 'false' ? OpenSSL::SSL::VERIFY_NONE : OpenSSL::SSL::VERIFY_PEER
     }
     config.action_mailer.default_options = {
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -58,9 +58,9 @@ Rails.application.configure do
       user_name: ENV.fetch('SMTP_USERNAME'),
       password: ENV.fetch('SMTP_PASSWORD'),
       authentication: ENV.fetch('SMTP_AUTH'),
-      enable_starttls_auto: ENV.fetch('SMTP_STARTTLS_AUTO', true),
-      enable_starttls: ENV.fetch('SMTP_STARTTLS', false),
-      tls: ENV.fetch('SMTP_TLS', 'false') != 'false',
+      enable_starttls_auto: ActiveModel::Type::Boolean.new.cast(ENV.fetch('SMTP_STARTTLS_AUTO', 'true')),
+      enable_starttls: ActiveModel::Type::Boolean.new.cast(ENV.fetch('SMTP_STARTTLS', 'false')),
+      tls: ActiveModel::Type::Boolean.new.cast(ENV.fetch('SMTP_TLS', 'false')),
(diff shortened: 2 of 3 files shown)


lib/tasks/configuration.rake
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+require_relative 'task_helpers'
+
+namespace :configuration do
+  desc 'Create an administrator account'
```
hadicheaito1, line 6: I think the `desc` needs to be changed

config/environments/development.rb
```diff
@@ -48,9 +48,9 @@
       user_name: ENV.fetch('SMTP_USERNAME', nil),
       password: ENV.fetch('SMTP_PASSWORD', nil),
       authentication: ENV.fetch('SMTP_AUTH', nil),
-      enable_starttls_auto: ENV.fetch('SMTP_STARTTLS_AUTO', true),
-      enable_starttls: ENV.fetch('SMTP_STARTTLS', false),
-      tls: ENV.fetch('SMTP_TLS', 'false') != 'false',
+      enable_starttls_auto: ActiveModel::Type::Boolean.new.cast(ENV.fetch('SMTP_STARTTLS_AUTO', 'true')),
```
KH-Amir-TN, line 51: Just a note, any none explicit none emptoy value of `'true'|'false'` will always be `true`,

lib/tasks/configuration.rake
```diff
@@ -0,0 +1,73 @@
...
+  failed("Error connecting to BigBlueButton server - #{e}")
+end
+
+def passed
+  success 'Passed'
+end
+
+def failed(msg)
+  err "Failed - #{msg}"
+  exit
```
KH-Amir-TN, line 72: err have `exit 1`
If this is intentional like for to increase resiliency and verbosity then `exit 1` may be more adequate,

lib/tasks/configuration.rake
```diff
@@ -0,0 +1,73 @@
...
+    info 'Checking connection to BigBlueButton:'
+    test_request(Rails.configuration.bigbluebutton_endpoint)
+    checksum = Digest::SHA1.hexdigest("getMeetings#{Rails.configuration.bigbluebutton_secret}")
+    test_request("#{Rails.configuration.bigbluebutton_endpoint}getMeetings?checksum=#{checksum}")
+    passed
+
+    if ENV['SMTP_SERVER'].present?
+      info 'Checking connection to SMTP Server'
+      begin
+        UserMailer.with(to: ENV.fetch('SMTP_SENDER_EMAIL', nil), subject: ENV.fetch('SMTP_SENDER_EMAIL', nil)).test_email.deliver_now
```
KH-Amir-TN, line 44: I think we're not enabling raising exceptions for failed emails delivery?

KH-Amir-TN: We may want to indicate on success to admins to check the inbox folder of the `SMTP_SENDER_EMAIl` address?

lib/tasks/configuration.rake
```diff
@@ -0,0 +1,73 @@
...
+    required_env_vars.each do |var|
+      failed("#{var} not set correctly") if ENV[var].blank?
+    end
+    passed
+
+    info 'Checking connection to Postgres Database:'
+    begin
+      ActiveRecord::Base.establish_connection # Establishes connection
+      ActiveRecord::Base.connection # Calls connection object
+      failed('Failed to connect to Database') unless ActiveRecord::Base.connected?
```
KH-Amir-TN, line 21: Just for clarity `failed` already have a `Failed - ` prefix so this will be `Failed - Failed to connect to Database`.
I think something like `Failed - Unable to connect to Database` may be better.

config/environments/development.rb
farhatahmad: Yep - expected

lib/tasks/configuration.rake
farhatahmad: It will be mentioned in the docs that about the email that will be sent to SMTP_SENDER_EMAIl

config/environments/production.rb
```diff
@@ -70,7 +70,7 @@
     config.action_mailer.perform_deliveries = false
   end
 
-  config.action_mailer.raise_delivery_errors = false
+  config.action_mailer.raise_delivery_errors = true
```
KH-Amir-TN, line 73: When `perform_deliveries` is set to `false` (when SMTP isn't used),
Does the mailer will keep throwing exceptions then?
If yes, then I think this will block us from using GL3 without SMTP configured.

lib/tasks/configuration.rake
farhatahmad: Yep

config/environments/production.rb
farhatahmad: Nope - it just skips over it:
```
Skipped delivery of mail 63ea81b08a2de_d502ba7c307e4@ahmad-computer.mail as `perform_deliveries` is false
```

hadicheaito1 (approved):

PR #4790, opened 2023-02-10
Update Frontend

scouillard:
## Description

- Update Errors, "CantCreateRoom" landing page and Emails layout.
- Add "Enter Meeting URL" feature on "CantCreateRoom" page.


app/javascript/components/rooms/CantCreateRoom.jsx
```diff
@@ -1,28 +1,40 @@
...
+        <div className="p-5 pb-4">
+          <h2><strong>{t('homepage.enter_meeting_url')}</strong></h2>
+          <h5>{t('homepage.enter_meeting_url_instruction')}</h5>
+        </div>
+        <Card.Footer className="bg-white p-0">
+          <div className="p-5 pt-4">
+            <Stack direction="horizontal" gap={3}>
+              <input name="meetingUrl" className="form-control" id="joinUrl" onChange={handleChange} />
+              <Button
+                onClick={() => navigate(`/${meetingUrl.toString().split('/').slice(-3).join('/')}`, { replace: true })}
```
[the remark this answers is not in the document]
scouillard: Updated to a REGEX that will parse a 4 x 3 keys code from the input

app/javascript/components/rooms/CantCreateRoom.jsx
```diff
@@ -1,28 +1,44 @@
...
+
+  const [meetingUrl, setMeetingUrl] = useState('');
+
+  const handleChange = (event) => {
+    setMeetingUrl(event.target.value);
+  };
+
+  // Parse the 12 keys friendly id from the input
+  const regex = /(\w{3}-\w{3}-\w{3}-\w{3})/;
+  const parsedUrl = regex.exec(meetingUrl);
```
farhatahmad, line 20: Based on the little I know, you should be using `meetingUrl.match(regex)`. `.exec` was meant to be used in an array or loop or something like that

app/javascript/components/rooms/CantCreateRoom.jsx
```diff
@@ -1,28 +1,44 @@
...
+// This page is shown if the user does NOT have the CreateRoom permission
 export default function CantCreateRoom() {
   const { t } = useTranslation();
+  const navigate = useNavigate();
+
+  const [meetingUrl, setMeetingUrl] = useState('');
+
+  const handleChange = (event) => {
+    setMeetingUrl(event.target.value);
+  };
```
farhatahmad, line 16: Why the state variable? Why not just take the sibling value onClick? I think we do that in a few places already

app/javascript/components/rooms/CantCreateRoom.jsx
scouillard: They seem to be the same. match() syntax is cleaner IMO

app/javascript/components/rooms/CantCreateRoom.jsx
scouillard: `handleChange` seems to be more in line with the React way to do things.

scouillard: I am also not sure what is meant by sibling value - isn't that more of a jQuery thing?

farhatahmad: https://github.com/bigbluebutton/greenlight/blob/v3/app/javascript/components/admin/site_settings/registration/Registration.jsx#L80

farhatahmad: Don't have a strong opinion so I'll leave it up to you

scouillard: I like how it's done with the useState hook

PR #4805, opened 2023-02-15
Deployment: Added `gl-install`

KH-Amir-TN:
## Description

Introducing `gl-install` a script to auto deploy and upgrade a standalone deployment of Greenlight v3.

- The following are a set of increments to meet a production standalone deployment of Greenlight v3 similar to `bbb-install`:
- [X] Deploy GL with no SSL (HTTP only).
- [X] HTTP traffic is proxied via nginx (nginx as a daemon not as a container).
- [X] Enable the provision of SSL certificates and extend the script accordingly (generating or renewing certs at this step isn't our concern).
- [X] Extend the script to auto generate Let's encrypt certificate and provision it (leveraging 3. logic), certs renewal still isn't our concern.
- [X] ADD optional Keycloak (new docker-compose.kc.yml and keycloak.nginx).
- [ ] Handle certs renewal.
- [ ] Move nginx to a container.
- [ ] Enable proxy mode (system should support being behind a front end reverse proxy).
- [ ] Use certbot as a container (certs issuing and renewal goes through the container).
	
=> Meeting all of the increments should provide a complete self sufficient and isolated standalone system that can be deployed almost on any environment with little requirements (supports dockers containers, have a public FQDN and open ingress web traffic).

## Testing Steps

After releasing:
 - Create a clean and publicly accessible Ubuntu 20.04 x64 instance opened for ingress web traffic (both TCP 80 and 443).
 - Have a public FQDN that points to the instance public IP.
 - Run the `gl-install.sh` script with the preferred options set, for more information kindly consult the help page by using the `-h` option.
 
 You can run the latest version of the script through `wget -qO- https://raw.githubusercontent.com/bigbluebutton/greenlight/v3/gl-install.sh | bash -s -- [OPTIONS]`
 OR through a container:
 `docker run --rm --entrypoint cat bigbluebutton/greenlight:v3 gl-install.sh | bash -s -- [OPTIONS]`

### DIFF
--- a/docker-compose.kc.yml
+++ b/docker-compose.kc.yml
@@ -1,0 +1,23 @@
+
+  keycloak:
+    command: start
+    image: quay.io/keycloak/keycloak:20.0
+    container_name: keycloak
+    restart: unless-stopped
+    environment:
+      - KEYCLOAK_ADMIN=admin
+      - KEYCLOAK_ADMIN_PASSWORD=
+      - KC_DB=postgres
+      - KC_DB_URL_HOST=postgres
+      - KC_DB_USERNAME=postgres
+      - KC_DB_URL_DATABASE=keycloakdb
+      - KC_DB_PASSWORD=
+      - KC_HOSTNAME_STRICT=false
+      - KC_HTTP_RELATIVE_PATH=/keycloak
+      - KC_PROXY=edge
+    ports:
+      - 127.0.0.1:5151:8080
+    logging:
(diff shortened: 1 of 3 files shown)


farhatahmad:
Going to merge as is for now - will review as I test

farhatahmad:
Problem 1: If your certificate generation fails, and you re-run it, it just skips over the cert stuff the next time and succeeds

PR #4828, opened 2023-02-22
Edit Recordings Processing Message

hadicheaito1:
- Changed the message for processing recording


scouillard:
I was thinking we could add a "three-dot" pulse effect with the message

https://codepen.io/nzbin/pen/GGrXbp

hadicheaito1:
> I was thinking we could add a "three-dot" pulse effect with the message
> 
> https://codepen.io/nzbin/pen/GGrXbp

Yeah this would be a nice touch. I'll add this

app/assets/stylesheets/recordings.scss
```diff
@@ -106,3 +106,76 @@
...
+  0% {
+    box-shadow: 10014px 0 0 -5px;
+  }
+  30% {
+    box-shadow: 10014px 0 0 2px;
+  }
+  60%, 100% {
+    box-shadow: 10014px 0 0 -5px;
+  }
+}
```
farhatahmad, line 181: This is a lot for a simple loading animation - I think we should either look into a library or scrap this concept completely

hadicheaito1: I feel a library would be "heavy" for this, especially if we just using it here.

I think it's a nice touch though, if you think it's too much we can scrap it all

scouillard: I think it's nice.
The color needs to be changed from `#9880ff` to `whitesmoke`, `silver` or even `brand-color`

farhatahmad: Yeah to me this is just a mess when it comes to maintaining. Libraries are easy to maintain because someone else does it for you. If we find an issue with this code, we have to manually go in and fix it, and then test it against a bunch of different scenarios

hadicheaito1: Yeah makes sense. I'm thinking of using react-bootstrap since we already have it, and to just use this and change it's colour.
https://react-bootstrap.github.io/components/spinners/#:~:text=Copy-,Grow%20Spinner,-%2D%20grow

Besides that there isn't a small library for just this

farhatahmad (approved):

PR #4887, opened 2023-02-26
Update ru.json

alex-smirnov-sh:
Updated the installed tags following the example of the app/assets/locales/en.json file


gauss-lvs-dev:
Hi Alex, all translation files are managed with transifex. I tried to update the German language file the same way as you did and the pull request got rejected.

https://www.transifex.com/bigbluebutton/greenlight-v3

PR #4898, opened 2023-02-27
Add Sync Server Recordings task

scouillard:
Add Rake Task to Sync the Server Recordings.

### DIFF
--- a/lib/tasks/server_recordings_sync.rake
+++ b/lib/tasks/server_recordings_sync.rake
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
-desc 'Recordings sync with BBB server'
+desc 'Server Recordings sync with BBB server'
 
-task recordings_sync: :environment do
+task server_recordings_sync: :environment do
   Recording.destroy_all
 
   Room.select(:id, :meeting_id).in_batches(of: 25) do |rooms|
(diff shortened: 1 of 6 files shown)


scouillard:
RecordingsSync (service) and ServerRecordingsSync (rake task) share similar logic.  But I think it is different enough to warrant the duplication.

lib/tasks/server_recordings_sync.rake
```diff
@@ -0,0 +1,32 @@
...
+
+# frozen_string_literal: true
+
+desc 'Server Recordings sync with BBB server'
+
+task server_recordings_sync: :environment do
+  Recording.destroy_all
+
+  Room.select(:id, :meeting_id).in_batches(of: 25) do |rooms|
+    meeting_ids = rooms.pluck(:meeting_id)
```
farhatahmad, line 25: Can you not just do 
```
 Room.select(:meeting_id).in_batches(of: 25) do |rooms|
```

scouillard: No, all of the "batches" methods seem to return an ActiveRecord object.
Rails does not seem to provide a built-in way to [batch and pluck](https://gist.github.com/iamatypeofwalrus/f9f6a3049f63dd05d02c)

scouillard: There is a way to return the ids/index without firing an extra query https://stackoverflow.com/a/66885913

farhatahmad (approved):

PR #4900, opened 2023-02-27
Add recorded_at Column to Recordings

scouillard:
**Problem:**
Recording `created_at` timestamp is used to sort, display the meeting/recording date of creation.
But, when Syncing Recordings with the BBB server, the process deletes all the current recordings and re-creates them from the BBB response.
Hence, upon syncing, the `created_at` will be the date of the sync instead of the date of the meeting.

Solution:
This PR attempts to solve the issue by creating a new `meeting_date` column on Recordings. The `meeting_date` is initiated from BBB `startTime` response.

Note: Upon migrating the new column, the users will have to sync  the recordings to populate the `meeting_date` column.

### DIFF
--- a/db/data/20230228193705_populate_recorded_at_recording.rb
+++ b/db/data/20230228193705_populate_recorded_at_recording.rb
@@ -20,6 +20,7 @@ class PopulateRecordedAtRecording < ActiveRecord::Migration[7.0]
   def up
     # rubocop:disable Rails/SkipsModelValidations
     Recording.update_all('recorded_at = created_at')
+    # rubocop:enable Rails/SkipsModelValidations
   end
 
   def down
--- a/db/data/20230228193705_populate_recorded_at_recording.rb
+++ b/db/data/20230228193705_populate_recorded_at_recording.rb
@@ -16,26 +16,13 @@
 
 # frozen_string_literal: true
 
-class Recording < ApplicationRecord
-  belongs_to :room
-  has_one :user, through: :room
-  has_many :formats, dependent: :destroy
+class PopulateRecordedAtRecording < ActiveRecord::Migration[7.0]
+  def up
+    # rubocop:disable Rails/SkipsModelValidations
+    Recording.update_all('recorded_at = created_at')
+  end
 
-  validates :name, presence: true
-  validates :record_id, presence: true
-  validates :visibility, presence: true
-  validates :length, presence: true
-  validates :participants, presence: true
-  validates :recorded_at, presence: true
-
-  scope :with_provider, ->(current_provider) { where(user: { provider: current_provider }) }
-
-  def self.search(input)
-    if input
-      return joins(:formats).where('recordings.name ILIKE :input OR recordings.visibility ILIKE :input OR formats.recording_type ILIKE :input',
-                                   input: "%#{input}%").includes(:formats)
-    end
-
-    all.includes(:formats)
+  def down
+    raise ActiveRecord::IrreversibleMigration
   end
 end
--- a/db/migrate/20230227213252_add_recorded_at_to_recordings.rb
+++ b/db/migrate/20230227213252_add_recorded_at_to_recordings.rb
@@ -18,7 +18,6 @@
 
 class AddRecordedAtToRecordings < ActiveRecord::Migration[7.0]
   def change
-    # rubocop:disable Rails/NotNullColumn
-    add_column :recordings, :recorded_at, :datetime, null: false
+    add_column :recordings, :recorded_at, :datetime
   end
 end
--- a/db/migrate/20230227213252_add_recorded_at_to_recordings.rb
+++ b/db/migrate/20230227213252_add_recorded_at_to_recordings.rb
@@ -16,8 +16,9 @@
 
 # frozen_string_literal: true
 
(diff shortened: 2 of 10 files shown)


scouillard:
I think adding a column to recordings is a better solution than overwriting created_at.
Both solution will cause some issues but, from what I have seen so far, overwriting the timestamps seems kind of hacky.

db/schema.rb
```diff
@@ -105,6 +89,7 @@
     t.integer "length", null: false
     t.integer "participants", null: false
     t.boolean "protectable"
+    t.datetime "meeting_date"
```
farhatahmad, line 92: I think I would call this `recorded_at`

farhatahmad:
This will break the setup for existing users - we'll need a data migration to accompany it

db/migrate/20230227213252_add_recorded_at_to_recordings.rb
```diff
@@ -0,0 +1,24 @@
...
+# PARTICULAR PURPOSE. See the GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License along
+# with Greenlight; if not, see <http://www.gnu.org/licenses/>.
+
+# frozen_string_literal: true
+
+class AddRecordedAtToRecordings < ActiveRecord::Migration[7.0]
+  def change
+    # rubocop:disable Rails/NotNullColumn
```
scouillard, line 21: Is this a new cop? It complains that we are adding `null: false` without providing default value but it does not seem to be a problem elsewhere.

db/data/20230228193705_populate_recorded_at_recording.rb
```diff
@@ -0,0 +1,28 @@
...
+# PARTICULAR PURPOSE. See the GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License along
+# with Greenlight; if not, see <http://www.gnu.org/licenses/>.
+
+# frozen_string_literal: true
+
+class PopulateRecordedAtRecording < ActiveRecord::Migration[7.0]
+  def up
+    # rubocop:disable Rails/SkipsModelValidations
```
scouillard, line 21: update_all is much faster than looping through all Recordings but it skips validations, which I don't think are needed.

scouillard:
> This will break the setup for existing users - we'll need a data migration to accompany it

> This will break the setup for existing users - we'll need a data migration to accompany it

Running the migration will now set the new `recorded_at` to `created_at` value.

I would recommend existing users to run a Recording sync at the same time of the migration.

db/migrate/20230227213252_add_recorded_at_to_recordings.rb
farhatahmad: Hmm this is a tricky scenario. essentially, we're adding a new column that can't be null without specifying a default value, which is problematic. Maybe drop the null: false?

scouillard: It should not be a problem if users do `migrate:with_data`, which will populate the column with the `created_at` value. No?
I think `null: false` should stay.

farhatahmad: data_migrations are schema_migrations are independent of each other. When the schema_migration runs, it adds a new column `recorded_at` but doesn't know what the value of the column should be. Since the column can't be null, it's complaining

farhatahmad: If you wan't to keep the `null: false` we need to define a default value

scouillard: We could set the default value to `created_at` instead of doing it in the data_migration? And keep the `null: false`

farhatahmad: I don't believe you can reference another column in that way

hadicheaito1: > data_migrations are schema_migrations are independent of each other. When the schema_migration runs, it adds a new column `recorded_at` but doesn't know what the value of the column should be. Since the column can't be null, it's complaining

I don't get what the difference is here compared to any other table we created with a column that has null false? Wouldn't they all be this same situation?  Why is only this one complaining

farhatahmad: it's only for `add_column` not for `create_table`. There's only 3 other cases of `add_column` and they all have default values defined

farhatahmad (approved):

farhatahmad:
@scouillard rubocop

farhatahmad (approved):

kasraJ:
Unfortunately the problem with recoding date still exist. I thought after updatig from v3 to v3.0.1 this issue going to resolve. But we still see "Invalid date" in recordings.
image

I also checked and as you can see here the migration `Populate recorded at recording` is up.

image

shabeepk:
> Unfortunately, the problem with recoding date still exists. I thought after updating from v3 to v3.0.1 this issue going to resolve. But we still see `Invalid Date` in recordings.
> 
> image
> 
> I also checked and as you can see here the migration `Populate recorded at recording` is up.
> 
> image

Same thing here.

PR #4907, opened 2023-03-01
Add Provider

scouillard:
Create Provider table
Add Provider model
Rename Admins controller to Provider

### DIFF
--- a/app/assets/locales/hu.json
+++ b/app/assets/locales/hu.json
@@ -1,6 +1,7 @@
 {
   "start": "Indítás",
   "search": "Keresés",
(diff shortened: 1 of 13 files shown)


farhatahmad:
To avoid confusion, I'm tempted to call this something else. I think we can call it `tenants`

scouillard:
> To avoid confusion, I'm tempted to call this something else. I think we can call it `tenants`

I've renamed Providers to Tenants.
Why do you think Providers table is confusing?

farhatahmad (approved):

PR #4922, opened 2023-03-02
Add Provider to v3 migration endpoints

scouillard:
Add provider to the migration endpoints and specs


app/controllers/api/v1/migrations/external_controller.rb
```diff
@@ -47,14 +47,19 @@ class ExternalController < ApiController
         def create_role
           role_hash = role_params.to_h
 
-          return render_data status: :created if Role.exists? name: role_hash[:name], provider: 'greenlight'
```
scouillard, line 50: Had to rework the `exists?` logic a bit because this would return prematurely if the default Role from v2 would have different permissions than default permissions.

PR #4943, opened 2023-03-09
Add SuperAdmin provider bypass

scouillard:
# Context
A SuperAdmin should be able to login into any tenant and bypass all the provider checks.

# How the feature was tested:
Create initial resources

- Created a `super_admin_user`
- Created two tenants from the Admin Panel: `tenant_one` and `tenant_two`.

Setup multitenancy

- Assigned `tenant_one.greenlight.com` and `tenant_two.greenlight.com` to locahost via /etc/hosts
- Add `LOADBALANCER_ENDPOINT` and `LOADBALANCER_SECRET` var in .env

Create multitenancy resources

- Create `tenant_one_user` via `tenant_one` sign up form

Testing steps

- Successfully log in `tenant_one_user` into `tenant_one`
- Error when trying to log in `tenant_one_user` into `tenant_two`
- Successfully log in `super_admin_user` in both tenants

### DIFF
--- a/app/controllers/api/v1/sessions_controller.rb
+++ b/app/controllers/api/v1/sessions_controller.rb
@@ -35,8 +35,10 @@ module Api
         # TODO: Add proper error logging for non-verified token hcaptcha
         return render_error if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
-        user = User.find_by(email: session_params[:email], provider: [current_provider, 'bn'])
+        # Search for a user within the current provider and, if not found, search for a super admin within bn provider
+        user = find_user(session_params[:email], current_provider)
 
+        # Return an error if the user is not found
         return render_error if user.blank?
 
         # Will return an error if the user is NOT from the current provider and if the user is NOT a super admin
@@ -68,6 +70,13 @@ module Api
         params.require(:session).permit(:email, :password, :extend_session)
       end
 
+      def find_user(email, provider)
+        user = User.find_by(email:, provider:)
+        # If the user is not found in the current provider, search for a super admin with bn provider
+        user = User.find_by(email:, provider: 'bn') if user.blank?
+        user
+      end
+
       def sign_in(user)
         user.generate_session_token!(extended_session: session_params[:extend_session])
 
@@ -94,3 +103,4 @@ module Api
     end
   end
 end
+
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/sessions_controller.rb
```diff
@@ -32,13 +32,18 @@ def index
       # POST /api/v1/sessions
       # Signs a user in and updates the session cookie
       def create
-        user = User.find_by(email: session_params[:email], provider: current_provider)
-
         # TODO: Add proper error logging for non-verified token hcaptcha
         return render_error if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
+        user = User.find_by(email: session_params[:email])
```
scouillard, line 38: It is assumed that an email won't be duplicated in the database

farhatahmad: It is definitely possible that there will be the same email available in separate providers. This breaks that functionality

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -32,13 +32,18 @@ def index
       # POST /api/v1/sessions
       # Signs a user in and updates the session cookie
       def create
-        user = User.find_by(email: session_params[:email], provider: current_provider)
-
         # TODO: Add proper error logging for non-verified token hcaptcha
         return render_error if hcaptcha_enabled? && !verify_hcaptcha(response: params[:token])
 
+        user = User.find_by(email: session_params[:email], provider: [current_provider, 'bn'])
```
scouillard, line 38: This should solve the issue if a user has multiples accounts in different tenants.
But, this will default to a 'bn' account if the user email is associated with one - which seems reasonable.

scouillard: This is not behaving as expected upon more testing.

farhatahmad: Nope. We should prioritize current provider, and if not found, it should then search for bn

scouillard: Will now initially search for a user within the current_provider. 
If not found, it will search for a user with bn provider.

farhatahmad (approved):

Issue #4923, opened 2023-03-03
[v3] Invitation not translated


hiroshisuga:
Invitation screen is not completely translated, as far as I check it in Japanese.

![Animation1](https://user-images.githubusercontent.com/45039819/222681101-eaec2bfb-1e36-4895-9b78-d1a72b948525.gif)

Zer0CooL4ek:
Same on 3.0.2 and 3.0.3.1
List of proposals that remain untranslated: for example RU (although in transifex everything is translated)
Name
Enter your name
Please enter your full name. (red text (if no name is given))
_____
Enter the access code
_____
I acknowledge that this session may be recorded. This may include my voice and video if enabled.
_____
![photo_2023-04-11_12-00-49](https://user-images.githubusercontent.com/22744153/231110736-7471db92-e33f-403c-9c21-9e155e7480ee.jpg)

PhilippeAccorsi:
Same issue with french language.
Is it possible to participate to translate app ? (in weblate or something like that)

Zer0CooL4ek:
regarding the current problem, nothing has changed in release-3.0.4

scouillard:
Hi,

We are sorry - this was an oversight on our part.
We forgot to refactor the form into our new template that allows translation.
The support for the translation should be added for the upcoming next release but the strings for your languages will have to be added to [the GLv3 transifex](https://app.transifex.com/bigbluebutton/greenlight-v3/dashboard/).

Thank you for your patience.

farhatahmad:
https://github.com/bigbluebutton/greenlight/pull/5212

PR #4944, opened 2023-03-09
[v3] Add translation to room join portal

hiroshisuga:
This patch adds translation to the room join panel, closing #4923 .

Before:
222681101-eaec2bfb-1e36-4895-9b78-d1a72b948525

After:
スクリーンショット 2023-03-10 7 50 50


farhatahmad:
While this does the job, the actual solution to this a little bit more complex and requires some refactoring from the forms point of view. We'll get it done soon

hiroshisuga:
@farhatahmad , have you realised that the room join portal has not been fully translated yet? It would be great if you fix it soon (or later). Japanese students will not understand the recording consent in English;-)

PR #4946, opened 2023-03-10
Fix Super Admin Role select

scouillard:
**Problem:**
A Super Admin cannot see its Role in Profile because the roles are scoped by the site provider, and Super Admin is "bn"
image

**Solution:**
Hardcode the Role input to "Super Admin" and disable it.
image

I don't see a reason why a Super Admin would want to change its role.

Another solution would be to create a new endpoint to fetch the Roles if the user is a super admin.

**How To Test:**

- Create a Super Admin
- Go to Profile and verify that Role is set to Super Admin and that you can't change it
- Go to Manage Users, and try to edit another user's role
- Log into a regular admin user and make sure that your role is not set to Super Admin


farhatahmad:
I would just hide the role dropdown if the user is a superadmin - would simplify things greatly

farhatahmad:
@scouillard eslint

scouillard:
> This has changes from another branch - You need to rebase it on upstream master

Can it wait for #4943

farhatahmad (approved):

PR #5028, opened 2023-03-13
Allowed for + symbols in emails

alihadi-mazeh:


app/models/user.rb
```diff
@@ -42,7 +42,7 @@ class User < ApplicationRecord
                    length: { minimum: 2, maximum: 255 } # TODO: amir - Change into full_name or seperate first and last name.
 
   validates :email,
-            format: /\A[\w\-.]+@[\w\-.]+\.[a-z]+\z/i,
+            format: /\A[\w\-.+]+@[a-z\d-]+(\.[a-z\d-]+)*\.[a-z]+\z/i,
```
scouillard, line 45: Just adding the "+" in the first list seems to work:
\A[\w\-.+]+@[\w\-.]+\.[a-z]+\z

PR #5029, opened 2023-03-14
Add Reset Password Email Task

scouillard:
Upon migrating the users from V2 to V3, the user's passwords are reset.
This tasks aims to provide the system administrator an easy way to inform the users that their password have been reset.
SA will have to manually provide the `root_url` for now.

How To Test:

- Needs #5039 for proper email template
- Run the command below from the project dir. Pleas set your base url as an argument to the rake task.
`rake migration:reset_password_email\[<base url>]`
`rake migration:reset_password_email\[https://b1b3-24-203-33-70.ngrok.io]`
- The users that have been sent an email should appear in the console
image

### DIFF
--- a/lib/tasks/reset_password_email.rake
+++ b/lib/tasks/reset_password_email.rake
@@ -19,17 +19,17 @@
 require_relative 'task_helpers'
 
 namespace :migration do
-  include ActionView::Helpers::AssetUrlHelper
+  BATCH_SIZE = 500
 
   desc 'Send a reset password email to users'
-  task :reset_password_email, %i[provider start stop base_url] => :environment do |_task, args|
+  task :reset_password_email, %i[base_url provider] => :environment do |_task, args|
     args.with_defaults(provider: 'greenlight')
-    start, stop = range(args)
 
     root_url = "#{args[:base_url]}/"
 
+    # After the migration, all the users except those with an external_id will need to reset their password
     User.where(external_id: nil, provider: args[:provider])
-        .find_each do |user|
+        .find_each(batch_size: BATCH_SIZE) do |user|
       token = user.generate_reset_token!
 
       UserMailer.with(user:,
@@ -39,22 +39,9 @@ namespace :migration do
 
       success "Sent reset password email to #{user.email}"
     end
-
   end
 
   private
-
-  def range(args)
-    start = args[:start].to_i
-    start = 1 unless start.positive?
-
-    stop = args[:stop].to_i
-    stop = nil unless stop.positive?
-
-    raise red "Unable to migrate: Invalid provided range [start: #{start}, stop: #{stop}]" if stop && start > stop
-
-    [start, stop]
-  end
 
   def reset_password_url(root_url, token)
     "#{root_url}reset_password/#{token}"
--- a/lib/tasks/reset_password_email.rake
+++ b/lib/tasks/reset_password_email.rake
@@ -19,7 +19,7 @@
 require_relative 'task_helpers'
 
 namespace :migration do
-  BATCH_SIZE = 500
+  batch_size = 500
 
   desc 'Send a reset password email to users'
   task :reset_password_email, %i[base_url provider] => :environment do |_task, args|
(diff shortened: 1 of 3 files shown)


lib/tasks/reset_password_email.rake
```diff
@@ -0,0 +1,60 @@
...
+namespace :migration do
+  include ActionView::Helpers::AssetUrlHelper
+
+  desc 'Send a reset password email to users'
+  task :reset_password_email, %i[provider start stop root_url] => :environment do |_task, args|
+    args.with_defaults(provider: 'greenlight')
+    start, stop = range(args)
+
+    User.where(external_id: nil, provider: args[:provider])
+        .find_each do |user|
```
scouillard, line 30: find_each is not working with arguments. unsure why

lib/tasks/reset_password_email.rake
```diff
@@ -0,0 +1,60 @@
...
+      token = user.generate_reset_token!
+
+      UserMailer.with(user:,
+                      reset_url: reset_password_url(args[:root_url], token),
+                      base_url: args[:root_url],
```
scouillard, line 35: Looking into difference between `root_url` and `base_url` and possible pitfalls

lib/tasks/reset_password_email.rake
```diff
@@ -0,0 +1,62 @@
...
+
+namespace :migration do
+  include ActionView::Helpers::AssetUrlHelper
+
+  desc 'Send a reset password email to users'
+  task :reset_password_email, %i[provider start stop base_url] => :environment do |_task, args|
+    args.with_defaults(provider: 'greenlight')
+    start, stop = range(args)
+
+    root_url = "#{args[:base_url]}/"
```
scouillard, line 29: Difference between `root_url` and `base_url` seems to be the `/` at the end. Unsure if this is always the case.

scouillard:
This seem to be working.
But, as reported by one of or users, the emails seem to be broken - currently investigating the problem.
Also, trying to improve the logs if the task fails.

lib/tasks/reset_password_email.rake
```diff
@@ -0,0 +1,61 @@
...
+          info    "  name: #{user.name}"
+          info    "  email: #{user.email}"
+        end
+      rescue StandardError => e
+        err "Unable to send reset password email - #{e}"
+      end
+    else
+      err 'SMTP Server not set. Skipping sending reset password emails.'
+      exit 0
+    end
```
farhatahmad, line 53: Rather than an `if else` you can move this to the top 
```
if ENV['SMTP_SERVER'].blank?
  err 'SMTP Server not set. Skipping sending reset password emails.'
  exit 0
end
...
 
```

lib/tasks/reset_password_email.rake
```diff
@@ -0,0 +1,61 @@
...
+                          base_url: args[:base_url],
+                          provider: args[:provider]).reset_password_email.deliver_now
+
+          success 'Successfully sent reset password email to:'
+          info    "  name: #{user.name}"
+          info    "  email: #{user.email}"
+        end
+      rescue StandardError => e
+        err "Unable to send reset password email - #{e}"
+      end
```
farhatahmad, line 49: In this case, if 1 email fails, the whole thing fails. Why don't we move inside the `.find_each` so that this doesn't happen?

lib/tasks/reset_password_email.rake
scouillard: Have you seen this @farhatahmad

lib/tasks/reset_password_email.rake
```diff
@@ -0,0 +1,68 @@
...
+
+    info 'Checking connection to SMTP Server...'
+    if ENV['SMTP_SERVER'].present?
+      begin
+        UserMailer.with(to: ENV.fetch('SMTP_SENDER_EMAIL', nil), subject: ENV.fetch('SMTP_SENDER_EMAIL', nil)).test_email.deliver_now
+      rescue StandardError => e
+        failed("Unable to connect to SMTP Server - #{e}")
+        exit 0
+      end
+    end
```
scouillard, line 43: I have added this block (copied from configuration.rake) to exit the rake tasks if the connection to the SMTP server fails.

farhatahmad (approved):

Issue #5027, opened 2023-03-13
"Account Activation" email sends link in HTML-part of email in a way that Thunderbird "Simple HTML" view will not show the link


EmmyGraugans:
(related to #5024, but not identical)

When registering a new user-account in Greenlight-V3, the email contains in "multipart/alternative" an HTML-Part that will not be displayed correctly (at least not usefully) in Thunderbird with setting `View -> Message Body As -> Simple HTML` (which is the default), It will be displayed as

```
Welcome to BigBlueButton!

To get started, please activate your account by clicking on the button below.

The link will expire in 24 hours.

If the link expires, please log in and request a new activation email.
```

The Link is sent as `<a href=...><button>...</button></a>`, however "button" is one of those advanced HTML-elements that Thunderbird leaves out in Simple HTML, so that the entire link isn't shown at all.

scouillard:
Hi,

This will be looked into.

Thank you for the feedback.

EmmyGraugans:
An addition that I noticed only later: In Thunderbird, the Button doesn't even work when in `Original HTML`, probably because of the "target=_blank" attribute, which will open the link in a new Tab in Thunderbird, but it won't load the webpage. Instead one has to do a right-click on the button and say "Open Link in Browser".

farhatahmad:
https://github.com/bigbluebutton/greenlight/releases/tag/release-3.0.2

PR #5039, opened 2023-03-16
Fix Mailer Views

scouillard:
This aims to solve #5027 

Remove the <button> element which is not a proper simple HTML element - and is useless.
Fix the .text.erb templates

Emails now seem to work on GMAIL, Thunderbird. In HTML, simple HTML, text.


scouillard:
Did not find an (easy) way to remove the (license) comments from the plain text emails and the console logs.
Unless plain text emails are popular amongst our users, I think it can stay as it is.

farhatahmad (approved):

PR #5042, opened 2023-03-17
Fix Room Join spinner

scouillard:
Fix the Spinner in the join room button.
We were loading the default (bigger) bootstrap Spinner for some reason.


farhatahmad:
Can you add a screenshot with the difference?

scouillard:
image
image

PR #5072, opened 2023-03-27
fix(room join): room join waiting queue improvement

scouillard:
As request by #5071 
- Add sound when user successfully join a meeting
- Reduce waiting queue delay from 7 seconds to in between 2 and 5 seconds


scouillard:
I am using this opportunity to try and improve the waiting queue experience in RoomJoin.

The original 7 seconds of buffer when receiving the confirmation from the BBB server that the meeting is starting and attempting to redirect the user in the waiting queue seems long - and it was pretty clunky with the popping sound.

Do we have any reason to believe that we need a 7 seconds buffer in between? @KH-Amir-TN 

The method I implemented will return a random delay between 2 and 5 seconds and is easy to tune.

The randomness is to help with the traffic in case there are many users in the waiting queue attempting to join at the same time.

app/javascript/components/rooms/room/join/RoomJoin.jsx
```diff
@@ -100,15 +101,42 @@ export default function RoomJoin() {
...
     //  When meeting starts this logic will be fired, indicating the event to waiting users (through a toast) for UX matter.
     //  Logging the event for debugging purposes and refetching the join logic with the user's given input (name & codes).
-    //  With a delay of 7s to give reasonable time for the meeting to fully start on the BBB server.
     if (hasStarted) {
       toast.success(t('toast.success.room.meeting_started'));
-      console.info(`Attempting to join the room(friendly_id): ${friendlyId} meeting in 7s.`);
-      setTimeout(methods.handleSubmit(handleJoin), 7000); // TODO: Amir - Improve this race condition handling by the backend.
+      playNotifySound();
+      console.info(`Attempting to join the room(friendly_id): ${friendlyId} meeting in 3s.`);
+      setTimeout(methods.handleSubmit(handleJoin), joinDelay()); // TODO: Amir - Improve this race condition handling by the backend.
```
farhatahmad, line 139: Can we move the sound to `handleJoin`? With this way, its possible that you hear the sound and just sit on the page waiting for 5 seconds

esbuild.dev.mjs
```diff
@@ -10,6 +10,7 @@ await esbuild.build({
   loader: {
     '.png': 'dataurl',
     '.svg': 'text',
+    '.mp3': 'base64',
```
KH-Amir-TN, line 13: I guess this way we're bundling the audio as part of our javascript assets which is not required,

app/javascript/components/rooms/room/join/RoomJoin.jsx
scouillard: Yes, of course.
It was very clunky with the 7 seconds, but since I've reduced the delay it felt pretty good.
Should we move the toast pop up in handleJoin as well?

KH-Amir-TN (changes requested):

app/javascript/components/rooms/room/join/RoomJoin.jsx
```diff
@@ -100,15 +101,42 @@ export default function RoomJoin() {
     }
   }, [roomStatusAPI.isSuccess]);
 
+  // Plays a notify sound when the meeting starts if the user was in a waiting queue
+  const playNotifySound = () => {
+    const audio = new Audio(`data:audio/mpeg;base64,${NotifySound}`);
+    audio.play()
```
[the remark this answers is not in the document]
KH-Amir-TN: Playing the audio will not prevent the rest of the frontend from running?

app/javascript/components/rooms/room/join/RoomJoin.jsx
scouillard: I feel like this is a case of perfect is enemy of the good.
It is not a perfect solution, but it is better than what we had.
And did not take much resources.

scouillard:
How did you come up with the 7 seconds @KH-Amir-TN ?
7 seconds is too long and makes it very clunky.
Has it been tested or is it speculation?

app/javascript/components/rooms/room/join/RoomJoin.jsx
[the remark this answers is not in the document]
scouillard: It is async

app/javascript/components/rooms/room/join/RoomJoin.jsx
KH-Amir-TN: The random delay can be done on the client side.
However, the issue now is the delay minimum that may cause users to experience issues on the auto join.
I think we'd better tackle that independently and improve the join workflow form the API itself.
Then having a delay that is less than 5 seconds for the sake of buffering will not cause any issues since the join URL will only be received after the meeting has actually became ready.

KH-Amir-TN:
> How did you come up with the 7 seconds @KH-Amir-TN ? 7 seconds is too long and makes it very clunky. Has it been tested or is it speculation?

This design is optimistic as explained,
The 7 seconds is a result of some testing I've done without experiencing issues on the auto join.
It was 10 seconds but it felt too long.
At 5 seconds I remember experiencing some issues but for the record the staging server I used isn't that powerful.

scouillard:
> > How did you come up with the 7 seconds @KH-Amir-TN ? 7 seconds is too long and makes it very clunky. Has it been tested or is it speculation?
> 
> This design is optimistic as explained, The 7 seconds is a result of some testing I've done without experiencing issues on the auto join. It was 10 seconds but it felt too long. At 5 seconds I remember experiencing some issues but for the record the staging server I used isn't that powerful.

While I agree that this should be done in the backend, I don't think there are plans to re-work our API in the near future.
Meanwhile, this is a quick hack that improves the experience - and ultimately a step in the right direction.
If users complain (or we realise) that the connection fails, it is easy to change the min and max delay time.

app/javascript/components/rooms/room/join/RoomJoin.jsx
```diff
@@ -100,15 +101,42 @@ export default function RoomJoin() {
...
+    return min + (randomStep * step);
+  };
+
   useEffect(() => {
     // Meeting started:
     //  When meeting starts this logic will be fired, indicating the event to waiting users (through a toast) for UX matter.
     //  Logging the event for debugging purposes and refetching the join logic with the user's given input (name & codes).
-    //  With a delay of 7s to give reasonable time for the meeting to fully start on the BBB server.
     if (hasStarted) {
       toast.success(t('toast.success.room.meeting_started'));
```
farhatahmad, line 136: Can we move this into the handleJoin as well?

scouillard: Yep

app/javascript/components/rooms/room/join/RoomJoin.jsx
scouillard: Right. By moving the toast and sound into the delay it should be fine.

scouillard: Might have to do some tuning tho. I recall the Spinner stops at some point.

esbuild.dev.mjs
scouillard: Still has to be tested in prod

PR #5108, opened 2023-03-31
Client routes helpers: Minor improvements.

KH-Amir-TN:
## Description
This PR introduces some improvements (namespacing, refactoring) to the current `ClientRoutable` concern mainly to make the source of truth for all client routes helpers on the backend while being as closer as possible to the rails routes helpers API.
So, one should be able to work with the `ClientRoutable` with the same API of the `Rails.application.routes.url_helpers`.

> `client_` prefix was added indicate that the used route helper is for a client side route and to avoid conflicts with the rails routes helpers.

## Testing Steps
- Set the `RELATIVE_URL_ROOT` variable in `.env` to whatever you like e.g. `/gl`.
- Restart Greenlight: on dev run the `dev` script and for production use `khamirtn/greenlight:relative_path_fixes` docker image for testing or if having a production environment run the `start` script.
- The application should be served relatively to your `$RELATIVE_URL_ROOT` e.g. `http://localhost:5050/rooms becomes https://localhost:5050/gl/rooms` if `RELATIVE_URL_ROOT=/gl`.
- Everything on the application should work as expected and the issues mentioned above should not be observed.
- Try different `RELATIVE_URL_ROOT` and test the application.

### DIFF
--- a/app/controllers/api/v1/admin/invitations_controller.rb
+++ b/app/controllers/api/v1/admin/invitations_controller.rb
@@ -20,6 +20,8 @@ module Api
   module V1
     module Admin
       class InvitationsController < ApiController
+        include ClientRoutable
+
         before_action do
           ensure_authorized('ManageUsers')
         end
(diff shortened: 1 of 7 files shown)


KH-Amir-TN:
# DEPENDS ON #5106

PR #5183, opened 2023-05-15
fix(join meeting): Moderator are authorized to start the meeting

### LINKED ISSUE #5049: Share Access should have an option to share as Moderator or User
I created a new role "Teacher" in my GL installation.

I would like the Teacher to be able to a "Start Meeting" for the rooms that they have access to. They should also see their rooms etc without being able to create new rooms as room creation / management is a different mechanism which would otherwise confuse non-tech savvy teachers. While as of now it is all driven by the "Create Room" permission. Which I wouldn't want to give to teachers. 

So:

1. Either all users (not guests) that **have access** to a room should be able to Start Meeting.
2. Or while sharing access we should be able to choose between Moderator and User access to make a distinction as to who can "Start Meeting.

scouillard:
Moderators were not authorized to start a meeting. 
Add the logic to the MeetingsController.
Add the specs.
Fixes #5049 (and few threads on the google group)

### DIFF
--- a/spec/controllers/meetings_controller_spec.rb
+++ b/spec/controllers/meetings_controller_spec.rb
@@ -149,11 +149,12 @@ RSpec.describe Api::V1::MeetingsController, type: :controller do
       expect(JSON.parse(response.body)['data']).to eq({ 'joinUrl' => 'JOIN_URL', 'status' => true })
     end
 
-    it 'returns status false if the meeting is NOT running' do
+    it 'returns status false if the meeting is NOT running and the user is NOT authorized to start the meeting' do
       allow_any_instance_of(BigBlueButtonApi).to receive(:meeting_running?).and_return(false)
       expect_any_instance_of(BigBlueButtonApi).not_to receive(:join_meeting)
 
-      post :status, params: { friendly_id: room.friendly_id, name: user.name }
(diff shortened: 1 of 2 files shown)


spec/controllers/meetings_controller_spec.rb
```diff
@@ -149,11 +149,12 @@
       expect(JSON.parse(response.body)['data']).to eq({ 'joinUrl' => 'JOIN_URL', 'status' => true })
     end
 
-    it 'returns status false if the meeting is NOT running' do
+    it 'returns status false if the meeting is NOT running and the user is NOT authorized to start the meeting' do
       allow_any_instance_of(BigBlueButtonApi).to receive(:meeting_running?).and_return(false)
       expect_any_instance_of(BigBlueButtonApi).not_to receive(:join_meeting)
 
-      post :status, params: { friendly_id: room.friendly_id, name: user.name }
+      post :status, params: { friendly_id: test_room.friendly_id, name: user.name }
```
scouillard, line 156: I believe that #status should never return false to a Moderator as hitting the #status endpoint as a moderator/room owner will start the meeting.

scouillard: I was confused, this test should not have been passing - it is because the #status endpoint should never be hit by a room owner because the RoomJoin page is not available to room owners.

PR #5197, opened 2023-05-22
Fixed bad OmniAuth flow when deployed on a subdirectory (Closes #5176).

### LINKED ISSUE #5176: Sign in/up-Buttons don't respect RELATIVE_URL_ROOT
** How to reproduce **
1. Configuring GLv3-Container with 

```
OPENID_CONNECT_CLIENT_ID=greenlight-v3
#[...]
RELATIVE_URL_ROOT=/c
```
and configure nginx accordingly.
2. Visit `https://host-name/c`
3. See GLv3 start page
4. Click either "sign in" or "sign up" button
5. get transported to `https://host-name/auth/openid`, which in my case just shows a blank page.

** Expected behavior **
Get transported to `https://host-name/c/auth/openid`, so the request actually is sent to GLv3...

** Additional context **
`grep`ping in the container yields following result:
```
/usr/src/app/app/javascript/components/home/AuthButtons.jsx:      <Form action="/auth/openid_connect" method="POST" data-turbo="false">
```
This is not the only place where `RELATIVE_URL_ROOT` doesn't seem to be respected...

KH-Amir-TN:
## Testing scenarios:
When external AuthN is enabled and a relative URL root path is set to **any value** the following cases should behave as expected:
1. Users should be able authenticate through the identity provider and have an account created and be signed on the app.
> If the registration method is set to `invite` or `approval` the app should behave as expected when external AuthN is enabled and no matter what relative URL root path is used **including edge cases like with an invalid Invite**. 
2. Users should be authenticated and redirected back to Room join page when choosing to signin before joining (Rooms with require AuthN setting **OFF**).
3. Users should be redirected to the require AuthN page where applicable and be able to authenticate and be redirected to the room join page (Rooms with require AuthN setting **ON**).

### DIFF
--- a/app/controllers/external_controller.rb
+++ b/app/controllers/external_controller.rb
@@ -27,13 +27,8 @@ class ExternalController < ApplicationController
     provider = current_provider
 
     credentials = request.env['omniauth.auth']
-    user_info = {
-      name: credentials['info']['name'],
-      email: credentials['info']['email'],
-      language: extract_language_code(credentials['info']['locale']),
-      external_id: credentials['uid'],
-      verified: true
-    }
+
+    user_info = build_user_info(credentials)
 
     user = User.find_by(external_id: credentials['uid'], provider:)
     new_user = user.blank?
@@ -67,10 +62,14 @@ class ExternalController < ApplicationController
     session[:session_token] = user.session_token
 
     # TODO: - Ahmad: deal with errors
-    redirect_location = cookies[:location]
-    cookies.delete(:location)
-    return redirect_to redirect_location if redirect_location&.match?('\A\/rooms\/\w{3}-\w{3}-\w{3}-\w{3}\/join\z')
 
+    redirect_location = cookies.delete(:location)
+
+    return redirect_to redirect_location, allow_other_host: false if redirect_location&.match?('\/rooms\/\w{3}-\w{3}-\w{3}-\w{3}\/join\z')
+
+    redirect_to root_path
+  rescue ActionController::Redirecting::UnsafeRedirectError => e
+    Rails.logger.error("Unsafe redirection attempt: #{e}")
     redirect_to root_path
   rescue StandardError => e
     Rails.logger.error("Error during authentication: #{e}")
@@ -127,5 +126,15 @@ class ExternalController < ApplicationController
 
     # Try to delete the invitation and return true if it succeeds
     Invitation.destroy_by(email:, provider: current_provider, token:).present?
+  end
+
+  def build_user_info(credentials)
+    {
+      name: credentials['info']['name'],
+      email: credentials['info']['email'],
+      language: extract_language_code(credentials['info']['locale']),
+      external_id: credentials['uid'],
+      verified: true
+    }
   end
 end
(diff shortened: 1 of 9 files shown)


app/controllers/external_controller.rb
```diff
@@ -67,12 +67,12 @@ def create_user
     # TODO: - Ahmad: deal with errors
     redirect_location = cookies[:location]
     cookies.delete(:location)
-    return redirect_to redirect_location if redirect_location&.match?('\A\/rooms\/\w{3}-\w{3}-\w{3}-\w{3}\/join\z')
+    return redirect_to redirect_location if redirect_location&.match?(%r{\A(/\w{1,32}){0,3}/rooms/\w{3}-\w{3}-\w{3}-\w{3}/join\z})
```
einhirn, line 70: This matching seems a bit weird to me - why not use the "#{root_path}" macro here?
This will only match root_paths with up to three subdirs and each being from one to 32 characters long. An arbitrary limitation which probably won't bite anyone, but why risk it?

KH-Amir-TN: I see what you mean, I'll make sure to keep this in mind as we work on this issue.

app/javascript/components/rooms/room/join/RoomJoin.jsx
```diff
@@ -55,8 +55,7 @@ export default function RoomJoin() {
 
   const methods = useForm(joinFormConfig);
 
-  const location = useLocation();
-  const path = encodeURIComponent(location.pathname);
+  const path = encodeURIComponent(document.location.pathname);
```
KH-Amir-TN, line 58: `useLocation` seems to encourage determinism so it will always provide a relative path after a match (the basename will be filtered).
In such case and because of the API restrictions I guess there's no issues to use the `document.location` API.

app/controllers/external_controller.rb
```diff
@@ -25,13 +27,8 @@ def create_user
...
     credentials = request.env['omniauth.auth']
-    user_info = {
-      name: credentials['info']['name'],
-      email: credentials['info']['email'],
-      language: extract_language_code(credentials['info']['locale']),
-      external_id: credentials['uid'],
-      verified: true
-    }
+
+    user_info = build_user_info(credentials)
```
KH-Amir-TN, line 31: Reducing the method complexity,

app/controllers/external_controller.rb
```diff
@@ -58,21 +55,25 @@ def create_user
...
 
     # TODO: - Ahmad: deal with errors
-    redirect_location = cookies[:location]
-    cookies.delete(:location)
-    return redirect_to redirect_location if redirect_location&.match?('\A\/rooms\/\w{3}-\w{3}-\w{3}-\w{3}\/join\z')
 
-    redirect_to '/'
+    redirect_location = cookies.delete(:location)
+
+    return redirect_to redirect_location, allow_other_host: false if redirect_location
```
farhatahmad, line 68: Shouldn't we still scope this to only redirect to room join?

KH-Amir-TN: `return redirect_to redirect_location, allow_other_host: false if redirect_location&.match?('\/rooms\/\w{3}-\w{3}-\w{3}-\w{3}\/join\z'')` ?

farhatahmad: Yeah

farhatahmad (approved):

farhatahmad (approved):

PR #5206, opened 2023-05-29
Signup: Adding restrictions for edge cases. 

KH-Amir-TN:
- [x] Restricted API access when external AuthN is enabled.
- [x] Restricted UI access when external AuthN is enabled.
- [x] Restrict API access when external AuthN is DISABLED to admins only.
- [x] Hide the `New User` button on the UI when external AuthN is enabled.


app/javascript/components/users/authentication/Signup.jsx
```diff
@@ -22,18 +22,29 @@ import { toast } from 'react-toastify';
...
+    return <Navigate to="/" replace />;
+  }
+
+  if (RegistrationMethodSettingAPI.data === 'invite' && !inviteToken) {
     toast.error(t('toast.error.users.invalid_invite'));
     return <Navigate to="/" replace />;
   }
 
+  if (isLoading) {
+    return null;
```
KH-Amir-TN, line 45: We might want to add in the future a global loading state.
This however prevents the UI flickering for now.

app/controllers/api/v1/users_controller.rb
```diff
@@ -79,7 +82,7 @@ def create
           user.generate_session_token!
           session[:session_token] = user.session_token unless current_user # if this is NOT an admin creating a user
 
-          render_data data: current_user, serializer: CurrentUserSerializer, status: :created
+          render_data data: user, serializer: CurrentUserSerializer, status: :created
```
KH-Amir-TN, line 85: The API is returning the current user data and not the created user.

app/controllers/api/v1/users_controller.rb
```diff
@@ -39,28 +39,34 @@ def show
...
         end
 
+        # Users created by a user will have the creator language by default with a fallback to the server configured default_locale.
+        create_user_params[:language] = current_user&.language || I18n.default_locale if create_user_params[:language].blank?
+
+        user = UserCreator.new(user_params: create_user_params.except(:invite_token), provider: current_provider, role: default_role).call
+
+        smtp_enabled = ENV['SMTP_SERVER'].present?
+
+        user.verify! unless smtp_enabled
```
farhatahmad, line 68: We can just drop the `smtp_enabled` variable here since it's only used once

farhatahmad: It can just be `user.verify! unless ENV['SMTP_SERVER'].present?`

KH-Amir-TN: Okay

KH-Amir-TN: It's used also in L74,
Do you still wanna drop it?

farhatahmad: Ah no I missed that. All good then

PR #5207, opened 2023-05-29
Signin: Adding restrictions for edge cases.

KH-Amir-TN:
- [x] Restrict API access for unauthorized requests (authenticated requests...).
- [x] Improve feedback (UX).


KH-Amir-TN:
# DEPENDS ON #5206

PR #5219, opened 2023-06-05
Added OAuth functionality to MT

farhatahmad:


config/initializers/omniauth.rb
```diff
@@ -18,8 +18,25 @@
 
 Rails.application.config.middleware.use OmniAuth::Builder do
   issuer = ENV.fetch('OPENID_CONNECT_ISSUER', '')
+  lb = ENV.fetch('LOADBALANCER_ENDPOINT', '')
 
-  if issuer.present?
+  if lb.present?
+    provider :openid_connect, setup: lambda { |env|
+      request = Rack::Request.new(env)
+      current_provider = request.params['current_provider']
```
KH-Amir-TN, line 26: Why not infer the current_provider from the request object?
This may duplicate the `parse_user_domain` logic (which we can DRY if that's the intent) but makes the request the only source of truth for such value and It's somehow more secure (this doesn't mean that the proposed solution is insecure).
We may remove the ENV API in the future,

lib/tasks/admin.rake
```diff
@@ -32,4 +32,28 @@ namespace :admin do
...
+
+  desc 'Create an super admin account'
+  task :super_admin, %i[name email password] => :environment do |_task, args|
+    super_admin_email = "superadmin-#{args[:email]}"
+
+    user = User.create(
+      name: args[:name],
+      email: super_admin_email,
+      password: args[:password],
+      role: Role.find_by(name: 'SuperAdmin'),
```
KH-Amir-TN, line 44: Umm, Why don't we be more explicit and `find_by(:name, :provider)`?

app/controllers/api/v1/admin/tenants_controller.rb
```diff
@@ -37,7 +37,7 @@ def index
 
         # POST /api/v1/admin/tenants
         def create
-          name = tenant_params[:name].downcase.gsub(/[-\s]/, '_')
+          name = tenant_params[:name]
```
KH-Amir-TN, line 40: We do trust the admin to provide conform names?

farhatahmad: Yep - we need `-` in the names

lib/tasks/admin.rake
farhatahmad: Yeah we can do that

config/initializers/omniauth.rb
KH-Amir-TN: I see what you mean,

However what captured my intention is the fact that `request.params['current_provider']` equals the value of the hidden input field.
It's true that it's supposed to match the ENV API result when the page loads which equals to the `current_provider` from the API perspective.

It's true that a request headers can be easily manipulated but I thought that the request body is a bit easier to manipulate plus the flow will involve more steps and therefore more room for issues (to get the provider value instead of parsing the host header the value must propagated in a ping pong style with no verification). 

It's just a note not a recommendation,
I'm okay with both scenarios,

PR #5221, opened 2023-06-06
Recordings: Added `listed` recordings attribute. 

KH-Amir-TN:


farhatahmad:
Do we need a seperate variable here? I'd rather we just stick to one `visibility`

PR #5227, opened 2023-06-08
Public recordings: Add foundation and structural changes.

KH-Amir-TN:
- [x] Add public visibilities and their checks to Recording model.
- [x] Update `RecordingsController#update_visibility` API.
## TODO:
- [ ] Update `RecordingCreator` service.
- [ ] Add Public recordings API.
- [ ] Update UIs (RecordingRow, Room join,....)

### DIFF
--- a/app/controllers/api/v1/recordings_controller.rb
+++ b/app/controllers/api/v1/recordings_controller.rb
@@ -65,17 +65,16 @@ module Api
       # POST /api/v1/recordings/update_visibility.json
       # Update's a recordings visibility by setting publish/unpublish and protected/unprotected
       def update_visibility
-        new_visibility = params[:visibility]
-        old_visibility = @recording.visibility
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/recordings_controller.rb
```diff
@@ -114,6 +113,18 @@ def recording_params
       def find_recording
         @recording = Recording.find_by! record_id: params[:id]
       end
+
+      def visibility_params_of(visibility)
+        params_of = {
+          Recording::VISIBILITIES[:unpublished] => { publish: false, meta_hash: { protect: false, 'meta_gl3-listed': false } },
```
farhatahmad, line 119: Why `meta_gl3-listed`? We should be using the same value as v2 to allow for backwards compatibility

farhatahmad (approved):

Issue #5048, opened 2023-03-21
Browser title for the room page is wrong


shabeepk:
Browser title (<title> and <og:title> tags) for room's page should have the room's title rather than the word "BigBlutButton"

Sharing room links on social media and other places show the title as "BigBlutButton" while ideally it should show the title Room's title so the links have better og/thumbnail titles generated.

PR #5237, opened 2023-06-15
Added dynamic og titles and room titles

farhatahmad:
fixes https://github.com/bigbluebutton/greenlight/issues/5048

Added a dynamic page title to the browser title and the opengraph title specifically for the room page

Browser Title Example:
image

Example from Slack:
image

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -21,4 +21,12 @@ module ApplicationHelper
     asset_path = SettingGetter.new(setting_name: 'BrandingImage', provider: current_provider).call
     asset_url(asset_path)
   end
+
+  def page_title
+    match = request&.url&.match('\/rooms\/(\w{3}-\w{3}-\w{3}-\w{3})')
+    return 'BigBlueButton' if match.blank?
+
+    room_name = Room.find_by(friendly_id: match[1])&.name
+    room_name || 'BigBlueButton'
+  end
 end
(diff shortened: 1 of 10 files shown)


app/helpers/application_helper.rb
```diff
@@ -21,4 +21,12 @@ def branding_image
     asset_path = SettingGetter.new(setting_name: 'BrandingImage', provider: current_provider).call
     asset_url(asset_path)
   end
+
+  def page_title
+    match = request&.url&.match('\/rooms\/(\w{3}-\w{3}-\w{3}-\w{3})')
+    return 'BigBlueButton' if match.blank?
+
+    room_name = Room.find_by(friendly_id: match[1])&.name
+    room_name || 'BigBlueButton'
```
farhatahmad, line 30: Since opengraph won't run any JS on the page, we need to dynamically pull the room name on the server side for it to show in open graph

app/views/layouts/application.html.erb
```diff
@@ -17,11 +17,10 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>BigBlueButton</title>
```
KH-Amir-TN, line 20: Is there a reason why we'd remove this?

KH-Amir-TN: I think it can be used by search engines

app/javascript/components/rooms/room/Room.jsx
```diff
@@ -58,6 +59,10 @@ export default function Room() {
   return (
     <>
       <div className="wide-white">
+        <Helmet>
```
KH-Amir-TN, line 62: This is really nice !
We could create a component called Title with a children prop as the value maybe to improve the syntax?
```
<Title> My Title !</Title>
```

KH-Amir-TN (approved):

PR #5250, opened 2023-06-19
Set optional room configs to default_disabled

scouillard:
V2 Rooms configs that are set to "optional" are currently set to "default_enabled". 
V2 Rooms configs that are set to "optional" will now be set to "default_disabled".
Add exception for the Record Rooms Config, which if set to "optional" will be set to "default_enabled".

### DIFF
--- a/lib/tasks/migrations/migrations.rake
+++ b/lib/tasks/migrations/migrations.rake
@@ -202,9 +202,13 @@ namespace :migrations do
       PreuploadPresentation: setting.get_value('Preupload Presentation'),
     }.compact
 
+
+    # Sets Record to default_enabled in V3 if set to optional in V2
+    rooms_config_record_value = infer_room_config_value(setting.get_value('Room Configuration Recording'))
+
     # RoomConfigurations
     rooms_configurations = {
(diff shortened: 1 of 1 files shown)


lib/tasks/migrations/migrations.rake
```diff
@@ -202,9 +202,13 @@ namespace :migrations do
       PreuploadPresentation: setting.get_value('Preupload Presentation'),
     }.compact
 
+
+    # Sets Record to default_enabled in V3 if set to optional in V2
+    rooms_config_record_value = infer_room_config_value(setting.get_value('Room Configuration Recording'))
```
farhatahmad, line 207: Wont `infer_room_config_value` return `default_enabled`?

scouillard: Nope, `default_enabled` has been completely removed from `inter_room_config_value`'s return.

**V2                          V3**
enabled or true -> true
disabled            -> false
optional             -> optional (default_disabled)

A clause has been added line 211: if record is "optional", it is re-written to "default_enabled"

farhatahmad (approved):

PR #5266, opened 2023-06-20
Fix relative root in Migrations task

### LINKED ISSUE #5258: V2/3: Migration tasks do not seem to work with relative root
```
V3_ENDPOINT=https://greenlight-development.XYZ.de/gl3/
```

results in a request logged by nginx like this:

```
2023/06/20 13:10:33 [error] 30513#30513: *2064 open() "/usr/share/nginx/html/api/v1/migrations/roles.json" failed (2: No such file or directory),
 client: xxx.xxx.xxx.xxx,
 server: greenlight-development.XYZ.de,
 request: "POST /api/v1/migrations/roles.json HTTP/1.1",
 host: "greenlight-development.XYZ.de"
```

I'd guess the `/api/v1/migrations/roles.json` should rather be `api/v1/migrations/roles.json`?

scouillard:
Fixes #5258 

The reason why `URI::join` is preferred over setting path directly is that it can correctly handle different situations and ensure a valid URI. For example, if the base URI already ends with a slash or if the path already starts with a slash, `URI::join` will not add or remove any slashes, thus ensuring the correct URL structure.

### DIFF
--- a/lib/tasks/migrations/migrations.rake
+++ b/lib/tasks/migrations/migrations.rake
@@ -265,8 +265,8 @@ namespace :migrations do
   def uri(path)
     raise red 'Unable to migrate: No "V3_ENDPOINT" provided, please check your .env file.' unless ENV["V3_ENDPOINT"].present?
 
-    res = URI(ENV["V3_ENDPOINT"])
-    res.path = "/api/v1/migrations/#{path}.json"
+    base_uri = URI(ENV["V3_ENDPOINT"])
+    res = URI::join(base_uri, "/api/v1/migrations/#{path}.json")
     res
   end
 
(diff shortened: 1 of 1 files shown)


lib/tasks/migrations/migrations.rake
```diff
@@ -265,8 +265,8 @@ namespace :migrations do
   def uri(path)
     raise red 'Unable to migrate: No "V3_ENDPOINT" provided, please check your .env file.' unless ENV["V3_ENDPOINT"].present?
 
-    res = URI(ENV["V3_ENDPOINT"])
-    res.path = "/api/v1/migrations/#{path}.json"
+    base_uri = URI(ENV["V3_ENDPOINT"])
+    res = URI::join(base_uri, "/api/v1/migrations/#{path}.json")
```
farhatahmad, line 269: I think this has to `res = URI::join(base_uri, "api/v1/migrations/#{path}.json")`

scouillard: You are right - it will be treated as an absolute path and base_uri will be discarded if it starts with a slash

farhatahmad (approved):

Issue #5059, opened 2023-03-22
Password reset function is displayed even though no smtp-server is configured


dmtrhfr:
Currently the password reset button is displayed even though there is no smtp-server is configured.
So it's possible to reset my password and lock myself out of my account, with no option to retrieve the account and it's rooms.

scouillard:
Hi,
This will be looked into.
Thank you for the feedback.

herzkerl:
Dear @scouillard, this issue has been closed on May 9th, but without any other issues or PRs linked to it. Has this been resolved?

scouillard:
Hi @herzkerl
You are right - this was closed by mistake.
It has not been solved yet.
Thank you for the report.

PR #5303, opened 2023-07-04
Add SMTP check for reset password link

scouillard:
Solves #5059 

The "Reset Password link on the sign in page should be removed if the SMTP server is not set up.
The `/forget_password` component should redirect to homepage if the SMTP server is not set up.

### DIFF
--- a/app/javascript/components/users/password_management/ForgetPassword.jsx
+++ b/app/javascript/components/users/password_management/ForgetPassword.jsx
@@ -16,13 +16,21 @@
 
 import React from 'react';
 import Card from 'react-bootstrap/Card';
-import { Link } from 'react-router-dom';
+import { Link, Navigate } from 'react-router-dom';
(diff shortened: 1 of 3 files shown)


app/javascript/components/users/password_management/ForgetPassword.jsx
```diff
@@ -16,13 +16,21 @@
...
 import { useTranslation } from 'react-i18next';
 import ForgetPwdForm from './forms/ForgetPwdForm';
 import Logo from '../../shared_components/Logo';
+import useEnv from '../../../hooks/queries/env/useEnv';
 
 export default function ForgetPassword() {
   const { t } = useTranslation();
+  const { data: env, isLoading } = useEnv();
+
+  if (isLoading) return null;
```
farhatahmad, line 29: Haven't we stopped doing this due to weird loading issues?

Issue #5284, opened 2023-06-23
[v3] A user associated to a role to manage users can upgrade user roles to administration role


michaaa:
Greenlight V3 (release-3.0.5) doesn't seem to check this:

- a user of a role that can manage users can give other users an admin role
- it is possible to edit the own user account to "upgrade" it to a higher privileges role (i.e. admin)

<img width="698" alt="Bildschirmfoto_2023-06-23_um_13_58_05" src="https://github.com/bigbluebutton/greenlight/assets/134079/ec42ef59-5fc0-4cfb-8a05-10f2f2c2c447">

User is associated to "Verwaltung" and can change its own role to "Administrator" (and this will be executed because of "manage user" ability):

<img width="478" alt="Bildschirmfoto_2023-06-23_um_14_27_08" src="https://github.com/bigbluebutton/greenlight/assets/134079/42991b61-6e13-431b-9927-9b99d61a3c8a">

In Greenlight V2 it is **not** possible this way:

- can't edit the own user account attributes for users with role "manage users" 
- give a role to users with higher access privileges than its own role

<img width="455" alt="Bildschirmfoto_2023-06-23_um_14_35_13" src="https://github.com/bigbluebutton/greenlight/assets/134079/6eaf4fde-3afa-432b-b1af-55288eb22ba0">

scouillard:
Hi,
After discussion, the users will not allowed to change their own role under any circumstance. 
However, users with the ManageUsers permission will still be able to change the roles of other users to any available role, even those with higher privileges than their own.
This opens up a variety of potential risks, as some permissions can have destructive effects. 
But, it is nearly impossible to safeguard against all possible misuse scenarios.
Please take this into account when assigning the ManageUsers permission to a user.
Thank you very much for the report.

michaaa:
Well, I think better solution would be: Only Admin Role owner should be able to give Admin Role to others. 

A user account with ManageUsers role won't change its own role (after this is fixed), but it's possible to add a new admin account then and use this one to change its own role (by using the new admin). So remove the ability to give Admin roles for non admins. However, I think this was implemented this way in Greenlight V2.

PR #5304, opened 2023-07-04
Add restriction to update user role

scouillard:
Solves #5284 

**Potentially risky situation in which an administrator with lower-level privileges is able to assign higher-level administrative roles to other users**

Users with ManageUsers role permission should not be able to assign a role to another user if the role is higher in the hierarchy than its own role. By higher role, we mean if the updated role has permissions that the user role does not have.

Currently, the roles list for the update user form is being fetched from the admin roles controller. It would make sense that the roles list should instead be in the admin users controller as it is an operation on a user, not a role.

```
  User can update someone a user role if:
  1. The user they are trying to update has a role that is not higher or equal to their own
  2. The user has the manage users permission
  3. The user is not trying to update themselves
```

- Add new `user_roles` endpoint that returns the list of roles that the current user can assign to other users
- Add new hook
- TODO: Add specs

### DIFF
--- a/app/controllers/api/v1/admin/users_controller.rb
+++ b/app/controllers/api/v1/admin/users_controller.rb
@@ -39,20 +39,18 @@ module Api
         # GET /api/v1/admin/users/user_roles.json
         # Returns the list of roles that the current user can assign to other users
         def user_roles
-          # Returns the permissions that the current user role has
-          current_user_permission_ids = current_user.role.role_permissions.where(value: 'true').pluck(:permission_id)
+          # Returns the permissions that the current user role does not have
+          current_user_missing_permission_ids = current_user.role.role_permissions.where(value: 'false').pluck(:permission_id)
 
-          # Returns the roles that have role permissions that the current user role does not have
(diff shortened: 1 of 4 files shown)


scouillard:
The logic is currently negative, I am working on finding a way to make it positive.

farhatahmad:
This seems like a slight over complication to me. Can we not just stop people from changing their own roles?

scouillard:
> This seems like a slight over complication to me. Can we not just stop people from changing their own roles?

I don't think this is the issue.

The issue is that users with ManageUsers permission can update another user's role to a role with more permissions.

For example, make another user an Administrator (which has all the permissions and is at the top of hierarchy).

It does not make sense to be able to give another user a higher role than your own.

I don't think there is an easy solution for this one.

scouillard:
Also, a user with ManageUsers permission could delete and ban a user with a higher role than their own.
This is potentially dangerous and kind of in the same scope of this PR.
The frontend logic could easily be duplicated to fix this.

image

PR #5310, opened 2023-07-06
Add force_path_style to storage.yml (#5296)

### LINKED ISSUE #5296: Allow the "force_path_style" option for S3 (active storage)
Hi,

In an environment that's using minio without using the bucket name as a subdomain, it would be nice to be able to set the "force_path_style" configuration of the S3 gem. 
Doc : https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Client.html

What would be needed to support this, without impacting any current installation, would be to add the following line under the `s3.bucket` at line 39 in the `config/storage.yml` file : 

`<%= ENV.fetch("S3_FORCE_PATH_STYLE", false) %>`

In some use cases, if your s3/minio is using a subdomain, you can cheat by moving it as the bucket name, and just appending the bucket name to the endpoint, but it's not really clean and I can't be sure that it wouldn't cause any issue down the road.

Because of the low impact and simplicity of the change, it seemed more pertinent to create an issue than a PR, since I'm not a contributor already.

Is that something that could be done ?

Thanks for your time.

Deeptwix:
Fixes #5296 

This variable allows user to use S3 servers configured to have the bucket name in urls in a `https://mys3server.com/my-bucket/` fashion rather than being forced to have it in the subdomain.

Adding the setting using an `ENV.fetch` makes sure that this doesn't brake any existing installation and it preserves the default `false` value which is probably the correct one for most users.

### DIFF
--- a/config/storage.yml
+++ b/config/storage.yml
@@ -36,6 +36,7 @@ s3:
   secret_access_key: <%= ENV['S3_SECRET_ACCESS_KEY'] %>
   region: <%= ENV['S3_REGION'] %>
   bucket: <%= ENV['S3_BUCKET'] %>
+  force_path_style: <%= ENV.fetch("S3_FORCE_PATH_STYLE", false) %>
 
 # Remember not to checkin your GCS keyfile to a repository
 google:


farhatahmad:
Don't forget to add this option to the `sample.env` so others know it exists

Deeptwix:
> Don't forget to add this option to the `sample.env` so others know it exists

Sorry I missed that, I've updated the branch.
I just added the option, without adding a comment above that explains what it does though, if you feel that this should be done, please let me know.

farhatahmad (approved):

PR #5313, opened 2023-07-06
Translate "'s Room" in default room

### LINKED ISSUE #5255: V3: Translate default room
image

It would be nice if there was a translation string to change the caption to "Marc Kohaupts Raum".

Mariam05:
Closes #5255

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -63,7 +63,7 @@ class ApplicationController < ActionController::Base
     return unless user.rooms.count <= 0
     return unless PermissionsChecker.new(permission_names: 'CreateRoom', user_id: user.id, current_user: user, current_provider:).call
 
-    Room.create(name: "#{user.name}'s Room", user_id: user.id)
+    Room.create(name: t('room.new_room_name', username: user.name, locale: user.language), user_id: user.id)
   end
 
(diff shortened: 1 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -63,7 +63,8 @@ def create_default_room(user)
     return unless user.rooms.count <= 0
     return unless PermissionsChecker.new(permission_names: 'CreateRoom', user_id: user.id, current_user: user, current_provider:).call
 
-    Room.create(name: "#{user.name}'s Room", user_id: user.id)
+    locale = I18n.default_locale unless valid_locale?(user.language.to_sym)
+    Room.create(name: t('room.new_room_name', username: user.name, locale:), user_id: user.id)
```
farhatahmad, line 67: I believe that we can assume the user's language is always valid - why the extra check here?

Mariam05: I added the check because the tests were failing due to the user's language being set to "language". I don't know whether or not it's possible for a user's language to exist without a corresponding locale file, but if not I can also change the tests to pass a valid language

farhatahmad: Yeah I would update the tests instead

farhatahmad (approved):

PR #5325, opened 2023-07-11
Various improvements throughout the app

farhatahmad:
Improvements: 

1. Hide recording formats and copy button if recording is unpublished
2. Hide statistics format from public recordings
3. Fix admin create task to display error
4. Replaced recording visibility dropdown with a nicer one

image

### DIFF
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -441,6 +441,29 @@ input.search-bar {
   }
 }
 
+.simple-select {
+  button {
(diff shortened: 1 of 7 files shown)


farhatahmad:
@scouillard I can't reporduce that - which page does it happen on?

scouillard:
> @scouillard I can't reporduce that - which page does it happen on?

Room Recordings

scouillard:
There is an overflow issue with the dropdown
image

scouillard (approved):

PR #5327, opened 2023-07-11
Improvements for handling missed "meeting_ended" and "recording_ready" callbacks

scouillard:
This PR addresses an issue where Recordings and Room status would be out of sync with the BBB server when the "meeting_ended" and/or "recording_ready" callbacks miss.

Two rake tasks are now introduced, `meetings_poller` and `recordings_poller`, which periodically check the status of meetings and recordings. 

`meetings_poller` reaches the BBB server to confirm every meeting where `online: true` are truly online. This verification provides an additional layer of assurance and data consistency in case we fail to receive or process the "meeting_ended" callbacks. It is also needed for the task above.

`recordings_poller` reaches the BBB server to fetch the recordings for every meeting that recently ended. If any recordings are not present in our system but exist on the server, they are added to our database, ensuring we maintain an up-to-date list of recordings even if we miss the "recordings_ready" callbacks.

**The Trigger**
A new service named `greenlight-v3-poller` has been introduced in our docker-compose setup. This service is essentially a clone of greenlight-v3, designed specifically to run these polling tasks.

The entry point for this service is set to bin/poller that is a script that triggers the rake tasks when the application is built.

**NOTE: The trigger has not been tested yet.**


docker-compose.yml
```diff
@@ -33,3 +33,11 @@ services:
     depends_on:
       - postgres
       - redis
+
+  greenlight-v3-poller:
+    entrypoint: [bin/poller]
+    image: bigbluebutton/greenlight:v3
+    env_file: .env
+    depends_on:
+      - postgres
```
farhatahmad, line 42: I would make sure `redis` is up and running before as well

lib/tasks/poller.rake
```diff
@@ -0,0 +1,84 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public License along
+# with Greenlight; if not, see <http://www.gnu.org/licenses/>.
+
+# frozen_string_literal: true
+
+namespace :poller do
+  # Does a check if a meeting set as online is still online
+  task :meetings_poller, %i[provider] => :environment do |_task, args|
+    args.with_defaults(provider: 'greenlight')
```
farhatahmad, line 22: We don't need provider scoping for these tasks

lib/tasks/poller.rake
```diff
@@ -0,0 +1,84 @@
...
+
+    RunningMeetingChecker.new(rooms: online_meetings, provider: args[:provider]).call
+
+  rescue StandardError => e
+    err "Unable to poll meetings. Error: #{e}"
+  end
+
+  # Does a check if a recording from a recent meeting has not been created in GL
+  task :recordings_poller, %i[provider] => :environment do |_task, args|
+    args.with_defaults(provider: 'greenlight')
```
farhatahmad, line 35: Same as above

lib/tasks/poller.rake
```diff
@@ -0,0 +1,84 @@
...
+      record_enabled = RoomsConfiguration.joins(:meeting_option).find_by(provider: args[:provider], 'meeting_option.name': 'record').value
+      if record_enabled
+        begin
+          Rake::Task['poller:recordings_poller'].execute
+        rescue StandardError => e
+          err "An error occurred in recordings poller: #{e.message}. Continuing..."
+        end
+      end
+
+      sleep args[:interval].to_i
```
farhatahmad, line 81: Check the scalelite tasks, theres something with re-enabling the tasks that you have to do at the end

lib/tasks/poller.rake
```diff
@@ -0,0 +1,84 @@
...
+  # Does a check if a recording from a recent meeting has not been created in GL
+  task :recordings_poller, %i[provider] => :environment do |_task, args|
+    args.with_defaults(provider: 'greenlight')
+
+    recent_meetings = Room.joins(:user)
+                          .where(users: { provider: args[:provider] }, last_session: 1.hour.ago..Time.zone.now, online: false)
+
+    # Fetch all rooms associated with the recent meetings in advance.
+    recent_rooms = Room.where(meeting_id: recent_meetings.pluck(:meeting_id))
+                       .index_by(&:meeting_id)
```
farhatahmad, line 42: I'm not sure I understand what the purpose of this is - you're already fetching the meetings above, why re-fetch them?

lib/tasks/poller.rake
```diff
@@ -0,0 +1,84 @@
...
+
+      rescue StandardError => e
+        err "Unable to poll Recording:\nRecordID: #{recording[:recordID]}\nError: #{e}"
+        next
+      end
+    end
+  end
+
+  task :run_all, %i[provider interval] => :environment do |_task, args|
+    args.with_defaults(provider: 'greenlight', interval: 3600)
```
farhatahmad, line 62: Same as above - I would drop the poller

farhatahmad: In Rails, you can do `30.minutes` or `60.minutes` instead of `3600`

lib/tasks/poller.rake
scouillard: It's to reset the room.processing_recording

farhatahmad: Why can't you just do `meeting.update(recordings_processing....`?

scouillard: Right

scouillard: Thanks I had tunnel vision lol

lib/tasks/poller.rake
scouillard: I've read that `.reenable` is not needed when you use `.execute` instead of `.invoke`
`.execute` should be used when there is no dependencies.
However, I will use the same process as Scalelite for consistency

scouillard:
Since last review:
- remove provider scope, rework `RecordingCreator` service
- add `disabled_recordings` query to sort out providers with `record` disabled in the `recent_meeting` query
- add logs

lib/tasks/poller.rake
```diff
@@ -0,0 +1,76 @@
...
+    RunningMeetingChecker.new(rooms: online_meetings).call
+
+  rescue StandardError => e
+    err "Unable to poll meetings. Error: #{e}"
+  end
+
+  # Does a check if a recording from a recent meeting has not been created in GL
+  task :recordings_poller, %i[interval] => :environment do |_task, args|
+    # Returns the providers which have recordings disabled
+    disabled_recordings = RoomsConfiguration.joins(:meeting_option).where('meeting_option.name': 'record', value: 'false').pluck(:provider)
```
farhatahmad, line 33: Small knitpick - I would keep regular syntax here for consistency `where(meeting_option: { name: 'record' }, value...`

lib/tasks/poller.rake
```diff
@@ -0,0 +1,76 @@
...
+  task :run_all, %i[interval] => :environment do |_task, args|
+    args.with_defaults(interval: 60)
+
+    poller_tasks = %w[poller:meetings_poller poller:recordings_poller]
+
+    loop do
+      poller_tasks.each do |poller_task|
+        info "Running #{poller_task} at #{Time.zone.now}"
+        Rake::Task[poller_task].invoke(args[:interval])
+        Rake::Task[poller_task].reenable
```
farhatahmad, line 67: Can we move the re enable to after the sleep

lib/tasks/poller.rake
```diff
@@ -0,0 +1,76 @@
...
+
+      rescue StandardError => e
+        err "Unable to poll Recording:\nRecordID: #{recording[:recordID]}\nError: #{e}"
+        next
+      end
+    end
+  end
+
+  task :run_all, %i[interval] => :environment do |_task, args|
+    args.with_defaults(interval: 60)
```
farhatahmad, line 59: Sharing the same interval value is normally not the best idea. It's better to run the poller every (example) 30 mins, and then check for rooms that have been started in the last hour

farhatahmad: This ensures that any missed calls get re-tried naturally

scouillard: Would you hardcode the intervals?
Or we can take an optional value for the poller interval, and double/increase that value for the rooms last_meeting interval

docker-compose.yml
```diff
@@ -8,8 +8,9 @@ services:
     volumes:
       - ./data/postgres/14/database_data:/var/lib/postgresql/data
     environment:
+      - POSTGRES_DB=postgres
       - POSTGRES_USER=postgres
-      - POSTGRES_PASSWORD=
+      - POSTGRES_PASSWORD=password
```
farhatahmad, line 13: This should stay empty. The script generates a random secure password

docker-compose.yml
```diff
@@ -20,10 +21,12 @@ services:
 
   greenlight-v3:
     entrypoint: [bin/start]
-    image: bigbluebutton/greenlight:v3
+    image: gl-job-tasks
```
farhatahmad, line 24: These changes need to be reverted

lib/tasks/poller.rake
farhatahmad: Optional value and then  double it

lib/tasks/poller.rake
gitface4: It's cleaner inside the iteration. Does this has any impact?

farhatahmad (approved):

PR #5403, opened 2023-09-07
Fixing missing default room name for German

jacotec:
This fixes a missing default room name for a new user when the default language is set to German.


farhatahmad:
Fixed here: https://github.com/bigbluebutton/greenlight/pull/5378

Issue #5409, opened 2023-09-13
`bin/start`: line 3: `config.env`: No such file or directory


debuglevel:
https://github.com/bigbluebutton/greenlight/blob/f739387a304b7d8f8d28b5cf5d96e801f7f60546/bin/start#L3

I guess this should be `bin/config.env` as the working directory is the parent.

PR #5410, opened 2023-09-13
fix: config.env not found

debuglevel:
`bin/start` should most likely read `bin/config.env` instead of `config.env` as the working directory is probably *not* `./bin` but `.`


debuglevel:
Fixes #5409

Issue #5062, opened 2023-03-23
[v3.0.1] LDAP / Active Directory / Kerberos auth, misleading GUI


motsiw:
_System:_
BBB 2.6.0 / GL 3.0.1 with external KeyCloak and User Federation against our Active Directory using Kerberos authentication. 'Require users to be signed in before joining' is 'Force enabled', so no anonymous users in the system.

1. The button 'Sign up' on the top of the GL screen is obsolete (users are always added to the Active Directory another way), but I can't remove this button. Is there a way, a hidden .env parameter or similar, or could this be implemented?

2. Using Kerberos authentication with KC is very handy, because from the user's perspective there's no need to login manually or entering any credentials. It would be extremely useful, if there was some kind of 'auto login' parameter, so anyone accessing a GL page not being authenticated would automatically be so via KC. At the moment, people have to manually click either on 'Sign in' or 'Sign up', which is annoying and misleading to many users.

TIA

dmtrhfr:
Up this - it's confusing that there is a sign up button, when there is another oidc-provider. Either make it hideable or disable it by default, if a keycloak is used.

dmtrhfr:
Up! Quiet misleading

sheyk87:
> 

I agree to the questions!
- How to disable/hide the Sing Up Button?
- How to configure Active Directory auth?

PR #5431, opened 2023-10-02
Remove Sign Up button when using OIDC

jdr00t:
Solves Part 1 of #5062
Sign Up button is not needed when using OpenID Connect


antobinary:
Thank you for sending us a CLA too!

farhatahmad:
Funny story - Greenlight v2 used to do this (maybe 3 years ago) and then people complained that users didn't understand that `Sign in` and `Sign up` for external accounts was the same thing - leading to confusion (ie "where is the signup button")

Hence, the `Sign up` button was added to solve this issue

Ithanil:
@jdr00t Suggestion: Remove the Sign-Up Button also from the RequireAuthentication card in app/javascript/components/rooms/room/join/RequireAuthentication.jsx

farhatahmad (approved):

farhatahmad:
closes #5517

PR #5436, opened 2023-10-03
fix logger, so it can log into default production.log

galupa:
Hi,

I found out that if we not set up `RAILS_LOG_TO_STDOUT` and (`RAILS_LOG_REMOTE_NAME` or `RAILS_LOG_REMOTE_PORT`) env variable (e.q. we want to use default log to log into production.log file), Greenlight will raise error because the `logger` on  line 151 (`logger.formatter = config.log_formatter`) is not initialised.

So this PR basically to fix it. I just move it inside the if statement.

Best Regards,


farhatahmad:
Fixed here: https://github.com/bigbluebutton/greenlight/pull/5520

PR #5449, opened 2023-10-11
Fix issue with recording resync deleting recordings

farhatahmad:


app/services/recording_creator.rb
```diff
@@ -68,7 +68,7 @@ def get_recording_length(recording:)
     length = 0
     if recording[:playback][:format].is_a?(Array)
       recording[:playback][:format].each do |formats|
-        length = formats[:length] if formats[:type] == 'presentation'
+        length = formats[:length] if formats[:type] == 'presentation' || formats[:type] == 'video'
```
Ithanil, line 71: Is it necessary to use specific formats here? Besides presentation or video it could also be podcast, notes, screenshare or potentially a custom format.

Ithanil: Actually, nevermind. The other formats probably don't provide a length value? I'm also wondering what is it even used for?

Issue #5459, opened 2023-10-12
Bug: Protected recording links lead to recording formats being added repeatedly on resync or recording ready notification


Ithanil:
Because the recording URL is used in together with the recording format name in calls of Format.find_or_create_by, a change in the format URL that is provided by BBB or SL (e.g. protected recordings!) leads to the addition of new format entries for formats that are already present with a different URL, whenever a recording resync or repeated recording ready notification happens.

PR #5460, opened 2023-10-12
Update changed recording URLs of already present formats instead of creating new format entries

Ithanil:
Currently, a change in the format URL that is provided by BBB or SL (e.g. protected recordings!) leads to the addition of new format entries for formats that are already present with a different URL, whenever a recording resync or repeated recording ready notification happens. Instead of creating new entries, the URLs of present entries should be updated.

This PR implements that by using only the recording id & format name in the Format.find_or_create_by calls and then updates the recording URL afterwards on the found or created Format. 

Fixes https://github.com/bigbluebutton/greenlight/issues/5459

### DIFF
--- a/app/services/recording_creator.rb
+++ b/app/services/recording_creator.rb
@@ -80,11 +80,11 @@ class RecordingCreator
   def create_formats(recording:, new_recording:)
     if recording[:playback][:format].is_a?(Array)
       recording[:playback][:format].each do |format|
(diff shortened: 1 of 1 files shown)


farhatahmad (approved):

farhatahmad:
I've seen this happen once before and was never able to figure out why it happened - thanks for the fix

Ithanil:
Thanks for merging :+1:

PR #5475, opened 2023-10-20
Support for Redis Sentinel

Ithanil:
This PR enables the use of Redis Sentinel Clusters with GL3. We are using this feature in production, but it was not developed with the goal to be merged upstream. However, I'd like to provide our development to the public for anyone in need of this and would be happy to help getting this ready for merging if there is interest.

This is the beginning of a series of similar PR submissions (Redis Sentinel, SAML integration, LDAP integration).

Details / How to use:
- See sample.env for some documention on configuration
- Provide Redis Sentinel hostnames & ports in the variable REDIS_SENTINEL
- If REDIS_SENTINEL is set, REDIS_URL will be redis://mymaster per default

### DIFF
--- a/app/controllers/health_checks_controller.rb
+++ b/app/controllers/health_checks_controller.rb
@@ -45,7 +45,11 @@ class HealthChecksController < ApplicationController
   end
 
   def check_redis
-    Redis.new.ping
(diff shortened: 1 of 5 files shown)


ffdixon:
Thank you for sharing this @Ithanil.

PR #5476, opened 2023-10-20
Implement SAML authentication for GL3

Ithanil:
Description WIP (missing detailed explanation)

This PR enables SAML authentication for GL3 and is based on similar PRs for GL2: https://github.com/bigbluebutton/greenlight/pull/1334 and https://github.com/bigbluebutton/greenlight/pull/2270 . We are using this feature in production, but it was not developed with the goal to be merged upstream. Because SAML for GL2 was never merged, I guess it won't happen for GL3 as well. However, I'd like to provide our development to the public for anyone in need of this.

This is the part of a series of similar PR submissions (Redis Sentinel, SAML integration, LDAP integration).

PR #5480, opened 2023-10-23
Add env EXTERNAL_AUTH and replace checks for OPENID_CONNECT with checks for EXTERNAL_AUTH where applicable

Ithanil:
Since OIDC is currently the only implemented external authentication provider, in all the places where you would normally check for external auth usage there is instead a check for OIDC usage. This PR introduces a new env EXTERNAL_AUTH and replaces these checks with a check for EXTERNAL_AUTH. Also a few comments are left at places where additional providers would have to be "added".

This paves the way for addition of other external auth providers and, even if there are no plans to develop these from upstream side, makes it easier to develop and maintain extensions such as https://github.com/bigbluebutton/greenlight/pull/5476 . Also, it will not require any additional effort to maintain the changes made in this PR and it does not change the current behavior of GL3. 

If any changes have to be made to make this mergeable, I'm happy to work on them. Please let me know!

### DIFF
--- a/app/controllers/api/v1/env_controller.rb
+++ b/app/controllers/api/v1/env_controller.rb
@@ -25,7 +25,6 @@ module Api
       # Returns basic NON-CONFIDENTIAL information on the environment variables
       def index
         render_data data: {
-          OPENID_CONNECT: ENV['OPENID_CONNECT_ISSUER'].present?,
           EXTERNAL_AUTH: ENV['OPENID_CONNECT_ISSUER'].present?, # currently only OIDC is implemented
           HCAPTCHA_KEY: ENV.fetch('HCAPTCHA_SITE_KEY', nil),
           VERSION_TAG: ENV.fetch('VERSION_TAG', ''),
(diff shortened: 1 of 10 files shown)


app/controllers/api/v1/env_controller.rb
```diff
@@ -26,6 +26,7 @@ class EnvController < ApiController
       def index
         render_data data: {
           OPENID_CONNECT: ENV['OPENID_CONNECT_ISSUER'].present?,
```
farhatahmad, line 28: We can remove this now no?

farhatahmad:
Great idea for this PR and definitely makes it easier for us to manage - thanks for your hard work

app/controllers/api/v1/env_controller.rb
Ithanil: Currently it would be unused, yes. Not sure if there could be a future use for this though.

farhatahmad: I would remove it then if possible please

Ithanil: Done.

farhatahmad (approved):

Issue #5321, opened 2023-07-10
Reset password does not work because CAPS


lonesomewalker:
Since the users are not able to recover their password, we had to debug a bit.

Seems that Greenlight now is case-sensitive for email addresses.
This is a bit problematic... :-)

lonesomewalker:
Register user with: abC@example.org
Try to login or recover password for aBc@example.org

All email addresses should be stored lowercase, because everything else makes a) no sense and violates the typical approach how to store email addresses and b) creates unneccessary bugs ;-)

PR #5485, opened 2023-10-30
Fix case sensitive emails 

### LINKED ISSUE #5405: Gl 3.0.6.1: Login/EMail case sensitive
Hi,
I've just installed BBB with Gl 3.0.6.1 and added several users with a script("docker exec greenlight-v3 bundle exec rake user:create[...]").
Each emailadress/login is in lower case.
Login with emailadress in 3.0.6.1 is case sensitive(!), so everyone must login with lower case adress.

In the old Gl-version (2.x) login with emailadress is case INsensitive. (fixed in 2.5.5 see issue #1494).
This should be fixed also in 3.x

Thanks
Holger

farhatahmad:
fixes #5405


lonesomewalker:
And fixes #5321

Issue #5417, opened 2023-09-14
Separate password reset mail for `migration:reset_password_email`


debuglevel:
There is a `migration:reset_password_email` task. IMHO it does not make much sense:

* The user has absolutely no idea why they received this mail. -> mention that Greenlight was updated to a new version and all users have to supply a new password.
* The link is valid for 1 hour. I guess no even 5% of my users will accidentally check their mail and be able to click on this link in time. -> create a token that is valid for at least 7 days.

PR #5507, opened 2023-11-10
feat(migrations): Import password digests from Greenlight2

danimo:
This is a proof-of-concept. It works under the following provision that `SECRET_KEY_BASE` is reused from Greenlight2 (keeps the salt stable). Pushing this as a draft for subject to discission. Requires the following patch on the Greenlight2-side:

```patch
diff --git a/lib/tasks/migrations/migrations.rake b/lib/tasks/migrations/migrations.rake
index 8450e664..98f451ac 100755
--- a/lib/tasks/migrations/migrations.rake
+++ b/lib/tasks/migrations/migrations.rake
@@ -69,7 +69,7 @@ namespace :migrations do

     user = User.unscoped
     user = user.where(provider: args[:provider]) if args[:provider].present?
-    user = user.select(:id, :uid, :name, :email, :social_uid, :language, :role_id, :created_at)
+    user = user.select(:id, :uid, :name, :email, :password_digest, :social_uid, :language, :role_id, :created_at)
                .includes(:role)
                .where.not(roles: { name: COMMON[:filtered_user_roles] }, deleted: true)

@@ -78,6 +78,7 @@ namespace :migrations do
       params = { user:
                    { name: u.name,
                      email: u.email,
+                     password_digest: u.password_digest,
                      external_id: u.social_uid,
                      provider: u.provider,
                      language: u.language,
```
The risk seems rather low, as in the worst case (wrong hash) the user will be without a known password just like it is the case in the current code base already. Probably the documentation should be adjusted to suggest keeping the `SECRET_KEY_BASE`.


debuglevel:
It would be great if users would not had to change their password! If this gets accepted, we can maybe also close #5417.

SamuelWei:
As mentioned to @danimo, this could even be independent of the `SECRET_KEY_BASE', which makes it even simpler.
According to my research, the  `SECRET_KEY_BASE' only affects encrypted cookies, signed cookies, Active Storage files (I guess if they are encrypted), messages, etc. The hasing using by rail is bycrypt, therefor there is no separate salt. The salt is part of the hash.

danimo:
And it might not be used as pepper either. So it would be much simpler indeed.

farhatahmad:
I'm honestly back and forth on this. Sending the password over the networks could lead to major security breaches if the SECRET_KEY_BASE is compromised.

What if we change the login flow to allow the users to set a new password on first login instead? That way we get the best of both solutions?

farhatahmad:
If the SECRET_KEY_BASE is compromised, the attackers could decrypt the params passed (since they are just encrypted using the SECRET_KEY_BASE as the key), and then potentially brute force the passwords.

When it comes to passwords, I prefer staying extra safe. 

> What if we change the login flow to allow the users to set a new password on first login instead? That way we get the best of both solutions?

Thoughts on this?

debuglevel:
But SECRET_KEY_BASE is just a environment variable on my machine which is never transferred, isn't it? So in order to know my SECRET_KEY_BASE, someone needs access to my machine. And in this case I've got probably some serious other problems.

danimo:
This sounds like a scoping problem. The way I understand it, you are hesitant to implement migrating the salted password hashes, because the transfer of these could be insecure. At a first glance, that makes sense; if salted password hashes leak, they could attacked using brute-force or rainbow tables. However, since you're using bcrypt, and a salt, this is no longer easily viable.

An administrator performing a migration has access to these hashes already, or could even just log the passwords when they are transmitted to the webserver when a user logs in.

As such, it's the administrator's responsibility when performing a migration to transfer the data securely, and set up the webserver securely. The best thing the Greenlight project can do here is to help administrators understand the risks and make an informed decision.

If e.g. HTTPS isn't enabled on the destination, the passwords would be transferred in plaintext whenever a user logs in. If an administrator decides to perform a migration without TLS, that would incur a risk, but might -- depending on e.g. the network setup -- be acceptable. As a developer, you can't tell the specific setup that's being used.
One thing Greenlight could do here is provide a warning when migrating over a plaintext connection.

OTOH, forcing users to re-set their passwords incurs a pretty significant security risk. Password change fatigue causes users to change their passwords in predictable and insecure ways, use simpler password, and makes them more vulnerable to phishing. This is even true when enforcing password "complexity" requirements.

defnull:
So we have two options:
* Transferring properly salted password hashes, over HTTPS, with an additional encryption layer, just once, while the admin is actively participating and knows exactly what happens.
* Or sending out thousands of password reset mails that are only valid for one hour, and simply hope that all those users check their mails and set a new password within that time frame, instead of kicking in our office door with pitchforks in their hands and telling us how great BBB works for them in the middle of a semester.

defnull:
> What if we change the login flow to allow the users to set a new password on first login instead? That way we get the best of both solutions?

Do you mean, migrating and accepting the old password, but force users to set a new one on first login? That would be a nice functionality if implemented properly, but unnecessary I guess. Forcing password changes is no longer considered good practice and the transfer was as secure at it gets, both transport (HTTPS) and e2e (AES) encrypted.

EmmyGraugans:
> Do you mean, migrating and accepting the old password, 

Or setting the password to <blank value> and telling the user kn login-attempt: due to migration to a new system, your old password is no longer valid, a token for setting a new password was sent to your email-address. (basically, trigger the „forgot password“ workflow the first time after migration)

defnull:
> Or setting the password to [blank value] and telling the user on login-attempt: due to migration to a new system, your old password is no longer valid, a token for setting a new password was sent to your email-address. (basically, trigger the „forgot password“ workflow the first time after migration)

Also a sensible idea. But I still don't get why the old (hashed) password should be considered any less secure, just because it was moved (over a double-encrypted tightly controlled channel) to a new server.

farhatahmad:
Fair points - I'll do some tests on this to make sure it works as expected and if so, I can update the docs and merge this in

farhatahmad:
I can confirm this works as expected - working on the v2 PR and will make some comments on this

farhatahmad:
(I'll fix the tests in another repo)

PR #5510, opened 2023-11-13
added delete presentation string closes #5396

### LINKED ISSUE #5396: Wrong title for presentation deletion
When deleting a presentation via the bin icon, the form DeletePresentationForm.jsx is called. The title set for the form however is recording.delete_recording (instead of something presentation related). As far as I can see there is no language string for delete_presentation as of now.

Tested with 3.0.5

SilentFlameCR:


SilentFlameCR:
closes #5396

farhatahmad (approved):

PR #5574, opened 2023-11-23
User redirected to JoinCard for room after SSO login for joinRoom

### LINKED ISSUE #5516: [v3] User not redirected to room after SSO login.
Room that "Require users to be signed in before joining" will display a "Sign Up / Sign In" dialog to unauthenticated users. Clicking on "Sign In" will redirect the user to the OIDC identity provider, but after a successful login, the user will return to the home page instead of the room. The information for which room the user was supposed to login is lost in the process. The `location` cookie is checked, but never set. 

To reproduce:
* Create a room.
* Visit the "/rooms/xxx-xxx-xxx-xxx/join" page in a private browser tab.
* Click on "Sign In" and sign in.
* Notice you are redirected to the home page and not the room join page.

SilentFlameCR:
fixes #5516

### DIFF
--- a/config/locales/tr.yml
+++ b/config/locales/tr.yml
@@ -47,23 +47,23 @@
 
 tr:
   opengraph:
-    description: "Sorunsuz sanal işbirliği ve çevrim içi öğrenme deneyimleri sağlayan, güvenilir ve açık kaynaklı internet üzerinden görüşme çözümü BigBlueButton uygulamasını kullanarak öğrenin."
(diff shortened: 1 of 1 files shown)


app/javascript/components/users/authentication/forms/SigninForm.jsx
```diff
@@ -38,10 +38,24 @@ export default function SigninForm() {
...
+    const roomUrl = new URLSearchParams(location.search).get('location');
+    if (roomUrl) {
+      // Redirect the user to the joinCard for the room
+      navigate(roomUrl);
+    } else {
+      // Redirect the user to the rooms homepage
+      navigate('/rooms');
+    }
+
     return createSessionAPI.mutate({ session, token });
```
farhatahmad, line 59: Look into the `useCreateSession` mutation - that's where the current redirect happens

PR #5587, opened 2023-11-27
added user rake task set_admin_role

### LINKED ISSUE #5515: [v3] No practical way to create/use (super) admin accounts with SSO enabled.
After enabling OpenID Connect there is no obvious way to login with a local account (e.g. the local admin account).

SilentFlameCR:
Fixes #5515

### DIFF
--- a/lib/tasks/user.rake
+++ b/lib/tasks/user.rake
@@ -60,8 +60,7 @@ namespace :user do
     # return err if Administrator role not found
     err "Role 'Administrator' not found for provider 'greenlight'" if role.blank?
 
-    user.role = role
-    user.save!
+    user.update(role:)
     success "User role set to Administrator for email: #{email}"
   end
 
--- a/lib/tasks/user.rake
+++ b/lib/tasks/user.rake
@@ -47,23 +47,22 @@ namespace :user do
   task :set_admin_role, %i[email] => :environment do |_task, args|
     email = args[:email]
 
-    if email.blank?
-      info 'Please provide an email address of the user you wish to set to Administrator role.'
-    else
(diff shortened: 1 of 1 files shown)


lib/tasks/user.rake
```diff
@@ -43,6 +43,29 @@ namespace :user do
...
+          user.save!
+          success "User role set to Administrator for email: #{email}"
+        else
+          info "Role 'Administrator' not found for provider 'greenlight'"
+        end
+      else
+        info "User with email: #{email} not found"
+      end
+    end
+  end
```
farhatahmad, line 67: The nested ifs make this super difficult to follow. Might be clear to use guard clauses here instead 

```
user = User.find_by(email:, provider: 'greenlight')
err 'User with email: #{email} not found' if user.blank?
...
```

SilentFlameCR: Yeah that sounds good. I updated the code with the guard clauses so its easier to read!

lib/tasks/user.rake
```diff
@@ -43,6 +43,28 @@ namespace :user do
...
+    # return err if user not found
+    err "User with email: #{email} not found" if user.blank?
+
+    role = Role.find_by(name: 'Administrator', provider: 'greenlight')
+
+    # return err if Administrator role not found
+    err "Role 'Administrator' not found for provider 'greenlight'" if role.blank?
+
+    user.role = role
+    user.save!
```
farhatahmad, line 64: You can do this in one line `user.update(role:)`

SilentFlameCR: got it and updated!

farhatahmad (approved):

PR #5591, opened 2023-11-30
Fix site settings not updating

farhatahmad:


farhatahmad:
Thanks for the heads up @ahmadnofal1993 - should be fixed once 3.1.0.2 is released

PR #5614, opened 2023-12-13
Limit recording visibility by Role

farhatahmad:
Added the ability to restrict Recording Visibilities based on role permissions. By default, all roles have access to all visibilities

image


farhatahmad:
(I'll come back to the eslint issues in a follow up PR)

PR #5635, opened 2023-12-21
Added copy buttons for room access codes

### LINKED ISSUE #5578: Button for Copy access code
Like in V2, there should be a button for copying the access code and not only for copying the join link, after selecting a room.

SilentFlameCR:
Fixes #5578 

image
image

### DIFF
--- a/app/javascript/components/rooms/room/Room.jsx
+++ b/app/javascript/components/rooms/room/Room.jsx
@@ -49,12 +49,12 @@ export default function Room() {
     if (role === 'viewer') {
       navigator.clipboard.writeText(roomSettings?.data?.glViewerAccessCode);
       toast.success(t('toast.success.room.copied_viewer_code'));
-      return
+      return;
     }
     if (role === 'moderator') {
       navigator.clipboard.writeText(roomSettings?.data?.glModeratorAccessCode);
       toast.success(t('toast.success.room.copied_moderator_code'));
-      return
+      return;
     }
     navigator.clipboard.writeText(`${window.location}/join`);
     toast.success(t('toast.success.room.copied_meeting_url'));
--- a/app/javascript/components/rooms/room/Room.jsx
+++ b/app/javascript/components/rooms/room/Room.jsx
@@ -49,15 +49,13 @@ export default function Room() {
     if (role === 'viewer') {
       navigator.clipboard.writeText(roomSettings?.data?.glViewerAccessCode);
       toast.success(t('toast.success.room.copied_viewer_code'));
-      return;
-    }
-    if (role === 'moderator') {
+    } else if (role === 'moderator') {
       navigator.clipboard.writeText(roomSettings?.data?.glModeratorAccessCode);
       toast.success(t('toast.success.room.copied_moderator_code'));
-      return;
+    } else {
+      navigator.clipboard.writeText(`${window.location}/join`);
+      toast.success(t('toast.success.room.copied_meeting_url'));
     }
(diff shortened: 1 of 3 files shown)


farhatahmad:
Can we try changing the color of them to differentiate them from the copy room url button?

app/javascript/components/rooms/room/Room.jsx
```diff
@@ -42,6 +43,18 @@ export default function Room() {
...
+  function copyAccessCode(role) {
+    if (role === 'viewer') {
+      navigator.clipboard.writeText(roomSettings?.data?.glViewerAccessCode);
+      toast.success(t('toast.success.room.copied_viewer_code'));
+    }
+    if (role === 'moderator') {
+      navigator.clipboard.writeText(roomSettings?.data?.glModeratorAccessCode);
+      toast.success(t('toast.success.room.copied_moderator_code'));
+    }
+  }
```
farhatahmad, line 57: Just to clean this up - can it be merged in with the `copyInvite` method below it?

app/javascript/components/rooms/room/Room.jsx
```diff
@@ -85,18 +98,50 @@ export default function Room() {
...
-              ) : (
-                t('room.meeting.start_meeting')
-              )}
-            </Button>
-            <Button variant="brand-outline" className="mt-1 mx-2 float-end" onClick={() => copyInvite()}>
-              <Square2StackIcon className="hi-s me-1" />
-              { t('copy') }
-            </Button>
+            <Row>
+              <Col className="col-12">
```
farhatahmad, line 102: Is the extra row and column needed here? Considering it's already wrapped in a column

app/javascript/components/rooms/room/Room.jsx
SilentFlameCR: yup cleanup and merged!

app/javascript/components/rooms/room/Room.jsx
SilentFlameCR: Yeah without it looks like this. screenshot added.

image

app/javascript/components/rooms/room/Room.jsx
```diff
@@ -42,8 +43,19 @@ export default function Room() {
...
+      toast.success(t('toast.success.room.copied_viewer_code'));
+      return;
+    }
+    if (role === 'moderator') {
+      navigator.clipboard.writeText(roomSettings?.data?.glModeratorAccessCode);
+      toast.success(t('toast.success.room.copied_moderator_code'));
+      return;
+    }
     navigator.clipboard.writeText(`${window.location}/join`);
     toast.success(t('toast.success.room.copied_meeting_url'));
```
farhatahmad, line 60: Why not just do an `if else if` here? Rather than add the if + return

app/javascript/components/rooms/room/Room.jsx
farhatahmad: Try using `Stack` instead

app/javascript/components/rooms/room/Room.jsx
SilentFlameCR: fixed!

app/javascript/components/rooms/room/Room.jsx
SilentFlameCR: I tried with a Stack but it still was same as screenshot above...tried with a Stack inside the Row but same again

farhatahmad (approved):

Issue #5650, opened 2024-01-08
[v3] Session Cookie does not limited to a path when RELATIVE_URL_ROOT is set


wolbernd:
We plan to deploy multiple Greenlight instances behind a reverse proxy. All those instances should be reachable under the same subdomain but different paths. We do this by setting RELATIVE_URL_ROOT in the environemnt file for the respective greenlight container. 

This leads to the problem that both instances create cookies with the name `_greenlight-3_0_session` (hardcoded in https://github.com/bigbluebutton/greenlight/blob/master/config/initializers/session_store.rb) and the domain of the server.  
A user wanting to use both greenlight instances on the server (which is a valid use case in our scenario) will have to use two browsers or a private tab to do so since the greenlight instances will always overwrite the session cookie of the other instance since they can't be differentiated. 

### Example Setup:
* Domain: ourbbb.domain.tld
* URL for Greenlight instance 1: https://ourbbb.domain.tld/gl_one (`RELATIVE_URL_ROOT=gl_one`)
* URL for Greenlight instance 2: https://ourbbb.domain.tld/gl_two (`RELATIVE_URL_ROOT=gl_two`)

When a user opens https://ourbbb.domain.tld/gl_one a session cookie with `domain: ourbbb.domain.tld, path: /` is created. This same cookie will also be used by https://ourbbb.domain.tld/gl_two (or rather will be discarded, because instance 2 has a different secret key etc)

### Possible solution

A possible solution to this problem would be to also set a path attribute in the cookie, when RELATIVE_URL_ROOT is set. This way there can be multiple cookies for different paths but the same domain.  I will provide a PR in a few minutes with an implementation that does that. 

I concede that this use-case might not be very common, however the proposed solution in the PR should not have any negative side effects on more "normal" Greenlight deployments. Also limiting the visibility of cookies to the intended path of the application should also be a good thing security wise.

PR #5651, opened 2024-01-08
Set path attribute in session cookie when relative url is set

wolbernd:
This PR sets an explicit path to the session cookie which allows the use of multiple Greenlight instances on the same domain. 

Detailed description in issue #5650


farhatahmad (approved):

farhatahmad:
Thanks for the quick PR - will get this in before the next release

PR #5664, opened 2024-01-17
added checks for confirm terms and email notifications

SilentFlameCR:
Screenshots of new sign up page additions and page existing users that have not confirmed terms will see if the site setting is enabled

image
image

### DIFF
--- a/app/assets/locales/en.json
+++ b/app/assets/locales/en.json
@@ -286,7 +286,11 @@
         "allow_users_to_preupload_presentation": "Allow Users to Preupload Presentations",
         "allow_users_to_preupload_presentation_description": "Users can preupload a presentation to be used as the default presentation for that specific room",
         "default_visibility": "Default Recording Visibility",
-        "default_visibility_description": "All newly created recordings will have this visibility by default"
+        "default_visibility_description": "All newly created recordings will have this visibility by default",
+        "confirm_terms": "Confirm terms and conditions",
+        "confirm_terms_description": "Require users to confirm terms and conditions",
+        "disable": "Disable",
+        "enable": "Enable"
       },
       "registration": {
         "registration": "Registration",
@@ -401,7 +405,8 @@
         "brand_image_deleted": "The brand image has been deleted.",
         "privacy_policy_updated": "The privacy policy has been updated.",
         "helpcenter_updated": "The help center link has been updated.",
-        "terms_of_service_updated": "The terms of service have been updated."
+        "terms_of_service_updated": "The terms of service have been updated.",
+        "confirm_terms_updated": "The confirmation for terms of service have been updated."
       },
       "recording": {
         "recording_visibility_updated": "The recording visibility has been updated.",
@@ -460,6 +465,12 @@
     "resend_activation_link": "If you haven't received an activation email or if you are having a problem using it, click on the button below to request a new activation email.",
     "resend_btn_lbl": "Resend Verification"
   },
+  "confirm_terms_page": {
+    "title": "Confirm terms and conditions",
+    "account_unconfirmed": "Your account has not accepted the terms and conditions.",
+    "message": "Please accept the terms and conditions for Greenlight.",
+    "confirm_btn_lbl": "Confirm"
+  },
   "forms": {
     "validations": {
       "full_name": {
@@ -486,6 +497,9 @@
       "password_confirmation": {
         "required": "Please enter a password confirmation",
         "match": "The passwords do not match"
+      },
+      "confirm_terms": {
+        "required": "Please accept the terms and conditions"
       },
       "emails": {
         "required": "Please enter at least one valid email",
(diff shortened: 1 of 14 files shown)


app/javascript/components/users/account_activation/ConfirmTerms.jsx
```diff
@@ -0,0 +1,96 @@
...
+      email_notifs: isCheckedEmails,
+    });
+  };
+
+  return (
+    <div className="vertical-center">
+      <div className="text-center pb-4">
+        <Logo />
+      </div>
+      <Card className="col-md-4 mx-auto p-4 border-0 card-shadow">
```
farhatahmad, line 48: We can increase the width of the box here abit to stop the title from wrapping around

app/serializers/user_serializer.rb
```diff
@@ -30,4 +30,8 @@ def language
   def avatar
     user_avatar(object)
   end
+
+  delegate :confirm_terms, to: :object
+
+  delegate :email_notifs, to: :object
```
farhatahmad, line 36: I don't think these are necessary - the serializer should be able to do this automatically (as it does with the other values (ie name, email)

db/migrate/20240108202126_add_confirm_terms_to_site_settings.rb
```diff
@@ -0,0 +1,20 @@
...
+  def change
+    Setting.create!(name: 'ConfirmTerms') unless Setting.exists?(name: 'ConfirmTerms')
+
+    return if SiteSetting.exists?(setting: Setting.find_by(name: 'ConfirmTerms'))
+
+    SiteSetting.create!(
+      setting: Setting.find_by(name: 'ConfirmTerms'),
+      value: false,
+      provider: 'greenlight'
+    )
```
farhatahmad, line 13: You're doing a data migration inside of a schema migration. They should be seperated into 2 separate migrations

app/serializers/user_serializer.rb
```diff
@@ -19,7 +19,7 @@
 class UserSerializer < ApplicationSerializer
   include Avatarable
 
-  attributes :id, :name, :email, :provider, :language, :avatar, :verified, :created_at
+  attributes :id, :name, :email, :provider, :language, :avatar, :verified, :confirm_terms, :email_notifs, :created_at
```
farhatahmad, line 22: Do we need access to `email_notifs` on the front end?

db/migrate/20240108202126_add_confirm_terms_to_site_settings.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    change_table :users, bulk: true do |t|
+      t.boolean :confirm_terms, default: false
+      t.boolean :email_notifs, default: false
```
farhatahmad, line 17: For simplicities sake, I would just call this `terms` and `marketing`

app/javascript/components/users/account_activation/ConfirmTerms.jsx
```diff
@@ -0,0 +1,96 @@
...
+          <Stack>
+            {t('forms.user.signup.fields.email_notifs.label')}
+          </Stack>
+          <div className="form-switch">
+            <input
+              className="form-check-input fs-5"
+              type="checkbox"
+              checked={isCheckedEmails}
+              onChange={() => setIsCheckedEmails(!isCheckedEmails)}
+            />
```
farhatahmad, line 79: The check boxes should be on the left side of the text

app/assets/locales/en.json
```diff
@@ -460,6 +465,12 @@
     "resend_activation_link": "If you haven't received an activation email or if you are having a problem using it, click on the button below to request a new activation email.",
     "resend_btn_lbl": "Resend Verification"
   },
+  "confirm_terms_page": {
+    "title": "Confirm terms and conditions",
```
farhatahmad, line 469: I think this should `Accept Our Terms and Conditions`

app/assets/locales/en.json
```diff
@@ -460,6 +465,12 @@
...
+    "account_unconfirmed": "Your account has not accepted the terms and conditions.",
+    "message": "Please accept the terms and conditions for Greenlight.",
```
farhatahmad, line 471: It might be better to phrase this a bit more gentle. Also, we don't really refer to the app as Greenlight to the end user since they only know it as BBB.

`To continue using BigBlueButton, please accept our terms and conditions`

app/assets/locales/en.json
```diff
@@ -460,6 +465,12 @@
...
+    "confirm_btn_lbl": "Confirm"
```
farhatahmad, line 472: Is there no plain confirm button that we can reuse from somewhere else?

app/serializers/user_serializer.rb
SilentFlameCR: I just tested yeah your right its not needed and works fine without. removed!

app/serializers/user_serializer.rb
SilentFlameCR: Yes the email_notifs are needed on the ConfirmTerms page as along with option to accept terms they also have a toggle to receive email updates and the initial position of that toggle needs the data from the user. (email_notifs is now marketing).

app/assets/locales/en.json
SilentFlameCR: Couldn't find one so I reused the resend button from the VerifyAccount page for a plain confirm.

PR #5675, opened 2024-01-22
added maintenance banner site setting

### LINKED ISSUE #5446: Feature Request: Bring back the maintenance banner
The maintenance banner from GLv2 was really usefull. It would be great if we could get it too in Version 3.

SilentFlameCR:
Fixes #5446 

# image for users not signed in
image

# image for users signed in
image

### DIFF
--- a/app/javascript/App.jsx
+++ b/app/javascript/App.jsx
@@ -29,6 +29,9 @@ export default function App() {
   const currentUser = useAuth();
   const location = useLocation();
 
+  // check for the maintenance banner
+  const maintenanceBanner = useSiteSetting(['Maintenance']);
+
   // Pages that do not need a header: SignIn, SignUp and JoinMeeting (if the user is not signed in)
   const homePage = location.pathname === '/';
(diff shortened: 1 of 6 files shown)


Ithanil:
Looking forward to get this! But a small suggestion: Maybe it would look a bit nicer if the text was centered.

SilentFlameCR:
> Looking forward to get this! But a small suggestion: Maybe it would look a bit nicer if the text was centered.

text is now centered :)

app/javascript/App.jsx
```diff
@@ -29,6 +29,29 @@ export default function App() {
...
+        hideProgressBar: true,
+        closeOnClick: true,
+        pauseOnHover: true,
+        draggable: true,
+        progress: undefined,
+        theme: 'light',
+        className: 'text-center',
+      });
+    }
+  };
```
farhatahmad, line 48: Do we need the separate function? Can this not just go inside the `useEffect`?

SilentFlameCR: No the separate function is not needed moved it inside useEffect!

PR #5689, opened 2024-01-31
Reimplementing flag meta_bbb-origin-server-name from Greenlight V2

krim404:
This pull request addresses the need for reintroducing the meta_bbb-origin-server-name flag, a feature originally present in Greenlight V2 but missing in versions > 3. 

The flag is crucial for monitoring systems to identify the origin server name at a single access point, allowing for the generation of statistical analyses based on the utilized graphical user interface.


farhatahmad (approved):

Ithanil:
Sorry, I didn't search through the existing PRs before adressing the same issue in: https://github.com/bigbluebutton/greenlight/pull/5705 ... Main difference is that I'm using the uri module to get the hostname. Whether that is an actually more robust solution I honestly don't know.

Issue #5694, opened 2024-02-01
GL2 migration does not migrate existing presentation uploads


danimo:
When importing the room settings, presentations will not be migrated.

Fixing this needs to components:  

* Greenlight3: Allow for providing presentations along with a new room
* Greenlight2: Make use of that feature while migrating

PR #5695, opened 2024-02-01
feat(migrations): Import room presentation attachment (#5694)

danimo:
This is the Greenlight 3 side.

fixes: #5694 (1/2)


danimo:
This has migrated several instances of different sizes successfully, removing Draft state.

farhatahmad:
Thanks for the update - can you update this so that it passes rubocop please? If you're not able to, then I can take a look

danimo:
The force-pushed version should satisfy rubocop.

farhatahmad (approved):

PR #5696, opened 2024-02-01
feat: Implement room presentation migration (#5694)

danimo:
This is the Greenlight 2 side.

fixes: #5694 (2/2)


danimo:
This has migrated several instances of different sizes successfully, removing Draft state.

PR #5705, opened 2024-02-05
Add bbb-origin-server-name to metadata on meeting creation

Ithanil:
Many BBB frontends as well as GL2 (if I remember correctly) include the hostname from the base URL as 'bbb-origin-server-name' in the meeting metadata. This is useful in analytics / statistics to differentiate usage by different frontend deployments.

GL3 does not include the bbb-origin-server-name in metadata, it is only implicitly included in the callback-url. This PR makes GL3 include bbb-origin-server-name in the metadata on meeting creation.

### DIFF
--- a/app/services/meeting_starter.rb
+++ b/app/services/meeting_starter.rb
@@ -16,6 +16,8 @@
 
 # frozen_string_literal: true
 
+require 'uri'
+
 class MeetingStarter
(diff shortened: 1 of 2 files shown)


farhatahmad:
I like this approach better

farhatahmad (approved):

PR #5715, opened 2024-02-20
added changes required for upcoming docker compose

### LINKED ISSUE #5684: Install script uses only old docker-compose command
Hello,

The current greenlight install script needs to be fixed. The `docker-compose` command is no longer available on newer Docker installations. That is because Docker has integrated docker-compose into the standard docker command syntax. It is now possible to run docker-compose commands with the command `docker compose [options]`.

A possible solution would be to check if the legacy `docker-compose` command exists on the system, and if not switch to the new `docker compose [options]` command.

``` bash
docker_compose_cmd="docker-compose"
# Test if docker-compose command exists
if ! command -v $docker_compose_cmd &> /dev/null
then
        # Change the command to the build in docker compose command
        docker_compose_cmd="docker compose"
fi
$docker_compose_cmd ls
```

SilentFlameCR:
Fixes #5684
checking if docker-compose exists on the system if not using new docker compose

### DIFF
--- a/gl-install.sh
+++ b/gl-install.sh
@@ -96,6 +96,14 @@ main() {
   check_ubuntu_lts
   need_x64
 
+  DOCKER_COMPOSE_CMD="docker-compose"
+  # Test if docker-compose command exists
+  if ! command -v $DOCKER_COMPOSE_CMD &> /dev/null
+  then
+          # Change the command to the build in docker compose command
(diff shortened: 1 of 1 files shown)


farhatahmad (approved):

farhatahmad:
We need to make the same change to the bbb install script as well https://github.com/bigbluebutton/bbb-install/blob/v2.7.x-release/bbb-install.sh

PR #5723, opened 2024-02-23
Check external auth session

SilentFlameCR:
session timeout setting only visible if external auth is enabled

### DIFF
--- a/app/javascript/components/admin/site_settings/settings/Settings.jsx
+++ b/app/javascript/components/admin/site_settings/settings/Settings.jsx
@@ -54,7 +54,8 @@ export default function Settings() {
       )}
         value={siteSettings?.PreuploadPresentation}
       />
-      {envAPI.data?.EXTERNAL_AUTH && 
+      { envAPI.data?.EXTERNAL_AUTH
+        && (
         <SettingSelect
           defaultValue={siteSettings?.SessionTimeout}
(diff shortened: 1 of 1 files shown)


app/javascript/components/admin/site_settings/settings/Settings.jsx
```diff
@@ -21,12 +21,14 @@ import useSiteSettings from '../../../../hooks/queries/admin/site_settings/useSi
...
 import SettingSelect from './SettingSelect';
 import useUpdateSiteSetting from '../../../../hooks/mutations/admin/site_settings/useUpdateSiteSetting';
+import useEnv from '../../../../hooks/queries/env/useEnv';
 
 export default function Settings() {
   const { t } = useTranslation();
   const { data: siteSettings, isLoading } = useSiteSettings(['ShareRooms', 'PreuploadPresentation', 'DefaultRecordingVisibility', 'SessionTimeout']);
   const updateDefaultRecordingVisibility = useUpdateSiteSetting('DefaultRecordingVisibility');
   const updateSessionTimeout = useUpdateSiteSetting('SessionTimeout');
+  const envAPI = useEnv();
```
farhatahmad, line 31: You can break this down for simplicity `const { data: env } = useEnv();` so that below you can just do `env?.EXTERNAL_AUTH` (without the data)

farhatahmad (approved):

PR #5740, opened 2024-03-08
All users join moderator no longer allowing anyone to start meeting

### LINKED ISSUE #5714: "All users join as moderators" implies "Allow any user to start this meeting"
**Describe the bug**

Using greenlight v3, creating a room with "All users join as moderators" enabled makes this room accessible to anyone, regardless of what "Allow any user to start this meeting" is set to.

**To Reproduce**

- Create a new room.
- Enable "All users join as moderators"
- Disable "Allow any user to start this meeting"
- Join the room in you browser's private mode, without signing in.

**Expected behavior**

I realize that this issue is ambiguous. My expectation was the following. With  "All users join as moderators" enabled and "Allow any user to start this meeting" disabled, I would have expected have a room that has to be started by the owner/someone who is registered and has the required permissions. Then, once it is started, anyone can join as a moderator. I would have liked this setup for small internal meetings with people whom I trust (so they can all join as moderators), but I still want to have control over when the session is open and when not.

This may be related to #4286

SilentFlameCR:
Fixes #5714 
if glAnyoneJoinAsModerator is true then only the room owner can start the meeting or if glAnyoneCanStart is true anyone can
previously turning on glAnyoneJoinAsModerator allowed anyone to start the meeting

### DIFF
--- a/app/controllers/api/v1/meetings_controller.rb
+++ b/app/controllers/api/v1/meetings_controller.rb
@@ -125,7 +125,7 @@ module Api
 
       def authorized_to_start_meeting?(settings, bbb_role)
         if settings['glAnyoneJoinAsModerator'] == 'true'
-          settings['glAnyoneCanStart'] == 'true' || @room.user_id == current_user&.id
+          settings['glAnyoneCanStart'] == 'true' || @room.user_id == current_user&.id || current_user&.shared_rooms&.include?(@room)
         else
           settings['glAnyoneCanStart'] == 'true' || bbb_role == 'Moderator'
         end
--- a/app/controllers/api/v1/meetings_controller.rb
+++ b/app/controllers/api/v1/meetings_controller.rb
@@ -124,7 +124,11 @@ module Api
       end
 
       def authorized_to_start_meeting?(settings, bbb_role)
-        settings['glAnyoneCanStart'] == 'true' || bbb_role == 'Moderator'
+        if(settings['glAnyoneJoinAsModerator'])
+          settings['glAnyoneCanStart'] == 'true' || @room.user_id == current_user&.id
+        else
+          settings['glAnyoneCanStart'] == 'true' || bbb_role == 'Moderator'
+        end
       end
 
       def unauthorized_access?(settings)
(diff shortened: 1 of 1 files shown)


app/controllers/api/v1/meetings_controller.rb
```diff
@@ -124,7 +124,11 @@ def authorized_as_moderator?(mod_code:, viewer_code:, anyone_join_as_mod:)
       end
 
       def authorized_to_start_meeting?(settings, bbb_role)
-        settings['glAnyoneCanStart'] == 'true' || bbb_role == 'Moderator'
+        if settings['glAnyoneJoinAsModerator'] == 'true'
+          settings['glAnyoneCanStart'] == 'true' || @room.user_id == current_user&.id
+        else
+          settings['glAnyoneCanStart'] == 'true' || bbb_role == 'Moderator'
+        end
```
farhatahmad, line 131: Can we not just do only `settings['glAnyoneCanStart'] == 'true' || @room.user_id == current_user&.id || room is shared`

farhatahmad: It's a little more complicated actually

farhatahmad: `settings['glAnyoneCanStart'] == 'true' || @room.user_id == current_user&.id || current_user&.shared_rooms&.include?(@room)`

SilentFlameCR: updated as you mentioned!

app/controllers/api/v1/meetings_controller.rb
```diff
@@ -124,7 +124,11 @@ def authorized_as_moderator?(mod_code:, viewer_code:, anyone_join_as_mod:)
       end
 
       def authorized_to_start_meeting?(settings, bbb_role)
-        settings['glAnyoneCanStart'] == 'true' || bbb_role == 'Moderator'
+        if settings['glAnyoneJoinAsModerator'] == 'true'
+          settings['glAnyoneCanStart'] == 'true' || @room.user_id == current_user&.id || current_user&.shared_rooms&.include?(@room)
+        else
+          settings['glAnyoneCanStart'] == 'true' || bbb_role == 'Moderator'
+        end
```
farhatahmad, line 131: You can drop the whole if statement and just do `settings['glAnyoneCanStart'] == 'true' || @room.user_id == current_user&.id || current_user&.shared_rooms&.include?(@room)`

PR #5763, opened 2024-03-28
fix(migrations): Set user password_digest only if it was received

danimo:
This fixes situations where after a migration from v2 due to a peculiar behavior, which refuses empty passwords but not empty password hashes, could end up with an empty password.


danimo:
Can you please merge this? it's still security relevant.

PR #5778, opened 2024-04-12
sorted pending users as latest first

### LINKED ISSUE #5775: Pending tab in Administrator Panel should be sorted by latest first
The pending tab when "Approve/Decline" is set is sorted by the oldest first, when it should be the newest first

SilentFlameCR:
Fixes #5775


farhatahmad:
Merged this by accident - the sorting should always be done on the backend - not the frontend

SilentFlameCR:
okay will redo this!

PR #5779, opened 2024-04-15
fixed avatar eicar check

SilentFlameCR:
setAvatar now sending original_avatar and eidted avatar for the clamav check. Scanning for viruses fixed for avatar.


farhatahmad (approved):

defnull:
The avatar that is actually stored and later displayed do visitors is not scanned? I can simply send a normal image as `user[original_avatar]` and a malicious one as `user[avatar]` to bypass the clamav check.

defnull:
Ah, so this is just for generating a proper error message on upload, I see. Thanks for the explanation.

PR #5790, opened 2024-04-26
Support for Scalelite Tagged Servers

Ithanil:
This PR adds frontend support for "Tagged Servers" in Scalelite, a feature which is described here: https://github.com/blindsidenetworks/scalelite/pull/1049

## Description
This PR allows user to configure two new options for each room:
1) The server tags from a pre-defined list
2) Whether the the server tag is hard required or fallback is allowed when no server is available

In terms of UI, it presents itself as seen in the following screenshots:
Screenshot from 2024-04-29 15-48-45
Screenshot from 2024-04-29 15-48-56

The list of available options is defined by the Greenlight administrator and certain tags can be restricted to certain role ids. In the current iteration of this feature, this is done via .env config file and is documented in `sample.env`:

```
## Support for Tagged Servers
# If your Greenlight instance is connected to Scalelite or another Loadbalancer with enabled support for the 'meta_server-tag'
# parameter on create calls, you can use the following variables to configure support for this feature via the Greenlight UI.
# When this configuration is changed later, disallowed tags can be removed from the DB via `bundle exec rake server_tags_sync`
# Example configuration (delimiters are , : and /):
# SERVER_TAG_NAMES=tag1:Name 1,tag2:Name2  # defines available tags and their friendly names
# SERVER_TAG_ROLES=tag2:xyz-123-321-aaaa-zyx/abc-321-123-zzzz-cba  # allow tag only for given role ids (see role ids in DB)
```

The required database migration is included and well tested. No RSpec tests have been added, but the feature is tested in deployment for multiple rooms, configuration combinations etc.

## Notes
- ~~Currently, if a room has been configured with a tag and then later the tag is removed from the site-wide configuration or permission is revoked for the user's role, the room's configured tag will remain in the database and will still be used for create calls. However, the configuration field will show as empty in the UI and only allow changes to valid/allowed tags.~~ Solved by commit 29c0c4935de48f7ad8bc7e454d081072b85d50ac and further improved by 2928003293411faf1de74af77d316759fa0c0efd
- ~~Currently, the role-based restriction of tags is only done by sending a list of allowed tags to the JavaScript code and not again validated on the backend side. This seems OK as long as the impact of potential malicious users abusing known "VIP" tags is tolerable.~~ Solved by commits 2928003293411faf1de74af77d316759fa0c0efd & ffcd97280a5232593eb33cb73fd5a0245e025fb4, invalid/forbidden tags will be dropped by the meeting_starter and they show as default type / untagged in the UI. 
- The two server tag options have already been added as global room configuration options (for consistency and for not requiring any future migrations), but are neither available for configuration in the admin UI nor does their value override the individual room's configuration.

## What is "missing" / Future Development
- ~~Add a sync task for the administrator to remove invalid tags from rooms after configuration file change~~ Done (29c0c4935de48f7ad8bc7e454d081072b85d50ac)
- ~~Validate role-based tag restriction on backend side (I could use a hint on how and at which point this should be done ideally, would also be happy if someone else could contribute this)~~ Done (2928003293411faf1de74af77d316759fa0c0efd / ffcd97280a5232593eb33cb73fd5a0245e025fb4)
- Make the global server tag options configurable, via the global room setting UI, and make their values actually work as overrides. (I'm not going to contribute this)
- Translate the current file/env-based configuration to UI-based configuration in the site and role setting menus. (I'm not going to contribute this)


farhatahmad:
There are some architecture changes that I would make here to match the way the rest of the application works:

-  Don't add the env vars to react unless they are necessary
  - What I would do instead is create a `ServerTagsController` and push all of the logic related to fetching the tags into there (including the role stuff)
  - then, if server tags is not enabled, just return [ ] instead of a list of available tags (and dont show the dropdown)
  - You can then remove it completely from `CurrentUserSerializer`

Ithanil:
> There are some architecture changes that I would make here to match the way the rest of the application works:
> 
>     * Don't add the env vars to react unless they are necessary
> 
>     * What I would do instead is create a `ServerTagsController` and push all of the logic related to fetching the tags into there (including the role stuff)
> 
>     * then, if server tags is not enabled, just return [ ] instead of a list of available tags (and dont show the dropdown)
> 
>     * You can then remove it completely from `CurrentUserSerializer`
> 
> 
> Once this change is made, I think it would simplify the logic quite a bit

Thanks four your review and I agree with your suggestions. I knew my iteratively developed solution would need a refactoring, but I wanted to have your opinion first to avoid doing it twice. However, I think I won't get it done today as it's basically weekend already here. ;-)

Ithanil:
@farhatahmad That said, do you think the implementation can stay as is in terms of UI/functionality and DB migrations? Because then I would be able to deploy the feature soon for our users independent of the review/merge/release status.

Ithanil:
@farhatahmad I'm done with refactoring. It required me to understand more of the architecture, but as a result I think the code has greatly improved in quality. Also, now invalid/forbidden tags will silently be dropped by meeting_starter (i.e. falling back to untagged), which matches what a user would see in that case (default label, i.e. untagged).

farhatahmad:
Definitely a lot cleaner and in line with the way Greenlight is designed. I'll do a full indepth review tomorrow

app/controllers/api/v1/server_tags_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+
+module Api
+  module V1
+    class ServerTagsController < ApiController
+      # GET /api/v1/server_tags/:friendly_id
+      # Returns a list of all allowed tags&names for the room's owner
+      def show
+        room = Room.find_by!(friendly_id: params[:friendly_id])
+        tag_names = Rails.configuration.server_tag_names
+        tag_roles = Rails.configuration.server_tag_roles
```
farhatahmad, line 27: I think it may be better to add a guard clause here to handle the case where server tags is not enabled

```
tag_names = Rails.configuration.server_tag_names
tag_roles = Rails.configuration.server_tag_roles

return render_data data: {}, status: :ok if tag_names.blank?
```

app/controllers/api/v1/server_tags_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class ServerTagsController < ApiController
+      # GET /api/v1/server_tags/:friendly_id
+      # Returns a list of all allowed tags&names for the room's owner
+      def show
+        room = Room.find_by!(friendly_id: params[:friendly_id])
```
farhatahmad, line 25: `find_by!` throws an exception if the room_id is incorrect, but it's not handle with a rescue clause in this method

app/javascript/components/rooms/room/room_settings/RoomSettings.jsx
```diff
@@ -66,6 +69,15 @@ export default function RoomSettings() {
                 config={roomConfigs?.glModeratorAccessCode}
                 description={t('room.settings.generate_mods_access_code')}
               />
+              {(!serverTags.isLoading && Object.keys(serverTags?.data).length !== 0) && (
```
farhatahmad, line 72: I don't think we need to explicitly handle the `isLoading` here. You can just do 
`const { data: serverTags } = useServerTags(friendlyId)` and then I would just check if `serverTags?.data?.length` and that should work

app/controllers/api/v1/server_tags_controller.rb
Ithanil: Fair enough, but it's the same in any other controller with a find_room method, e.g. https://github.com/bigbluebutton/greenlight/blob/master/app/controllers/api/v1/rooms_controller.rb

app/javascript/components/rooms/room/room_settings/RoomSettings.jsx
Ithanil: True, I introduced the isLoading check when the second condition was not used yet.

app/controllers/api/v1/server_tags_controller.rb
Ithanil: At least according to my testing, right now exactly the same happens when the SERVER_TAG_NAMES env is unset or empty.

app/javascript/components/rooms/room/room_settings/RoomSettings.jsx
Ithanil: addressed in https://github.com/bigbluebutton/greenlight/pull/5790/commits/2919f08a6c7e030655af7dba104048d5dc5eef49

Ithanil: @farhatahmad I thought I had a working solution without isLoading, but it turns out it can still happen I get errors like "can't convert undefined to object".

app/controllers/api/v1/server_tags_controller.rb
Ithanil: Anyway addressed in https://github.com/bigbluebutton/greenlight/pull/5790/commits/b3c2b47b40fda56b101c513ff40f7c342558341b . The benefit is indeed that a useless Room.find is avoided.

app/javascript/components/rooms/room/room_settings/RoomSettings.jsx
Ithanil: resolved in 892385937b7a91d8c48bed0b9e9ab24280304acc

db/data/20240423162700_create_server_tags_option.rb
```diff
@@ -0,0 +1,61 @@
...
+    Tenant.all.each do |tenant|
+      unless RoomsConfiguration.exists?(meeting_option: tag_required_option, provider: tenant.name)
+        RoomsConfiguration.create!(meeting_option: tag_required_option, value: 'optional', provider: tenant.name)
+      end
+    end
+    Room.find_each do |room|
+      unless RoomMeetingOption.exists?(room:, meeting_option: tag_required_option)
+        RoomMeetingOption.create!(room:, meeting_option: tag_required_option, value: 'false')
+      end
+    end
```
farhatahmad, line 47: This definitely needs to be cleaned up and optimized for better performance. `find_each` for large deployments will cause the deployment to crash. We're also looping twice over all Rooms when we can loop once. Here's some suggested improvements (in kinda pseudo code)

```
create serverTag and serverTagRequired meeting options
create serverTag and serverTagRequired RoomConfigurations
create serverTag and serverTagRequired RoomConfigurations for all tenants

Room.find_in_batches |group|
  group.each do |room|
    create serverTag and serverTagRequired RoomMeetingOption
end
```

Ithanil: @farhatahmad I'm working on it right now and am testing it on our largest database.
Already one thing FYI: find_each does in fact use a batch size of 1000 per default. It's not like find_all.

Ithanil: With the latest commit [ca4217b](https://github.com/bigbluebutton/greenlight/pull/5790/commits/ca4217b0e37a4d9462927cad18361f516e171767) the migration will only take about a minute where it took over 2 hours before and still retains the "safe but slow"  variant if existing tag options are detected.

PR #5806, opened 2024-05-13
Mail Notification when User Role is changed

fmo1:
In Glv3 it is not as handy to map users to roles on behalf of ldap-attributes...
In our Use-Case everyone in our company has to be able to log in to our BigBlueButton-Server... (via KeyCloak) just not everyone should be authorized to create and start meetings... so we decided to instruct administratiors to promote users by hand if necessary when receiving "New User Sign Up" - Mails... This works well... but we want the users to get notfified via mail when their role gets promoted... The changes from this branch implements this "feature"

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -100,6 +100,8 @@ module Api
       # Updates the values of a user
       def update
         user = User.find(params[:id])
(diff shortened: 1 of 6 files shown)


fmo1 (commented):
It's a feature request altough calling this feature is exaggerated in my opinon

PR #5807, opened 2024-05-13
Do not require email for signup with invite

byteplow:
Currently, users must enter their email address when they sign up with an invitation. This email must match the email address for the invitation.
This pr changes that so that users do not have to enter their email address:
1. The email field is removed from the sign-up form when the registration method is invite.
2.  The user creation functions uses the email from the invite token.

Potential issue: The invite token is used as a unique index by itself. The documentations for has_secure_token states:
> Note that it’s still possible to generate a race condition in the database in the same way that validates_uniqueness_of can. You’re encouraged to add a unique index in the database to deal with this even more unlikely scenario.

I lack the understanding of active recording to judge, if this is the problem in this case.

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -49,8 +49,19 @@ module Api
 
         registration_method = SettingGetter.new(setting_name: 'RegistrationMethod', provider: current_provider).call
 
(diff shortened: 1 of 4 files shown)


farhatahmad:
Thanks for this PR - this definitely improves the workflow for invited users. I'll try to get this into release 3.3.7

Issue #5814, opened 2024-05-18
Japanese name transferred inverted from Keycloak to GLv3


hiroshisuga:
For external authentication we use Keycloak, which requires "first name" and "last name" when the account is created. They are compulsory fields; I cannot login GLv3 if leaving these fields blank.

It seems these first and last name will be transferred to GLv3 to show the user's name. In my case "Hiroshi" is the first name and "Suga" is the last name, and I see "Hiroshi Suga" nicely at the right top corner of GLv3. So far so good.

Now a problem happens when I enter Japanese name. My first name is "裕" and my last name is "菅", which is shown like "裕　菅" on GLv3. However, in Japanese we express our name in the "last name" "first name" order. So it has to be shown as "菅 　裕" on GLv3. 

Would it be possible to transfer the "Username" field of Keycloak to GLv3? If yes, I would propose the "first name" and "last name" field optional because they seem not to used any other place.

hiroshisuga:
This is the case for Microsoft365 authentication. I have not tested other external authentications.

hiroshisuga:
I asked ChatGPT:

What is the countries/regions where they arrange last name first?

ChatGPT
In several countries and regions, it is common practice to arrange names with the last name (surname) first, followed by the first name (given name). Here are some notable examples:

East Asia
China: Names are traditionally written with the family name first, followed by the given name. For example, "Mao Zedong" (Mao is the family name, Zedong is the given name).
Japan: Japanese names are typically written with the family name first. For example, "Tanaka Hiroshi" (Tanaka is the family name, Hiroshi is the given name). However, in international contexts, Japanese people may reverse the order to align with Western conventions.
Korea: Both South Korea and North Korea follow the practice of placing the family name first. For example, "Kim Jong-un" (Kim is the family name, Jong-un is the given name).
Southeast Asia
Vietnam: Vietnamese names are usually written with the family name first, followed by the given name. For example, "Nguyen Thi Minh Khai" (Nguyen is the family name, Thi Minh Khai is the given name).
Hungary
Hungary: In Hungarian, names are traditionally written with the family name first, followed by the given name. For example, "Nagy István" (Nagy is the family name, István is the given name).
Other Examples
Mongolia: Names are written with the family name first. For example, "Bat-Erdene Badmaanyambuu" (Bat-Erdene is the family name, Badmaanyambuu is the given name).
Basque Country (Spain): Basque names sometimes follow the surname-first format, though this is less common today.
Summary
[...]

PR #5840, opened 2024-06-10
External users can no longer change their names, similar to emails

### LINKED ISSUE #5418: Don't allow changing username when user is externalUser
I thing a user should not be able to edit it's displayName when it is an externalUser (OpenID-Connect).
Currently a user can edit it's name in the settings menu.

alihadimazeh:
resolves #5418

### DIFF
--- a/app/javascript/components/users/user/forms/UpdateUserForm.jsx
+++ b/app/javascript/components/users/user/forms/UpdateUserForm.jsx
@@ -70,7 +70,7 @@ export default function UpdateUserForm({ user }) {
 
   return (
     <Form methods={methods} onSubmit={updateUserAPI.mutate}>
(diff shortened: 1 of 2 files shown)


app/javascript/components/users/user/forms/UpdateUserForm.jsx
```diff
@@ -70,7 +70,7 @@ export default function UpdateUserForm({ user }) {
 
   return (
     <Form methods={methods} onSubmit={updateUserAPI.mutate}>
-      <FormControl field={fields.name} type="text" />
```
farhatahmad, line 73: This means it is always readonly - we want to make it readonly only if the user is external

hiroshisuga:
Would this (#5814) become a problem then?

PR #5851, opened 2024-06-20
Invitation expiration date changed to 7 days

### LINKED ISSUE #5847: Invitations Valid column always shows checkmark even if the invitation is no longer valid
Invitations should only be valid for 48 hours, if they are invalid, there should be a red X that appears instead of the checkmark

image

alihadimazeh:
resolves #5847 

- Invitation expiration date changed to 7 days. 
- Serializer updated to check for validity of invite.

### DIFF
--- a/app/serializers/invitation_serializer.rb
+++ b/app/serializers/invitation_serializer.rb
@@ -20,6 +20,8 @@ class InvitationSerializer < ApplicationSerializer
   attributes :id, :email, :updated_at, :valid
 
   def valid
-    object.updated_at.in(Invitation::INVITATION_VALIDITY_PERIOD)
+    expiration_time = object.updated_at + Invitation::INVITATION_VALIDITY_PERIOD
+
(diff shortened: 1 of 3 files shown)


app/serializers/invitation_serializer.rb
```diff
@@ -20,6 +20,8 @@ class InvitationSerializer < ApplicationSerializer
   attributes :id, :email, :updated_at, :valid
 
   def valid
-    object.updated_at.in(Invitation::INVITATION_VALIDITY_PERIOD)
+    expiration_time = object.updated_at + Invitation::INVITATION_VALIDITY_PERIOD
+
+    Time.current <= expiration_time
```
farhatahmad, line 25: We can actually make this a bit simpler

```
object.updated_at < Invitation::INVITATION_VALIDITY_PERIOD.ago
```

farhatahmad (approved):

Issue #5872, opened 2024-07-01
Fallback mechanism on external authentication effectively makes email address a primary identifier


Ithanil:
In app/controllers/external_controller.rb (https://github.com/bigbluebutton/greenlight/blob/6b1f93d8ca2e536702e0c6638627b31345d3665e/app/controllers/external_controller.rb#L34C1-L40C8) there is a fallback mechanism implemented, which effectively renders the email address a primary identifier of a user. According to the code, it behaves as follows:

1) On Login via external authentication, it checks if a user with `external_id` already exists
2) If it doesn't exist, it checks if the email address already exists for a user
3) If a user with that email is found, it updates the `external_id` of that user with the value according to the current login

This is a very unexpected behavior in the context of external authentication, where the `external_id` should be the identifier of the user, and results in problems in environments where email addresses may eventually be reassigned to new persons.

Consider the following:
1) User A logs into GL3 at some point
2) User A leaves the organization or changes his email address
3) User B joins the organization and receives the email address that User A once had
4) User B logs into GL3 and has now access to the account/rooms of User A, including all recordings

I'm not sure in which context you would want such a fallback, but it should be at the minimum configurable and not be the default.

farhatahmad:
This was done as a back up when we were migrating from v2 to v3. The fallback can be removed

FSeifer:
@farhatahmad, @Ithanil 


I generally agree that the external_id of a user should be the leading indentifier and thus we should be able to update a changed email address....

But what would be the expected behaviour in the following scenario:

- A GLV3-Setup with local accounts is switched to an IDP Setup
- Hence there are users, who have an email-address set but no external_id yet
- How would the initial mapping be done if not via the email?

I feel this is not something we can toggle via a configuration switch but rather a fundamental feature of the IDP-controller..

I hope you can understand what I mean.

Regards 
FSeifer

Ithanil:
@FSeifer What I meant with my response in https://github.com/bigbluebutton/greenlight/pull/5876 is a config flag to enable the previous behavior, i.e. fallback to mail as ID. If the previous behavior was what you wanted, that should serve you well, right?

Otherwise, in your situation I would go and update all the external_id entries in the database to their correct new values according to the IDP.

FSeifer:
@Ithanil 
But can that be our goal? 
[...]

PR #5876, opened 2024-07-05
Remove email fallback

Ithanil:
Don't fall back to email as identifier if no user with external_id found.

For more details see https://github.com/bigbluebutton/greenlight/issues/5872

## 

On top of this change, I introduced the following mechanism for our setup, to allow users to log in if they received an email that was previously owned by a different external account (which has previously logged into Greenlight). Without this code, Greenlight will not create the new user due to conflicting email addresses.
```
    # Replace any existing email with "dummy" value
    email_user = User.find_by(email: user_info[:email], provider:)
    if email_user.present? && user_info[:external_id].present? && email_user.external_id != user_info[:external_id]
      email_user.update(email: "#{email_user.id}@example.net")
    end
```
This changes the email associated with that other account to a dummy value, which does not cause issues and will be updated on the next login. The only alternative would be to completely delete the other account in Greenlight.

@farhatahmad Let me know if you want this code in GL, maybe behind a config flag.

### DIFF
--- a/app/controllers/external_controller.rb
+++ b/app/controllers/external_controller.rb
@@ -32,13 +32,6 @@ class ExternalController < ApplicationController
 
     user = User.find_by(external_id: credentials['uid'], provider:)
 
-    # Fallback mechanism to search by email
-    if user.blank?
-      user = User.find_by(email: credentials['info']['email'], provider:)
-      # Update the user's external id to the latest value to avoid using the fallback
-      user.update(external_id: credentials['uid']) if user.present? && credentials['uid'].present?
-    end
-
     new_user = user.blank?
 
     registration_method = SettingGetter.new(setting_name: 'RegistrationMethod', provider: current_provider).call


farhatahmad (approved):

FSeifer:
Hello @Ithanil 

May I ask wh this email fallback was removed?

This change broke our BBB-Setup, since we have an IDP configured for with pre-existing users and rely on this mechanism to set the external_id of a user that has not yet logged in via our IDP.

With this change, instead of a successful login, we get a generic error message and a support ticket...

Regards
FSeifer

Ithanil:
> Hello @Ithanil
> 
> May I ask why this email fallback was removed?
> 
> This change broke our BBB-Setup, since we have an IDP configured for with pre-existing users and rely on this mechanism to set the external_id of a user that has not yet logged in via our IDP.
> 
> With this change, instead of a successful login, we get a generic error message and a support ticket...
> 
> Regards FSeifer

I'm sorry that the change broke your setup, that's why I proposed to use a config flag for this behavior. Regarding the rationale, see the corresponding issue that was linked above: https://github.com/bigbluebutton/greenlight/issues/5872

PR #5891, opened 2024-07-22
Refactored LinksForm.jsx component to improve performance and avoid unnecessary API calls

nirajkumar999:
In continuation to my PR #5887, as discussed in my Seventh Improvement point, we can extend the performance optimization by preventing unnecessary onSubmit of Form in case of tnc, privacy policy and help center fields. So PATCH request will only be sent if there is actual change in the  value.

I have introduced a useRef variable formText which initially stores the actual value of these fields and further tracks any changes to the value.

In handleSubmit function we call mutation only when this formText variable shows any changes. Hence this way we prevent any unnecessary PATCH request when user continuously clicks the Change URL function.


farhatahmad:
Also needs to be rebased please

nirajkumar999:
Rebased and updated my commit message to include details for the changes done in code @farhatahmad

farhatahmad:
Needs to be updated again please

nirajkumar999:
Updated

PR #5893, opened 2024-07-24
Greenlight version number only appears for admins

### LINKED ISSUE #5884: [V3] Feature: Hide Greenlight Version Number
The initial issue #3810 was closed due to the release of greenlight 3. But the concern is still the same and still valid.

Telling every visitor the version number of greenlight without authentication is a security issue, because it tells the attacker valid and useful information.

my suggestion: move the version string in the login area for the administrators and remove it from the footer.

alihadimazeh:
- Changes to App.jsx: passed in currentUser to the Footer
- Footer renders the version number if the current user is an admin
- resolves #5884

### DIFF
--- a/app/javascript/components/shared_components/Footer.jsx
+++ b/app/javascript/components/shared_components/Footer.jsx
@@ -17,19 +17,21 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { Container } from 'react-bootstrap';
+import PropTypes from 'prop-types';
 import useEnv from '../../hooks/queries/env/useEnv';
 import useSiteSetting from '../../hooks/queries/site_settings/useSiteSetting';
 
-export default function Footer() {
+export default function Footer({ currentUser = null }) {
   const { t } = useTranslation();
   const { data: env } = useEnv();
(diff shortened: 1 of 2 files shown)


app/javascript/components/shared_components/Footer.jsx
```diff
@@ -17,19 +17,21 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { Container } from 'react-bootstrap';
+import PropTypes from 'prop-types';
 import useEnv from '../../hooks/queries/env/useEnv';
 import useSiteSetting from '../../hooks/queries/site_settings/useSiteSetting';
 
-export default function Footer() {
+export default function Footer({ currentUser = null }) {
```
farhatahmad, line 24: You don't need to get `currentUser` through props. Props are only needed if you need to track their state across multiple components. In this case, we can just get current user directly in the Footer component

farhatahmad (approved):

sebastianberm:
Hmm, why not make this a setting instead with a default of hidden? - We use this to see which instances we still need to update...

PR #5919, opened 2024-08-13
5815 allowed email domains

### LINKED ISSUE #5815: [3.0 feature request] Signup only with specific Email domains
Would it be possible to enable signing up only with specific Email domains when an external authentication is used such as microsoft365)?

It was possible on v2 (#1548), but on v3 I do not find how to set it up.
Maybe Keycloak has such function?

(I thought I posted this request somewhere, but I cannot find it.)

alihadimazeh:
resolves #5815


farhatahmad (approved):

hiroshisuga:
Thanks @alihadimazeh and @farhatahmad !
Just for user's information, I was using a workaround in the keycloak side:
https://keycloak.discourse.group/t/how-to-restrict-registration-to-specific-email-domains/702/19
And this also works well.

PR #5930, opened 2024-08-21
Flag to enable email as a fallback identifyer. Solution to issue #5872

SebastianAppDev:
Hello everyone,

As discussed in issue #5872, there are various pros and cons to using email as a fallback identifier.

As some setups depend on it, I have implemented a flag to enable the old behaviour of using email as a fallback identifier. Setting USE_EMAIL_AS_EXTERNAL_ID_FALLBACK=true will cause greenlight to behave as it did before the fallback was removed, which is important for the operation of some systems migrated from version 2, and others with users that existed before the use of an IDP.

If you have any questions, please ask.


cwaidelich:
Hi,
how can I help to have this Pull Request merged and integrated?
At our business we have this pull requested working in our PRD enviroment, with 250 users / daily, and it is working as expected with no issues.

farhatahmad (approved):

farhatahmad:
I'll merge this and fix rubocop in a separate PR

PR #5951, opened 2024-09-29
Fixes to roles_controller update & delete action. Changes need to be reflected in SiteSetting DefaultRole setting.

nirajkumar999:
### First Commit - Extended return condition to render error if Users are associated with role.

- I have tried to modify return condition for error status :method_not_allowed by adding additional OR condition which checks if any User is associated with the role because if we don't manually do it here we get a Internal Server Error from Postgres stating foreign key constraint violation which is correct at its place.

- But at UI we get "The action cant be completed" in the toast message. This don't align with our requirements. So i have manually added the OR condition check.

- This way I eliminate Internal Server Error stuff which is a good practice. Also the toast shows the proper message as already stated in useDeleteRole hook. The same message gets displayed when we try to delete the pre defined roles.

**PFA Screenshot of the issue :** 

Suppose I have a user with Role as : test-role

Now when i try to delete test-role i get toast as : 

image

and in my terminal i get : 

image
image

After Fixing the issue : 

Toast I get : 

image

And on terminal i get : 

image

********************************************************************************************************

### Second Commit - Modified update and destroy action route for roles_controller. Added further error handler to useDeleteRole.jsx

- In update action we chain the name update to SiteSetting DefaultRole. This ensures consistency.

- We avoid the delete action for the role that is assigned as DefaultRole in SiteSetting. If we don't do this we don't have any side effects as this case is already handled by setting fallback to User role, in case the DefaultRole is not present. But I think this should be handled properly by giving a correct error message. We return status: :forbidden in that case and the same is handled in the useDeleteRole hook.

- What I observe is that if the update and delete action is not aligned with SiteSetting Default Role then the drop-down appearing for the setting in the UI is left empty without any valid selection. This isn't good practice.

**PFA Screenshot of the issue :** 

Suppose I have a Role as : test-role. This i set as default role in site setting

image

Now when i update the test-role to test-role-update, it gets successfully updated while in the UI the dropdown selection gets empty.

image

Same situation happens when we try to delete this test-role. Deletion happens successfully.

Note : We can delete only if no users are associated to that role. Here I am considering no users are associated and then when we delete

After Fixing the issue : 

On Update  : 

image

On Delete the toast i get :

image

On terminal output shows as  :

image

### DIFF
--- a/app/assets/locales/en.json
+++ b/app/assets/locales/en.json
@@ -452,7 +452,8 @@
       "signin_required": "You must be signed in to access this page.",
       "malware_detected": "Malware Detected! The file you uploaded may contain malware. Please check your file and try again.",
       "roles": {
-        "role_assigned": "This role can't be deleted because it is assigned to at least one user."
+        "role_assigned": "This role can't be deleted because it is assigned to at least one user.",
+        "role_set_as_default": "This role can't be deleted because it is set as Default Role in Site Setting."
       },
       "users": {
         "signup_error": "You can't be authenticated. Please contact your administrator.",
--- a/app/controllers/api/v1/admin/roles_controller.rb
+++ b/app/controllers/api/v1/admin/roles_controller.rb
@@ -66,7 +66,7 @@ module Api
         # Deletes a role
         def destroy
           undeletable_roles = %w[User Administrator Guest]
-          return render_error errors: @role.errors.to_a, status: :method_not_allowed if undeletable_roles.include?(@role.name)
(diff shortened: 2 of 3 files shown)


nirajkumar999:
I think some test cases defined are not expected to work this way. I dont have good knowledge to change the test cases. I leave this on your team. 

For update actions its easy to chain the changes upto site setting default role. Hence no issue here.

For delete action, what i did is prevented the role from being deleted. But in your code you have already set a fallback to USER role. If this is the case let me know what design you want to follow. If you want a successful deletion and assign USER role to site setting default role. We can even do that.

nirajkumar999:
@farhatahmad please review

PR #5965, opened 2024-11-14
admins can change external users's names

alihadimazeh:


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2023_12_18_154727) do
```
farhatahmad, line 13: This file shouldnt't be included in the changes

app/controllers/api/v1/users_controller.rb
```diff
@@ -169,7 +169,9 @@ def create_user_params
       end
 
       def update_user_params
-        @update_user_params ||= if external_auth?
+        @update_user_params ||= if external_auth? && current_user.role.name.eql?('Administrator')
+                                  params.require(:user).permit(:name)
+                                elsif external_auth?
```
farhatahmad, line 174: Greenlight is a bit more tricky than this. The "Administrator" role can be renamed, or the permission to edit users can be given to other roles, meaning that just this check isn't enough. What you need to do instead is check if the user has the `ManageUsers` permission

farhatahmad: Here's how you would do that:

`PermissionsChecker.new(current_user:, permission_names: 'ManageUsers', current_provider:).call`

farhatahmad:
Also - did you test to make sure this works? There's a visual portion to this that is needed

Rspec tests are needed as well

app/controllers/api/v1/users_controller.rb
```diff
@@ -169,7 +169,11 @@ def create_user_params
...
       def update_user_params
-        @update_user_params ||= if external_auth?
+        is_admin = PermissionsChecker.new(current_user:, permission_names: 'ManageUsers', current_provider:).call
+
+        @update_user_params ||= if external_auth? && is_admin
+                                  params.require(:user).permit(:name)
+                                elsif external_auth?
                                   params.require(:user).permit(:password, :avatar, :language, :role_id, :invite_token)
                                 else
                                   params.require(:user).permit(:name, :password, :avatar, :language, :role_id, :invite_token)
```
farhatahmad, line 179: return params.require(:user).permit(....) if external_auth? && is_admin
return params.require(:user).permit(....) if external_auth?
params.require(:user).permit(....)

farhatahmad (approved):

Issue #5982, opened 2025-01-23
Hungarian language pack - dates do not appear correctly


a-ttila:
When I switch to Hungarian language, most of the dates and times do not appear properly:

![Image](https://github.com/user-attachments/assets/cb62477a-3211-4fea-a31f-478fb7b54e42)
![Image](https://github.com/user-attachments/assets/07dbf725-0342-49e9-91f7-a5f148acd908)

I tried many different languages, everyting works as expected.

There must be something wrong with the Hungarian translation or language pack.

a-ttila:
There is the same problem at line 224:

We need "user_created_at": "Created: {{localizedTime}}" too, instead of  "user_created_at": "Létrehozva: {{user.created_at}}",

Thank you so much.

Attila

PR #5990, opened 2025-02-05
Fix last_session time display for HU and CS languages

Ithanil:
Fixes https://github.com/bigbluebutton/greenlight/issues/5982


ffdixon:
Thanks for the polish Jan -- much appreciated.

PR #5998, opened 2025-02-13
Added avatar support when passed through Keycloak

### LINKED ISSUE #5987: openid user avatar sync not working
Hello!

Is syncing the users avatar picture from the openid providers source to greenlight supposed to work? I set it to true at nextcloud oidc but the avatar picture doesn't change.

cheers!

farhatahmad:
Added support for avatars that are passed through under the `image` key

For keycloak, there's some mapping changes that will need to be made to add support (atleast for Google, likely for others as well)

fixes #5987

### DIFF
--- a/spec/controllers/external_controller_spec.rb
+++ b/spec/controllers/external_controller_spec.rb
@@ -437,6 +437,65 @@ RSpec.describe ExternalController do
         expect(User.find_by(email: OmniAuth.config.mock_auth[:openid_connect][:info][:email]).role).to eq(role1)
       end
     end
+
+    context 'avatar' do
+      before do
+        OmniAuth.config.mock_auth[:openid_connect] = OmniAuth::AuthHash.new(
+          uid: Faker::Internet.uuid,
+          info: {
+            email: Faker::Internet.email,
+            name: Faker::Name.name,
+            image: Faker::Avatar.image
+          }
+        )
+
+        request.env['omniauth.auth'] = OmniAuth.config.mock_auth[:openid_connect]
+        stub_request(:get, OmniAuth.config.mock_auth[:openid_connect][:info][:image])
+          .to_return(body: file_fixture('default-avatar.png'), headers: { 'Content-Type' => 'image/jpeg' }, status: 200)
+      end
+
+      it 'attaches the avatar to the user' do
+        get :create_user, params: { provider: 'openid_connect' }
+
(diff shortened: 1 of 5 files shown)


spec/controllers/external_controller_spec.rb
```diff
@@ -437,6 +437,65 @@
...
+
+      it 'attaches the avatar to the user' do
+        get :create_user, params: { provider: 'openid_connect' }
+
+        expect(User.find_by(email: OmniAuth.config.mock_auth[:openid_connect][:info][:email]).avatar).to be_attached
+      end
+
+      it 'does not re-attach the avatar if it hasnt changed' do
+        reg_method = instance_double(SettingGetter)
+        allow(SettingGetter).to receive(:new).with(setting_name: 'ResyncOnLogin', provider: 'greenlight').and_return(reg_method)
```
alihadimazeh, line 465: why do we need the `ReSynvOnLogin` registration method in this case?

spec/controllers/external_controller_spec.rb
```diff
@@ -437,6 +437,65 @@
...
+        allow(reg_method).to receive(:call).and_return(true)
+
+        profile_file = URI.parse(OmniAuth.config.mock_auth[:openid_connect][:info][:image])
+        filename = File.basename(profile_file.path)
+
+        user = create(:user, email: OmniAuth.config.mock_auth[:openid_connect][:info][:email])
+        user.avatar.attach(io: fixture_file_upload('default-avatar.png'), filename:, content_type: 'image/png')
+
+        expect(User.find_by(email: OmniAuth.config.mock_auth[:openid_connect][:info][:email]).avatar).not_to receive(:attach)
```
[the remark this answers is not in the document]
farhatahmad: Its part of Active Storage - the rails extension that handles file attachements

https://guides.rubyonrails.org/active_storage_overview.html#has-one-attached

spec/controllers/external_controller_spec.rb
farhatahmad: By default, ResyncOnLogin in tests is set to false. This specific case tests that a user that has already logged in wont have the same image reattached. 

So, by mocking it to true, it ensures that `handle_avatar` method actually runs again the second time the user logs in

alihadimazeh (approved):

PR #6003, opened 2025-02-23
Update HUN locale file

a-ttila:
Minor modification to the Hungarian language file according to the original English one.


a-ttila:
It was NOT a TRANSLATION update but a minor CODE update, so I think this CODE modification is still necessary...

PR #6004, opened 2025-02-25
[Minor fix] : Reversed if else condition. More general toast error message should be placed in else block

rautniraj:
Tried to follow the design as used in other hooks. Most general error toast should be in else block. While serverTagUnavailable stuff should be in if block handling. This will help us having scope to add further error handling cases in future


rautniraj:
I think there is typo in translation file **toast.error.server_type_unavailable** should be **toast.error.server_tag_unavailable**.

Please let me know i will fix this

farhatahmad (approved):

PR #6016, opened 2025-03-14
feat: add an empty line after each migration for readable output grouping

debuglevel:
It's quite hard to identify migration errors. Adding a newline should help.


rautniraj:
Just adding my opinion. We have also ran migration. Really it was a hell for us. We r having many other db changes and new models so need to modify existing migration and also we wrote few new migrations. We did not ran migration all at once. It was one by one. After finishing one model, then we moved to second one. So for us it was easy to debug. Moreover we directed the error message to a txt file so that we can analyze them well.

farhatahmad (approved):

PR #6021, opened 2025-03-18
[Minor fix] : Removed extra checks from isAdmin variable in Footer.jsx

rautniraj:
We dont need these two checks : `currentUser && currentUser.role 
`
We can handle this in a good way : `currentUser?.role?.name === 'Administrator'`

### DIFF
--- a/app/javascript/components/rooms/room/shared_access/SharedAccess.jsx
+++ b/app/javascript/components/rooms/room/shared_access/SharedAccess.jsx
@@ -39,7 +39,7 @@ export default function SharedAccess() {
   const deleteSharedAccess = useDeleteSharedAccess(friendlyId);
   const { data: room } = useRoom(friendlyId);
   const currentUser = useAuth();
-  const isAdmin = currentUser?.role.name === 'Administrator';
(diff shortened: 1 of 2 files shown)


rautniraj:
What i observe, we should also let super admin to view version tag in footer. My suggestion 

`  const isAdmin = currentUser?.role.name === 'Administrator' || currentUser?.role.name === 'SuperAdmin';`

farhatahmad:
> What i observe, we should also let super admin to view version tag in footer. My suggestion
  const isAdmin = currentUser?.role?.name === 'Administrator' || currentUser?.role?.name === 'SuperAdmin'

Agreed - can you make this change?

rautniraj:
Its done now. In Footer.jsx + SharedAccess.jsx element i have modified isAdmin check to include SuperAdmin also. These were the two places where isAdmin was being checked.

> [!IMPORTANT]
>  Please share your view on my very old PR #5951 . Suggested changes will improve the UI and application logic.

farhatahmad (approved):

farhatahmad:
Was having issues merging this, so I created a new PR

PR #6027, opened 2025-04-04
Manually Verify Users in Unverified Tab

alihadimazeh:
This feature is to allow admins to view any unverified users in its own 'Unverified' Tab under Manage Users, and manually verify the user

[Screencast from 2025-04-15 05:41:50 PM.webm](https://github.com/user-attachments/assets/5ca563aa-3e92-4be3-9bf8-85fb3b5e0240)

### DIFF
--- a/app/controllers/api/v1/admin/users_controller.rb
+++ b/app/controllers/api/v1/admin/users_controller.rb
@@ -51,13 +51,13 @@ module Api
         end
 
         # GET /api/v1/admin/users/verified.json
-        # Fetches all active users
+        # Fetches all verified users
         def verified
           sort_config = config_sorting(allowed_columns: %w[name roles.name])
 
           users = User.includes(:role)
                       .with_provider(current_provider)
-                      .where(status: 'active')
+                      .where(verified: true)
                       .with_attached_avatar
                       .order(sort_config, created_at: :desc)&.search(params[:search])
 
--- a/app/controllers/api/v1/admin/users_controller.rb
+++ b/app/controllers/api/v1/admin/users_controller.rb
@@ -99,7 +99,7 @@ module Api
         private
 
         def user_params
-          params.require(:user).permit(:status)
+          params.require(:user).permit(:status, :verified)
         end
       end
     end
--- a/app/controllers/api/v1/admin/users_controller.rb
+++ b/app/controllers/api/v1/admin/users_controller.rb
@@ -66,6 +66,22 @@ module Api
           render_data data: users, meta: pagy_metadata(pagy), serializer: UserSerializer, status: :ok
         end
 
+        # GET /api/v1/admin/users/unverified.json
+        # Fetches all unverified users
+        def unverified
+          sort_config = config_sorting(allowed_columns: %w[name roles.name])
+
+          users = User.includes(:role)
+                      .with_provider(current_provider)
+                      .where(verified: false)
+                      .with_attached_avatar
+                      .order(sort_config, created_at: :desc)&.search(params[:search])
+
+          pagy, users = pagy(users)
+
+          render_data data: users, meta: pagy_metadata(pagy), serializer: UserSerializer, status: :ok
+        end
+
         # GET /api/v1/admin/users/banned.json
         # Fetches all banned users
         def banned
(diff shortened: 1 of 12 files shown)


app/javascript/components/admin/manage_users/UnverifiedUsers.jsx
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <div>
+      {
+        (searchInput && unverifiedUsers?.data.length === 0)
+          ? (
+            <div className="mt-5">
+              <NoSearchResults text={t('user.search_not_found')} searchInput={searchInput} />
+            </div>
+          ) : (
+            <ManageUsersTable users={unverifiedUsers?.data} isLoading={isLoading} pagination={unverifiedUsers?.meta} setPage={setPage} />
```
alihadimazeh, line 39: @farhatahmad would this need to have its own table component like Pending and Banned for it to be displayed in the Tabs. As of now it currently does not

app/controllers/api/v1/admin/users_controller.rb
```diff
@@ -51,13 +51,29 @@ def pending
...
 
         # GET /api/v1/admin/users/verified.json
-        # Fetches all active users
+        # Fetches all verified users
         def verified
           sort_config = config_sorting(allowed_columns: %w[name roles.name])
 
           users = User.includes(:role)
                       .with_provider(current_provider)
-                      .where(status: 'active')
```
farhatahmad, line 60: What is the `status` field used for? Did you double check before removing this?

app/controllers/api/v1/admin/users_controller.rb
```diff
@@ -83,7 +99,7 @@ def banned
         private
 
         def user_params
-          params.require(:user).permit(:status)
+          params.require(:user).permit(:status, :verified)
```
farhatahmad, line 102: Whats the purpose behind this change?

app/javascript/components/admin/manage_users/BannedPendingRow.jsx
```diff
@@ -30,13 +30,49 @@ import Avatar from '../../users/user/Avatar';
...
+    }
+    if (tableType === 'unverified') {
+      return (
+        <Dropdown.Item onClick={() => updateUserVerification.mutate({ id: user.id, verified: true })}>
+          <CheckIcon className="hi-s me-2" />
+          {t('admin.manage_users.verify')}
+        </Dropdown.Item>
+      );
+    }
+    return null;
```
farhatahmad, line 73: Making these `else if`s will save a bit of unnecessary checks

app/javascript/hooks/queries/admin/manage_users/useUnverifiedUsers.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+  const params = {
+    'sort[column]': searchParams.get('sort[column]'),
+    'sort[direction]': searchParams.get('sort[direction]'),
+    search: input,
+    page,
+  };
+
+  return useQuery(
+    ['getAdminUsers', { ...params }],
```
farhatahmad, line 32: I would use a different key here to avoid issues

app/controllers/api/v1/admin/users_controller.rb
alihadimazeh: Indicates whether is active or banned, but I don't think the `verified` method properly reflects that. Would you be open to me adding an `active` method?

app/controllers/api/v1/admin/users_controller.rb
alihadimazeh: to allow for the verified param to pass through to the controller

app/controllers/api/v1/admin/users_controller.rb
farhatahmad: Its not necesarrily about adding another method. If the `status` field filters `active`, `pending` and `banned` then what would removing it do?

app/controllers/api/v1/admin/users_controller.rb
farhatahmad: Ooo correct - my mistake

app/javascript/components/admin/manage_users/BannedPendingRow.jsx
alihadimazeh: eslint defaults to that

farhatahmad (approved):

Issue #5211, opened 2023-05-31
greenlight v3 - Logout not function with keycloak


alyfilipe:
Dear,

When I logout in the GL V3 it does not sensitize the keycloack, which keeps my session active. If I try to log in again, it doesn't ask for my credentials and already sends me logged in to GL . Is this correct? How to make GL inform the keycloak about the logout session? Logout URL to force user to log out of keycloack?

Thank you for options so that you can finalize the tests and put it into production.

drudgede:
At least I found a workaround for now.

Login to the Keycloak Admin Panel -> Head to the greenlight realm -> In the left sidebar, click "Authentication", select the "browser" login flow and set the first step called "Cookie" to "Disabled". 

I don't know yet which side effects this settings might have but at least the logout now works for me.

tzengshinfu:
Thanks for sharing. The results of my testing indicate that after signing out and attempting to log in with a different account, the error shown in the following image appears. It seems that logging in with the same account is the only viable option.
![image_20231114_110519](https://github.com/bigbluebutton/greenlight/assets/2170033/bba19f2e-c67a-4c9d-8379-dfaf25f8665d)

ffdixon:
On https://demo.bigbluebutton.org/, I can do the following  

1. Sign in and authenticate with my gmail account.
   - I'm prompted to pick a gmail account and authenticate
2. Signout of GreenLight
3. Sign in again -- this time it bypassess the prompt for sign-in options and automatically logs me in

In step (3), I had expected to get prompted again for sign-in options, which suggests GreenLight is skipping this step and my session token was still valid.

I've found if I wait a bit -- some number of minutes -- and then try signing in again, I'm prompted again with the sign-in options.

I think signing out should invalidate any cookies or cache of the previous sign-in choice.

tzengshinfu:
I found that if I sign out the session of GreenLight V3 in the KeyCloak Administration Console, I also need to manually sign out in GreenLight; otherwise, it stays logged in.

PS: Checking the GreenLight documentation (https://docs.bigbluebutton.org/greenlight/v3/external-authentication), there is no mention of setting a Backchannel logout URL.

gopyns:
You can disable cookie login for keycloak:

1. Login Keycloak greenlight realm as admin
2. Open your greenlight realm's authentication page (/keycloak/admin/master/console/#/greenlight/authentication/) 
3. Select "Browser flow" (default "Browser flow" is Browser)
[...]

PR #6042, opened 2025-04-24
Add logout logic when using external login

### LINKED ISSUE #6019: [3.x] - user logged in with openid does not actually logout from Keycloack.
Greenlight 3.x with Keycloack configured according to the  [manual](https://docs.bigbluebutton.org/greenlight/v3/external-authentication/.)

Step to reproduce:
- Login Greenlight with Google or Github provider via Keycloack login screen.
- Press **Sign out** button and see Greenlight home **Welcome to BigBlueButton** page. Session is still active in Keycloack in session list.
- Press  **login** in Greenlight and get to room list without Keycloack login screen.

According to the code https://github.com/bigbluebutton/greenlight/blob/v3/config/initializers/omniauth.rb
logout url is not called

also in this file https://github.com/bigbluebutton/greenlight/blob/v3/app/controllers/api/v1/sessions_controller.rb
Greenlight delete own session only without calling Keycloack logout URL.

farhatahmad:
fixes #6019 #5211 

This PR adds an explicit call to the openid_connect logout endpoint to ensure the user is logged out when clicking the log out button. 

Behaviour for a local logout is unchanged

### DIFF
--- a/app/controllers/api/v1/sessions_controller.rb
+++ b/app/controllers/api/v1/sessions_controller.rb
@@ -68,8 +68,25 @@ module Api
       # DELETE /api/v1/sessions/signout
       # Clears the session cookie and signs the user out
       def destroy
-        sign_out
-        render_data status: :ok
+        id_token = session.delete(:oidc_id_token)
+
+        # Build RP-initiated logout URL
+        if id_token.present? && ENV['OPENID_CONNECT_ISSUER'].present?
+          end_session = File.join(
+            ENV.fetch('OPENID_CONNECT_ISSUER', nil),
+            'protocol', 'openid-connect', 'logout'
+          )
+          url = "#{end_session}?client_id=#{ENV.fetch('OPENID_CONNECT_CLIENT_ID', nil)}" \
+                "&id_token_hint=#{id_token}" \
+                "&post_logout_redirect_uri=#{CGI.escape(root_url(success: 'LogoutSuccessful'))}" \
+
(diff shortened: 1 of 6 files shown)


app/controllers/external_controller.rb
```diff
@@ -28,6 +28,8 @@ def create_user
 
     credentials = request.env['omniauth.auth']
 
+    session[:oidc_id_token] = credentials.dig('credentials', 'id_token')
```
alihadimazeh, line 31: what's the point of setting this here? Does it get passed to somewhere else?

farhatahmad: its used to pass the `id_token_hint` param in the logout url

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -68,8 +68,25 @@ def create
...
+                "&id_token_hint=#{id_token}" \
+                "&post_logout_redirect_uri=#{CGI.escape(root_url(success: 'LogoutSuccessful'))}" \
+
+          # also revoke local session so SPA state is consistent
+          sign_out
+          render_data data: url, status: :ok
+        else
+          sign_out
+          render_data status: :ok
+        end
```
alihadimazeh, line 89: can we write a test for this?

farhatahmad: No I should be able - let me try adding a test

alihadimazeh (approved):

PR #6046, opened 2025-05-02
[Minor Fix] : missing table sort params added to useBannedUsers and usePendingUsers hook

rautniraj:
Sort params were missing in both the hooks. This enables table sorting based on allowed columns


farhatahmad:
I don't think these should necessarily be sortable. They should just be sorted by created at desc

rautniraj:
Depends on the requirement. We had that requirement and I found the same was missing in your code. If you dont want that to be sortable then we should remove the same from frontend also. Then it will be consistent. Let me know If you want that update

PR #6050, opened 2025-05-08
Added superadmin check to version number

farhatahmad:
closes #6021


rautniraj:
PR #6021 Do we dont need same modification to SharedAccess.jsx where we are checking isAdmin ?
I have added the change but it was not merged.

Issue #5816, opened 2024-05-18
[3.0] Need to clarify visibility states (which is confusing..)


hiroshisuga:
Somewhere in the document, we need to clarify what the visibility states of recording stand for.

In my understanding,
Public/Protected: Title visible without sign-in, but the user cannot watch the recording itself
Public: Without signing in, the user can see the title and watch the video
Protected: Title visible only when the user signs in, but the recording is not watchable for him
Published: The user can see the title and watch the video, but only he signs in. 
Unpublished: Nobody (the recording owner neither) can see the title nor the recording itself.

Are the correct? 

In my local environment, "Public/Protected" and "Public" have no difference. Even without signing in, the user can see the list ant watch the recording. 

Also the words "protected" and "unpublished" are not so distinctive for users, maybe "unlisted" and "private" like in Youtube?

To be less confusing, what about a setting like:

Recording list visible in the room: [everybody] [only sign-in users] [only the owner]
Recording watchable: [everybody] [only sign-in users] [only the owner]

A recording that is visible in the list only for the owner but watchable for everybody means the"unlisted" in Youtube, so everybody can watch it if he knows the URL.

rosa2:
I was looking for documentation, because I don't understand the difference. My testing:

- Public/Protected: Without signing in, the title visible, the user gets the icon to copy the url and can watch the recording itself
- Public: Without signing in, the title visible, the user  gets the icon to copy the url and can watch the recording itself
- Protected: Title visible only when the user signs in, but the recording is not watchable for him or an unsigned user.
- Published: The user can see the title and watch the video, but only he signs in.
- Unpublished: Nobody (the recording owner neither) can see the title nor the recording itself.

Maybe is the server configuration, I don't know.
It would be great one possibility with password, as for the conference.

Thanks thanks a lot!

loicfaure-univ-rennes:
Hello, I'm bringing this subject up again. It's very difficult to understand how record access links work.

PR #6064, opened 2025-05-14
Add access code for public recordings

sebclemens:
This pull request adds support for protecting **public recordings** with a dedicated **access code**.

Here is my first request to this feature:
https://github.com/bigbluebutton/greenlight/issues/6043

**Feature Demo**
Below is a short demonstration of the new functionality in action:

add-access-code-for-public-recordings

**What this adds**

- Users can now share public recordings securely by assigning an access code.
- Viewers must enter the access code to access the recording.

This improves security and flexibility for sharing content with external audiences.

PR #6136, opened 2025-10-23
enh: Add contextual popovers for recording visibility help

Ndaboom:
- explain the visibility states and formats via localized popovers on the recordings table
- Popper auto-place the dialogs so they stay readable near screen edges

Screenshot From 2025-10-23 19-16-19

Related issue: https://github.com/bigbluebutton/greenlight/issues/5816

### DIFF
--- a/app/assets/locales/en.json
+++ b/app/assets/locales/en.json
@@ -188,11 +188,11 @@
     "formats_help": "Shows the available playback types (for example, Presentation or Video). Clicking a format opens the BigBlueButton recording in a new tab.",
     "visibility_help": {
       "title": "Recording visibility options",
-      "public": "Public: Listed on the public recordings page and viewable without signing in.",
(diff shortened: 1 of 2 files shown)


Ndaboom:
> @Ndaboom I suggest the inline documentation follow the same order.
> 
> Ex.: Public/Protected is first in my instance.

Done, Thank you.

farhatahmad:
Thanks for submitting the PR - this is a great addition. Maybe just some small changes regarding the wording used:

Public/Protected - The recording is available on the Public Recordings page that can be reached from the Room Join page. The link can only be viewed from within Greenlight
Public - The recording is available on the Public Recordings page that can be reached from the Room Join page. The link can be shared with anyone
Protected - The recording must be viewed from within Greenlight
Published - The recording is accessible to anyone with the link
Unpublished - The recording is not accessible to anyone

marclaporte:
Related: https://github.com/bigbluebutton/greenlight/pull/6064

Issue #6257, opened 2026-04-08
i18n locale loading fails for regional locales (e.g. de-DE)


dev-cb:
When the application initializes, i18next attempts to load locale files based on the browser language (e.g. de-DE).

The backend only provides base language files (e.g. de.json, en.json), but not regional variants like de-DE.json.

**As a result:**
* GET /api/v1/locales/de-DE.json → 406 Not Acceptable
* Response body is empty (Content-Length: 0)
* The frontend does not handle this failure gracefully

This leads to a runtime error:
```
TypeError: Cannot read properties of null (reading 'data')
```

And finally React rendering fails during initialization.

**Observed Behavior:**
Despite valid fallback languages being loaded, the failed de-DE request causes the app to crash.

**Expected Behavior:**
The app should either
1. Normalize de-DE → de before requesting locale files, or
2. Gracefully fall back to de / en if a regional locale is not available

A failed locale request must not crash the application

**Assumed root cause:**
Frontend configuration uses:
```js
load: 'currentOnly'
```

This forces i18next to load the exact locale (de-DE) instead of falling back to the base language (de).

Backend only supports base locales (de, en) and returns 406 for unsupported locales.

Additionally, the frontend does not defensively handle failed HTTP responses.

---

Besides that, I suggest to ensure that HTTP failures are handled safely.
Avoid assuming response shape (response.data) without validation.

dev-cb:
### Update: Root Cause Identified

The issue is caused by how the active language is set in combination with the i18next configuration.

#### Relevant Code

In `app/javascript/App.jsx`:

```jsx
const { i18n } = useTranslation();

useEffect(() => {
  i18n.changeLanguage(currentUser?.language);
}, [currentUser?.language]);
```

At runtime:

```js
currentUser.language === 'de-DE'
document.documentElement.lang === 'de-DE'
```

#### Problem

i18next is configured with:

```js
load: 'currentOnly'
```

This causes i18next to request the exact locale:

```text
/api/v1/locales/de-DE.json
```

However, the backend only provides base language files:

- `/api/v1/locales/de.json` ✅
- `/api/v1/locales/en.json` ✅
- `/api/v1/locales/de-DE.json` ❌ → 406 Not Acceptable

#### Result

- The request for `de-DE.json` fails with `406`
- The response body is empty
- The frontend does not handle this failure correctly
- This leads to a runtime error:

```text
TypeError: Cannot read properties of null (reading 'data')
```

[...]

PR #6258, opened 2026-04-08
Fix i18n locale loading for regional codes (de-DE, en-US, etc.)

dev-cb:
- fixes issue #6257
- Switch load strategy from 'currentOnly' to 'languageOnly' to normalize
  regional locale codes (de-DE → de) matching available backend files
- Simplify fallbackLng to static array since normalization is automatic
- Add error handling to i18n.changeLanguage() to prevent crashes on
  locale load failures

### DIFF
--- a/app/javascript/App.jsx
+++ b/app/javascript/App.jsx
@@ -85,8 +85,13 @@ export default function App() {
   // i18n
   const { i18n } = useTranslation();
   useEffect(() => {
-    i18n.changeLanguage(currentUser?.language);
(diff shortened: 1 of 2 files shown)


farhatahmad:
This breaks the current fallback logic. Today a browser language of de-DE produces the chain ["de-DE", "de", "en"], but load: 'languageOnly' reduces it to ["de", "en"], dropping the dialect.

We should always try  the dialect first when available, then fall back to the base language, and finally to English

PR #6278, opened 2026-05-11
Implementing custom vanity URL

lonesomewalker:
Changes neccessary for custom set URL.
Checks if URL already in database, gives hints how valid URL can be chosen.
Previous (already deleted) PR commited to root folder...


lonesomewalker:
Commit ad23f01a301b7dcd60250e9c6f9a2c1fe91d9460 is NOT neccessary, but might come when i am done writing more for the API and integrating the PAT (personal access token) to be able to connect with various external tools (like Thunderbird, Outlook, M365) for appointment planning.

farhatahmad:
Thanks for the PR - I'd much rather put this behind a Site Setting (with it defaulted to false) becuase I'm sure a lot of people won't be happy with this (for security reasons)

If you can make that change, I can look into merging this as part of the next release

PR #6292, opened 2026-06-11
Feature/dial in sip pin generation

dbm-dataport:


lonesomewalker:
Default is 9 digits PIN now

PR #6317, opened 2026-08-29
Gemfile: Remove binary platform from lock file

onny:
We have issues packaging Greenlight for Nix and manually need to patch out the custom x86_64 platform https://github.com/NixOS/nixpkgs/pull/550046
If you're not dependent on it, we would be happy for upstreaming this patch :)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -178,7 +178,6 @@ GEM
     faraday-net_http (3.4.4)
       net-http (~> 0.5)
     ffi (1.17.2)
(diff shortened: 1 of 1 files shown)


farhatahmad:
Thanks for the PR - the issue is that dropping the precompiled platform means the CI has to build nokogiri's vendored libxml2/libxslt from source on every lockfile change, which is going to increase the CI run time by quite a bit 

Considering you have another way around this (its non-blocking) - I'd prefer to keep it as is

